PR #15, opened 2016-07-11
derive clone for Theme and SyntaxDefinition

TyOverby:


trishume:
Thanks. IMO deriving debug and Clone should be the default, I also had to submit a clone implementation pull request (to rust-onig) in the making of syntect.

PR #16, opened 2016-07-11
Derive Eq and PartialEq for the two state structs [ DO NOT MERGE ]

TyOverby:
This PR will only compile after https://github.com/rust-onig/rust-onig/pull/17 has been merged.

### DIFF
--- a/src/parsing/parser.rs
+++ b/src/parsing/parser.rs
@@ -23,13 +23,13 @@ use fnv::FnvHasher;
 ///
 /// **Note:** Caching is for advanced users who have tons of time to maximize performance or want to do so eventually.
 /// It is not recommended that you try caching the first time you implement highlighting.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 pub struct ParseState {
     stack: Vec<StateLevel>,
     first_line: bool,
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq, Eq)]
 struct StateLevel {
     context: ContextPtr,
     prototype: Option<ContextPtr>,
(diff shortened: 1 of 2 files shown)


TyOverby:
I've noticed that the parsing and highlighting states are *very* compressible for large files just by using a basic run-length encoding.  

At the beginning of every line, the parser is almost always in the same parse state (usually "inside method body").  By implementing Eq, I can cache the hell out of the HighlightState and ParseState.

trishume:
Cool, that's one way to do it.

Did you read the note on caching in the Readme? I also wrote some more in a comment here: https://github.com/google/xi-editor/pull/43

IMO you would be better off just caching states every 100 lines. That will probably use less memory and you won't eat the cost of comparing parser state stacks. It also has better worst case memory/performance since you can construct terrible cases for run length encoding. It also gives strict maximums on incremental re-highlighting time.

TyOverby:
Oh yeah, for sure.  I'm probably going to implement a hybrid approach where I'll try run-length encoding, and if it's getting too noisy, I'll just have "holes" that can be re-colorized when they are requested (with a maximum hole-size of 100).

And that's not even counting the styled-text caching which is a given.

TyOverby:
> Re-highlighting 100 lines in a hard syntax takes 7ms so still possibly within one or two frames after the keystroke.

I'm aiming for 0-latency on a 90hz display, so 7ms is 60% of my frame-budget.  I'll play around with parameters though.  Maybe even make it heuristics based eventually.  I'm certainly going to start out with a dumb strategy though.

> the actual highlighting of the scope-annotated text is about 30x faster than the parsing

Ooh, that's interesting.  I'll keep that in mind.

TyOverby:
@trishume Now that the PR has been merged in upstream, could you kick off these tests again?

PR #9, opened 2016-06-29
Make Clippy Happy

killercup:
It still has some things it doesn't like (e.g. acronyms in doc comments) and wants me to use map_or and friends. I don't care about that stuff, though.

PR #19, opened 2016-08-25
fixed some clippy warnings

llogiq:
Just a few readability improvements.


trishume:
Thanks! Looks like this caught a bunch of things #9 didn't, I wonder if Clippy got better in the mean time or you were just more thorough.

llogiq:
We get better all the time. :smile:

PR #22, opened 2016-10-13
Update dependencies found via `cargo outdated`

robinst:
The only breaking change was that `bitflags!` no longer generates `pub` by default.


robinst:
Would you mind releasing a 1.1.0 or 1.0.1 with these changes? :)

trishume:
done. It's on crates.io. Thanks again for your contributions!

PR #27, opened 2016-12-22
Update Packages and add support for clear_scopes

### LINKED ISSUE #26: Update included dump of Sublime text syntax definitions?
Hey. It looks like there needs to be a release for #25.

Would it be too much to ask to update to the newest syntax definitions from Sublime text :)? Looks like there have been some updates for Rust, among others: https://github.com/sublimehq/Packages/commits/master/Rust

trishume:
- [x] Update the Sublime Packages submodule to the latest version
- [x] Add `clear_scopes` to the `SyntaxDefinition` type
- [x] Fix test failures due to updated syntaxes
- [x] Implement the new `clear_scopes` setting in the parser
- [x] Benchmark and ensure that performance hasn't regressed.

I'm currently waiting on a reply to my comment on https://github.com/sublimehq/Packages/pull/466 to see if I understand `clear_scopes` before implementing it. I'm planning on storing a `Vec<Scope>` in `StateLevel` of scopes to push back on to the stack when that level is popped.

fixes #26 

cc @robinst


trishume:
So I ran some benchmarks and it seems that these changes come with a 4% performance regression   with the same version of the JS sublime-syntax, and an 8% regression comparing this branch to master including the newer JS sublime-syntax. The new syntax has more features and gives better results so I think the 4% is a better number, and I think that's acceptable. I'll also try running the other benchmarks.

robinst:
Would you mind cutting a release with these changes included (probably `1.1.0`)? Switching over the regex engine might take a while :).

trishume:
@robinst done.

robinst:
Thanks, works nicely!

PR #32, opened 2017-01-16
Add an option to statically link libonig

johannhof:
We're using syntect over at [Cobalt](https://github.com/cobalt-org/cobalt.rs) and it's an awesome library. However, we currently can't enable it by default because Windows and musl and a couple of other targets require onig to be statically linked, I think.

This is a pull request that adds an optional feature that toggles the corresponding feature in rust-onig.

Note that I had some test failures locally (OSX), even before applying this change.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,5 +25,8 @@ bincode = "0.6"
 flate2 = "0.2"
 fnv = "1.0.2"
 
+[features]
+static-onig = ["onig/static-libonig"]
+
 # [profile.release]
 # debug = true


trishume:
Cool thanks! Glad to hear you're using `syntect` in Cobalt!

johannhof:
Neat! Would you mind bumping the version for us? We like to avoid using git dependencies if possible.

Thanks!

johannhof:
Also nvm the failing tests, I forgot to run `git submodule update --init`

trishume:
No problem, done.

Issue #33, opened 2017-01-24
600 line Rust source file takes more than 2 minutes


robinst:
How to reproduce ([link to view file](https://github.com/rust-lang/rust/blob/a6fe6c9be10c79d64c988d490519f5cc07e9060d/src/librustc_borrowck/borrowck/mir/dataflow/impls.rs)):

```
curl -LO 'https://github.com/rust-lang/rust/raw/a6fe6c9be10c79d64c988d490519f5cc07e9060d/src/librustc_borrowck/borrowck/mir/dataflow/impls.rs'
time target/release/examples/synstats impls.rs
```

Running with syntect from master (currently 431174d97fe8df2fc798e11c96991e989b23a1e9), this is the output of `time`:

```
target/release/examples/synstats impls.rs  141.49s user 0.54s system 99% cpu 2:22.71 total
```

Sublime text build 3125 seems to have no problem with the file and loads instantly.

trishume:
Whoah that's messed up. And even with release mode. I wonder if that's a regression or if all versions do badly on that file.

I'll try and take a look at this soon. Unfortunately I have school assignments, but I may find time this week, as well as for finishing #27.

robinst:
Thanks. Tried with 1.0.0 (7981ad3dfabb0d7921466b064f9e79f6f9bd4f50) just now, same problem (153.66s).

jmacdonald:
I think I've run into this; the Rust highlighter performance is terrible with one particular file (I've not run into this before, and have used syntect pretty extensively). [Here](https://gist.github.com/jmacdonald/78aa5b254c9824a96a9177de4beb7b4c)'s the file I was highlighting.

I profiled my application to isolate the poorly performing function, and it appears to be in the `onig` library:
![screen shot 2017-01-31 at 10 39 44 pm](https://cloud.githubusercontent.com/assets/458253/22494554/57af9f1e-e806-11e6-9863-17c63e72c592.png)

Happy to provide other details and/or help reproduce, as necessary!

trishume:
@jmacdonald Interesting. From what I remember of previous profiling, 50% of the time being taken in the Regex engine is pretty typical, it's in fact lame that it isn't more of the total run time, but the reengineering to fix that would be tricky. 

In fact if anything the fact that it is only 50% tells me that it is probably not just a super bad regex that takes exponential time or something like that. It's probably some set of parser states that is doing way too much work per character of the file parsed, maybe going over it a bunch of times when it otherwise shouldn't.

[...]

PR #34, opened 2017-02-09
[WIP] Kinda-working fancy-regex support

### LINKED ISSUE #76: Fix nonewlines mode
As noted in https://github.com/trishume/syntect/pull/34#issuecomment-280808755, `nonewlines` mode uses `\z` in character classes, but that just acts as `z` and not end of string. This presents a few problems:

- [ ] Patch it so `nonewlines` doesn't mess up in the presence of `z`.
- [ ] Act properly in terms matching the end of the string, this may require more advanced regex syntax rewriting rather than just running regex replacements on regexes.

Specifically, [this line](https://github.com/trishume/syntect/blob/4dd550da2cdbe3ec782eaad9f8a12c5b111474c2/src/parsing/yaml_load.rs#L271) needs to be fixed to do a proper replacement strategy that doesn't replace `\n` with `\z` when the `\n` is in a character class, and maybe replaces the character class with `(?:[...]|\z)` where `...` is everything that was in the character class except the `\n`. This probably would also need a different rewrite for inverted character classes.

cc @robinst

trishume:
This branch switches the regex engine to [fancy-regex](https://github.com/google/fancy-regex) or more specifically [my fork of it](https://github.com/trishume/fancy-regex).

Currently it only works for a few syntaxes because of a few different features fancy-regex doesn't support:
- [x] The `\n` escape (Everything, but fixed it my fork)
- [x] Unnecessary escapes in character classes like `[\<]`
- [x] The `\h` escape in character classes (Rust)
- [x] Fix #76 so `nonewlines` mode doesn't produce weird regexes.
- [x] Named backrefs `\k<marker>` (Markdown)
- [x] Fancy character class syntax `[a-w&&[^c-g]z]`
- [ ] Add support for match limit to fancy-regex: https://github.com/google/fancy-regex/issues/44

The jQuery highlighting benchmark now takes 1s instead of 0.66s. Which is super unfortunate given that I'd hoped it would be faster than Oniguruma. I have no idea why it is substantially slower.

@raphlinus @robinst

### DIFF
--- a/src/parsing/syntax_definition.rs
+++ b/src/parsing/syntax_definition.rs
@@ -256,9 +256,10 @@ impl MatchPattern {
 
     fn compile_regex(&mut self) {
         // TODO don't panic on invalid regex
-        println!("compiling {:?}", self.regex_str);
-        let compiled = fancy_regex::Regex::new(&self.regex_str).unwrap();
-        self.regex = Some(compiled);
+        self.regex = match fancy_regex::Regex::new(&self.regex_str) {
+            Ok(regex) => Some(regex),
+            Err(e) => panic!("Error compiling regex {:?}: {:?}", self.regex_str, e),
+        };
     }
 
     /// Makes sure the regex is compiled if it doesn't have captures.
--- a/src/parsing/syntax_definition.rs
+++ b/src/parsing/syntax_definition.rs
@@ -5,12 +5,13 @@
 //! into this data structure?
 use std::collections::{BTreeMap, HashMap};
 use std::hash::Hash;
-use onig::{Regex, RegexOptions, Region, Syntax};
+use fancy_regex;
 use std::rc::{Rc, Weak};
 use std::cell::RefCell;
-use super::scope::*;
 use regex_syntax::escape;
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use super::scope::*;
+use super::util::Region;
 
 pub type CaptureMapping = Vec<(usize, Vec<Scope>)>;
 pub type ContextPtr = Rc<RefCell<Context>>;
@@ -86,17 +87,29 @@ pub struct MatchIter {
     index_stack: Vec<usize>,
 }
 
-#[derive(Debug, Eq, PartialEq, Serialize, Deserialize)]
+#[derive(Debug, Serialize, Deserialize)]
 pub struct MatchPattern {
     pub has_captures: bool,
     pub regex_str: String,
     #[serde(skip_serializing, skip_deserializing)]
-    pub regex: Option<Regex>,
+    pub regex: Option<fancy_regex::Regex>,
     pub scope: Vec<Scope>,
     pub captures: Option<CaptureMapping>,
     pub operation: MatchOperation,
     pub with_prototype: Option<ContextPtr>,
 }
+
+impl PartialEq for MatchPattern {
+    fn eq(&self, other: &MatchPattern) -> bool {
+        self.regex_str == other.regex_str &&
+        self.scope == other.scope &&
+        self.captures == other.captures &&
+        self.operation == other.operation &&
+        self.with_prototype == other.with_prototype
+    }
+}
+
+impl Eq for MatchPattern {}
 
 /// This wrapper only exists so that I can implement a serialization
 /// trait that crashes if you try and serialize this.
@@ -236,20 +249,15 @@ impl MatchPattern {
 
     /// Used by the parser to compile a regex which needs to reference
     /// regions from another matched pattern.
-    pub fn compile_with_refs(&self, region: &Region, s: &str) -> Regex {
+    pub fn compile_with_refs(&self, region: &Region, s: &str) -> fancy_regex::Regex {
         // TODO don't panic on invalid regex
-        Regex::with_options(&self.regex_with_substitutes(region, s),
-                            RegexOptions::REGEX_OPTION_CAPTURE_GROUP,
-                            Syntax::default())
-            .unwrap()
+        fancy_regex::Regex::new(&self.regex_with_substitutes(region, s)).unwrap()
     }
 
     fn compile_regex(&mut self) {
         // TODO don't panic on invalid regex
-        let compiled = Regex::with_options(&self.regex_str,
-                                           RegexOptions::REGEX_OPTION_CAPTURE_GROUP,
-                                           Syntax::default())
-            .unwrap();
+        println!("compiling {:?}", self.regex_str);
+        let compiled = fancy_regex::Regex::new(&self.regex_str).unwrap();
         self.regex = Some(compiled);
     }
 
@@ -296,28 +304,29 @@ fn ordered_map<K, V, S>(map: &HashMap<K, V>, serializer: S) -> Result<S::Ok, S::
 }
 
 
-#[cfg(test)]
-mod tests {
-    use super::*;
-    #[test]
-    fn can_compile_refs() {
-        use onig::{SearchOptions, Regex, Region};
-        let pat = MatchPattern {
-            has_captures: true,
-            regex_str: String::from(r"lol \\ \2 \1 '\9' \wz"),
-            regex: None,
-            scope: vec![],
-            captures: None,
-            operation: MatchOperation::None,
-            with_prototype: None,
-        };
-        let r = Regex::new(r"(\\\[\]\(\))(b)(c)(d)(e)").unwrap();
-        let mut region = Region::new();
-        let s = r"\[]()bcde";
-        assert!(r.match_with_options(s, 0, SearchOptions::SEARCH_OPTION_NONE, Some(&mut region)).is_some());
-
-        let regex_res = pat.regex_with_substitutes(&region, s);
-        assert_eq!(regex_res, r"lol \\ b \\\[\]\(\) '' \wz");
-        pat.compile_with_refs(&region, s);
-    }
-}
+// TODO
+// #[cfg(test)]
+// mod tests {
+//     use super::*;
+//     #[test]
+//     fn can_compile_refs() {
+//         use onig::{SearchOptions, Regex, Region};
+//         let pat = MatchPattern {
+//             has_captures: true,
+//             regex_str: String::from(r"lol \\ \2 \1 '\9' \wz"),
+//             regex: None,
+//             scope: vec![],
+//             captures: None,
+//             operation: MatchOperation::None,
+//             with_prototype: None,
+//         };
+//         let r = Regex::new(r"(\\\[\]\(\))(b)(c)(d)(e)").unwrap();
+//         let mut region = Region::new();
+//         let s = r"\[]()bcde";
+//         assert!(r.match_with_options(s, 0, SearchOptions::SEARCH_OPTION_NONE, Some(&mut region)).is_some());
+// 
+//         let regex_res = pat.regex_with_substitutes(&region, s);
+//         assert_eq!(regex_res, r"lol \\ b \\\[\]\(\) '' \wz");
+//         pat.compile_with_refs(&region, s);
+//     }
+// }
--- a/src/parsing/yaml_load.rs
+++ b/src/parsing/yaml_load.rs
@@ -425,11 +425,60 @@ fn replace_posix_char_classes(regex: String) -> String {
 }
 
 
-/// Some of the regexes include `$` and expect it to match end of line. In fancy-regex, `$` means
-/// end of text by default. Adding `(?m)` activates multi-line mode which changes `$` to match
-/// end of line.
+/// Some of the regexes include `$` and expect it to match end of line.
+/// In fancy-regex, `$` means end of text by default. Adding `(?m)` activates
+/// multi-line mode which changes `$` to match end of line.
+///
+/// But it also changes `^` to match beginning of line, which with fancy-regex
+/// also matches at the end of e.g. `"test\n"`.
+///
+/// So, we rewrite `^` to `\A` to always mean beginning of text, and add `(?m)`
+/// to change the meaning of `$`.
 fn regex_for_newlines(regex: String) -> String {
-    format!("(?m){}", regex)
+    if !regex.contains(|c| c == '^' || c == '$') {
+        return regex;
+    }
+
+    let rewriter = RegexRewriterForNewlines {
+        parser: Parser::new(regex.as_bytes()),
+    };
+    rewriter.rewrite()
+}
+
+struct RegexRewriterForNewlines<'a> {
+    parser: Parser<'a>,
+}
+
+impl<'a> RegexRewriterForNewlines<'a> {
+    fn rewrite(mut self) -> String {
+        let mut result = Vec::new();
+        result.extend_from_slice(br"(?m)");
+        while let Some(c) = self.parser.peek() {
+            match c {
+                b'^' => {
+                    self.parser.next();
+                    result.extend_from_slice(br"\A");
+                }
+                b'\\' => {
+                    self.parser.next();
+                    result.push(c);
+                    if let Some(c2) = self.parser.peek() {
+                        self.parser.next();
+                        result.push(c2);
+                    }
+                }
+                b'[' => {
+                    let (mut content, _) = self.parser.parse_character_class();
+                    result.append(&mut content);
+                }
+                _ => {
+                    self.parser.next();
+                    result.push(c);
+                }
+            }
+        }
+        String::from_utf8(result).unwrap()
+    }
 }
 
 
@@ -446,30 +495,28 @@ fn regex_for_no_newlines(regex: String) -> String {
         return regex;
     }
 
-    let rewriter = RegexRewriter {
-        bytes: regex.as_bytes(),
-        index: 0,
+    let rewriter = RegexRewriterForNoNewlines {
+        parser: Parser::new(regex.as_bytes()),
     };
     rewriter.rewrite()
 }
 
-struct RegexRewriter<'a> {
-    bytes: &'a [u8],
-    index: usize,
+struct RegexRewriterForNoNewlines<'a> {
+    parser: Parser<'a>,
 }
 
-impl<'a> RegexRewriter<'a> {
+impl<'a> RegexRewriterForNoNewlines<'a> {
     fn rewrite(mut self) -> String {
         let mut result = Vec::new();
-        while let Some(c) = self.peek() {
+        while let Some(c) = self.parser.peek() {
             match c {
                 b'\\' => {
-                    self.next();
-                    if let Some(c2) = self.peek() {
-                        self.next();
+                    self.parser.next();
+                    if let Some(c2) = self.parser.peek() {
+                        self.parser.next();
                         // Replacing `\n?` with `\z?` would make parsing later fail with
                         // "invalid operand of repeat expression"
-                        if c2 == b'n' && self.peek() != Some(b'?') {
+                        if c2 == b'n' && self.parser.peek() != Some(b'?') {
                             result.extend_from_slice(br"\z");
                         } else {
                             result.push(c);
@@ -480,8 +527,8 @@ impl<'a> RegexRewriter<'a> {
                     }
                 }
                 b'[' => {
-                    let (mut content, matches_newline) = self.parse_character_class();
-                    if matches_newline && self.peek() != Some(b'?') {
+                    let (mut content, matches_newline) = self.parser.parse_character_class();
+                    if matches_newline && self.parser.peek() != Some(b'?') {
                         result.extend_from_slice(b"(?:");
                         result.append(&mut content);
                         result.extend_from_slice(br"|\z)");
@@ -491,22 +538,53 @@ impl<'a> RegexRewriter<'a> {
                 }
                 b'(' => {
                     // `\z` can't be used in repeat or look-behind, so strip that
-                    if self.skip_if(br"(?:\n)?") ||
-                        self.skip_if(br"(?<!\n)") ||
-                        self.skip_if(br"(?<=\n)") {
+                    if self.parser.skip_if(br"(?:\n)?") ||
+                        self.parser.skip_if(br"(?<!\n)") ||
+                        self.parser.skip_if(br"(?<=\n)") {
                         // continue after skipped text
                     } else {
-                        self.next();
+                        self.parser.next();
                         result.push(c);
                     }
                 }
                 _ => {
-                    self.next();
+                    self.parser.next();
                     result.push(c);
                 }
             }
         }
         String::from_utf8(result).unwrap()
+    }
+}
+
+struct Parser<'a> {
+    bytes: &'a [u8],
+    index: usize,
+}
+
+impl<'a> Parser<'a> {
+    fn new(bytes: &[u8]) -> Parser {
+        Parser {
+            bytes,
+            index: 0,
+        }
+    }
+
+    fn peek(&self) -> Option<u8> {
+        self.bytes.get(self.index).map(|&b| b)
+    }
+
+    fn next(&mut self) {
+        self.index += 1;
+    }
+
+    fn skip_if(&mut self, s: &[u8]) -> bool {
+        if self.bytes[self.index..].starts_with(s) {
+            self.index += s.len();
+            true
+        } else {
+            false
+        }
     }
 
     fn parse_character_class(&mut self) -> (Vec<u8>, bool) {
@@ -533,15 +611,13 @@ impl<'a> RegexRewriter<'a> {
             match c {
                 b'\\' => {
                     self.next();
+                    content.push(c);
                     if let Some(c2) = self.peek() {
                         self.next();
                         if c2 == b'n' && !negated && nesting == 0 {
                             matches_newline = true;
                         }
-                        content.push(c);
                         content.push(c2);
-                    } else {
-                        content.push(c);
                     }
                 }
                 b'[' => {
@@ -565,23 +641,6 @@ impl<'a> RegexRewriter<'a> {
         }
 
         (content, matches_newline)
-    }
-
-    fn peek(&self) -> Option<u8> {
-        self.bytes.get(self.index).map(|&b| b)
-    }
-
-    fn next(&mut self) {
-        self.index += 1;
-    }
-
-    fn skip_if(&mut self, s: &[u8]) -> bool {
-        if self.bytes[self.index..].starts_with(s) {
-            self.index += s.len();
-            true
-        } else {
-            false
-        }
     }
 }
 
@@ -807,6 +866,26 @@ mod tests {
             - match: ''
         "#,false, Some("C"));
         assert_eq!(def.unwrap().name, "C");
+    }
+
+    #[test]
+    fn can_rewrite_regex_for_newlines() {
+        fn rewrite(s: &str) -> String {
+            regex_for_newlines(s.to_string())
+        }
+
+        assert_eq!(&rewrite(r"a"), r"a");
+        assert_eq!(&rewrite(r"\b"), r"\b");
+        assert_eq!(&rewrite(r"(a)"), r"(a)");
+        assert_eq!(&rewrite(r"[a]"), r"[a]");
+        assert_eq!(&rewrite(r"[^a]"), r"(?m)[^a]");
+        assert_eq!(&rewrite(r"[]a]"), r"[]a]");
+        assert_eq!(&rewrite(r"[[a]]"), r"[[a]]");
+
+        assert_eq!(&rewrite(r"^"), r"(?m)\A");
+        assert_eq!(&rewrite(r"$"), r"(?m)$");
+        assert_eq!(&rewrite(r"^ab$"), r"(?m)\Aab$");
+        assert_eq!(&rewrite(r"\^ab\$"), r"(?m)\^ab\$");
     }
 
     #[test]
--- a/src/parsing/yaml_load.rs
+++ b/src/parsing/yaml_load.rs
@@ -254,12 +254,13 @@ impl SyntaxDefinition {
                            state: &mut ParserState)
                            -> Result<MatchPattern, ParseSyntaxError> {
         let raw_regex = get_key(map, "match", |x| x.as_str())?;
-        let regex_str_1 = Self::resolve_variables(raw_regex, state);
-        // if the passed in strings don't include newlines (unlike Sublime) we can't match on them
+        let resolved_regex = Self::resolve_variables(raw_regex, state);
         let regex_str = if state.lines_include_newline {
-            regex_str_1
+            regex_for_newlines(resolved_regex)
         } else {
-            rewrite_regex(regex_str_1)
+            // If the passed in strings don't include newlines (unlike Sublime) we can't match on
+            // them, so try to rewrite the regex to not rely on newlines.
+            regex_for_no_newlines(resolved_regex)
         };
         // println!("{:?}", regex_str);
 
@@ -410,6 +411,15 @@ impl SyntaxDefinition {
     }
 }
 
+
+/// Some of the regexes include `$` and expect it to match end of line. In fancy-regex, `$` means
+/// end of text by default. Adding `(?m)` activates multi-line mode which changes `$` to match
+/// end of line.
+fn regex_for_newlines(regex: String) -> String {
+    format!("(?m){}", regex)
+}
+
+
 /// Rewrite a regex that matches `\n` to one that matches `\z` (end of string) instead.
 /// That allows the regex to be used to match lines that don't include a trailing newline character.
 ///
@@ -418,7 +428,7 @@ impl SyntaxDefinition {
 ///
 /// Note that the rewrite is just an approximation and there's a couple of cases it can not handle,
 /// due to `\z` being an anchor whereas `\n` matches a character.
-fn rewrite_regex(regex: String) -> String {
+fn regex_for_no_newlines(regex: String) -> String {
     if !regex.contains(r"\n") {
         return regex;
     }
@@ -787,9 +797,9 @@ mod tests {
     }
 
     #[test]
-    fn can_rewrite_regex() {
+    fn can_rewrite_regex_for_no_newlines() {
         fn rewrite(s: &str) -> String {
-            rewrite_regex(s.to_string())
+            regex_for_no_newlines(s.to_string())
         }
 
         assert_eq!(&rewrite(r"a"), r"a");
--- a/src/parsing/yaml_load.rs
+++ b/src/parsing/yaml_load.rs
@@ -253,14 +253,15 @@ impl SyntaxDefinition {
     fn parse_match_pattern(map: &Hash,
                            state: &mut ParserState)
                            -> Result<MatchPattern, ParseSyntaxError> {
-        let raw_regex = get_key(map, "match", |x| x.as_str())?;
-        let resolved_regex = Self::resolve_variables(raw_regex, state);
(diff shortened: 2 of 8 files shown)


raphlinus:
I took a quick look at this, profiling the highlighting of jquery. It's promising but clearly not compelling yet. It seems to be spending most of its time delegating to regex, but in the VM. This suggests that it's doing backtracking, and might not even be using the NFA (it delegates just to get classes). I have a bunch of ideas on how to optimize more, but don't have insight into specifically what's slow now. The best case would be something like `a(?=b)`, which currently throws `a` into fancy mode, but could be optimized to just `(a)b` with fixup of the captures.

The way to make progress here is to capture _which_ regexes are consuming the most time. I'd add some profiling, something like a lazy_static hash table on the side, so that every time the VM runs it increments a count for that regex, and accumulates the time. Then just go down the list in terms of which regexes burn the most time.

I'd be tempted to investigate myself, but am currently trying to really focus on incremental update in xi. Thanks for pushing this forward!

trishume:
@raphlinus Yah that was my thought on what to investigate as well. Since it's single-threaded I can do even better than a count for each regex, I can actually measure the total elapsed time and count per regex to figure out which ones are slow. Then I can run it again with Oniguruma and see which regexes are faster with fancy-regex and which are slower.

Unfortunately, I'm back to being busy with school work and I'm not sure when/if I'll have time to do this. The perf regression combined with missing features means it's going to be a bunch of work. Not an undoable amount, but still substantial.

TimNN:
@trishume: I'm trying to collect some per-regex timings, however trying to run the jquery highlighting benchmark fails because of the `Ability to use \z in a character class` problem. How did you get around that?

TimNN:
So I did some initial benchmarking of the jquery benchmark (measuring how long each regex matching took), the result are in this gist: https://gist.github.com/b6bb756f96b58e52b3299b709fa785dd

`CUM`: is the cumulative time spend on a regex, `AVG`, the average time, the lists are sorted by average time (times are in seconds).

The code is available in the respective branches in the [TimNN/syntect](https://github.com/TimNN/syntect) repo.

The worst offenders by far (based on both, `AVG` and `CUM`) are the following:

```txt
CUM: PT7.310148847S AVG: PT0.000004021S REGEX: [_$[:alpha:]][_$[:alnum:]]*(?=\s*[\[.])
CUM: PT17.180939801S AVG: PT0.000004806S REGEX: ([_$[:alpha:]][_$[:alnum:]]*)(?=\s*\()
CUM: PT21.358124652S AVG: PT0.000008019S REGEX: ([_$[:alpha:]][_$[:alnum:]]*)\s*(\.)\s*(prototype)\s*(\.)\s*(?=[_$[:alpha:]][_$[:alnum:]]*\s*=\s*(\s*\b(async\s+)?function\b|\s*(\basync\s*)?([_$[:alpha:]][_$[:alnum:]]*|\(([^()]|\([^()]*\))*\))\s*=>))
CUM: PT22.522137730S AVG: PT0.000008456S REGEX: ([_$[:alpha:]][_$[:alnum:]]*)\s*(\.)\s*(prototype)(?=\s*=\s*(\s*\b(async\s+)?function\b|\s*(\basync\s*)?([_$[:alpha:]][_$[:alnum:]]*|\(([^()]|\([^()]*\))*\))\s*=>))
```

They seem to match the "best case" mentioned by @raphlinus, which I guess is a good thing?

trishume:
@TimNN awesome thank you! That's definitely useful information since it does indeed match up with the case @raphlinus said could be optimized without too much difficulty. Thanks for the help.

And yes the jQuery benchmark breaks because of a substitution I perform for nonewlines mode. I fixed the benchmark to use line strings with newline characters, but didn't end up committing it, sorry.

TimNN:
So I've been hacking a bit on `fancy-regex` and managed to get the optimisation mentioned above working (at least I think so -- the code is *very* hacky and desperately needs cleaning up).

The results however look very promising so far: On my machine, highlighting jquery went from `1,228,132,968 ns/iter (+/- 63,891,621)` down to `858,410,622 ns/iter (+/- 83,053,993)`, thus an improvement of about 30%.

The code is in the `trailing-la-opt` branch in my `fancy-regex` fork, if you want to give it a go. I'll try to clean it up a bit over the weekend and send a PR to get some feedback from @raphlinus.

trishume:
@TimNN that's awesome! 858ms is still more time than it takes Oniguruma to highlight jQuery on my computer, but my computer also takes less than 1,228ms to highlight with fancy-regex so it is possible that on my computer fancy-regex will be just as fast.

I was hoping it would actually lead to a significant performance *increase* eventually but merely matching the performance of Oniguruma is enough for me to make it the default once the compatibility issues are fixed since it will fix #33 and make all dependencies pure rust.

I may be able to find the time to fix some of the smaller compatibility issues I listed. Specifically the first two unfinished ones listed (I sorted by estimated difficulty). Some of the issues look difficult though, specifically a full expression parser/rewriter for the character class operators.

TimNN:
I ran the jquery benchmark again with the oniguruma version and got `834,134,774 ns/iter (+/- 101,552,378)`, so the patch brings us at least to the same level as oniguruma on my machine for this benchmark.

(Note that my per regex benchmarking code is currently not very efficient since I had planned on collecting more stats than average & total time, so this may slow everything down a bit).

Also, using `RUSTFLAGS=-Ctarget-cpu=native` improved the runtime of the optimised `fancy-regex` version by about 50ms on my machine.

trishume:
@TimNN Excellent. There's probably more optimization possible but that's great for now.

It should be theoretically faster than Oniguruma at least on syntaxes which have been optimized for Sublime's sregex engine to not use many/any fancy regex features, which should allow the `rust` regex crate to do everything, and that engine should be faster than Oniguruma.

I'm not sure even that would match Sublime Text's performance though. I think it uses something like https://doc.rust-lang.org/regex/regex/struct.RegexSet.html to match many regexes at once in time proportional only to the number of characters, but with support for extracting captures and match positions (unlike the `regex` crate).

At the moment `fancy-regex` supports substantially more regexes than Sublime's sregex, which I think only supports the things in the `regex` crate plus the optimization you just implemented for translating lookaheads and possibly lookbehinds. Otherwise it falls back to Oniguruma. But the things that sregex supports should end up almost entirely delegating to `regex` under `fancy-regex`, so the syntaxes optimized for it should be fast under `fancy-regex`.

robinst:
Should we create issues in fancy-regex for the unsupported syntax? Seems like the better place to discuss these.

For fancy character classes (`[a-w&&[^c-g]z]`), it looks like that should be added to the regex crate. There's a comment in the source hinting that adding support for these is planned/welcome:

https://github.com/rust-lang/regex/blob/52fdae7169ec619530985a019184319ac4bbee5a/regex-syntax/src/lib.rs#L1408-L1410

(see [UTS#18 RL1.3](http://www.unicode.org/reports/tr18/tr18-19.html#Subtraction_and_Intersection))

I think that would also help with implementing things such as `\H` in character classes (which is `[^0-9A-Fa-f]`).

trishume:
@robinst Good point. I guess fancy-regex would be a better place.

I created an issue in regex (https://github.com/rust-lang/regex/issues/341), haven't created any in fancy-regex yet though.

BurntSushi:
I would be interested in lending some insight here if you folks wind up seeing bottlenecks inside the regex crate. The regex crate is fast in a lot of cases, but that doesn't mean it's fast in every case, so don't assume that the regex crate will always bail you out. :-) I'll get the ball rolling by throwing some things against the wall and seeing what sticks.

If a regex is particularly large, then it's possible that the DFA will be forced to bail out because it's thrashing its cache. When the DFA bails, it falls back to a much slower (by an ~order of magnitude) regex engine. You can tweak how much cache space is available with the [`dfa_size_limit`](https://docs.rs/regex/0.2.1/regex/struct.RegexBuilder.html#method.dfa_size_limit) option. By default, it's 2MB. The surefire indicator of regexes that might thrash the cache are regexes with large counted repetitions (e.g., `(foo){1000}`) or regexes with lots of Unicode classes. A few Unicode classes here and there aren't going to hurt, but if you combine large classes with counted repetitions, e.g., `\pL{100}`, then you're in for some pain.

`RegexSet` may be a little tricky to use since it is very limited in what it can do. All it can really tell you is which regex matches, but doesn't give you any position information, which means you need re-run the regexes in the set that matched to get that position information. It is worth thinking about and possibly even trying a `RegexSet`, but I wouldn't get your hopes up. :-)

Finally, have you folks seen any problems with performance for compiling the regexes? Does it add any noticeable overhead?

trishume:
@BurntSushi cool thank you. If we get around to optimizing it more than @TimNN already has, we could probably use your advice.

From the very basic benchmarks I did on this branch I didn't see anything noticeable from Regex compilation. It didn't seem to be very different from Oniguruma. I do make sure to compile each regex at most once and only compile them if they are actually needed.

keith-hall:
Hi, I just wanted to make a quick note that I've [submitted a PR to the ST Packages repo](https://github.com/sublimehq/Packages/pull/806) that removes the named backrefs compatibility issue from the Markdown syntax, so, depending whether any other syntaxes use named backreferences, you may get away without this support.

raphlinus:
Some comments.

First, I took a look at @TimNN 's optimization. It's definitely the optimization I had in mind, but is not quite suitable for merging yet (it changes the output, specifically adding more capture groups than were originally present). I am a bit surprised it's only a 30% gain, I would have expected more.

The instrumentation for total time spent, number of invocations, etc., sounds extremely useful, and I recommend that gets checked in. We'll want to track performance on an ongoing basis (assuming we go ahead with fancy-regex, and even then it's extremely useful for making that decision). Where is the time going _after_ the optimization is in place?

In my (admittedly not very thorough) testing, the time impact of regex compilation was minimal. For large files, it's spending seconds computing the highlighting.

What's the secret to super-fast performance? Is it running multiple regexes in parallel? Is it using `RegexSet`? I think the latter is promising. One idea that might be worth pursuing is the idea of an approximate (conservative) pure regex, which is guaranteed to match if the source regex does, but not conversely. This would be useful primarily in pruning the regexes that need to be evaluated. One downside, though: if there's a lot of pushing and popping, then it will run multiple sets over (almost) the entire source line.

trishume:
I just learned something from [this conversation on Reddit](https://www.reddit.com/r/rust/comments/5zit0e/regex_captures_slow_compared_to_python/) with @BurntSushi and @raphlinus that may be part of the cause of the lack of performance gains.

`syntect` currently always does a regex search with captures even when it doesn't actually need them. Given what @BurntSushi said on Reddit, with the `regex` crate (which fancy-regex will call into a lot) this has a substantial performance penalty over just finding the bounds of the match.

I think an optimization where it keeps track of if a certain rule needs the captures or not may help performance with fancy-regex and possibly with Oniguruma as well, although I'm not sure there's much of a penalty to getting captures with Oniguruma, especially since it lets you re-use the capture regions struct between calls.

Not sure exactly how much difference it would make, but it would probably help a bit.

BurntSushi:
> syntect currently always does a regex search with captures even when it doesn't actually need them. Given what @BurntSushi said on Reddit, with the regex crate (which fancy-regex will call into a lot) this has a substantial performance penalty over just finding the bounds of the match.

There's a lot of subtlety here. There are various factors at play:

1. Even if you use the `captures` method call, if the regex itself has no captures, then the regex engine notices this and doesn't run the NFA. See: https://github.com/rust-lang/regex/blob/d813518e2a199884cd38a4e32497a7453db79697/src/exec.rs#L523-L535
2. Does `fancy-regex` ever call the `captures` method? Or does it handle captures on its own?
3. What do your regexes actually look like? (I tried looking once but I got lost. :-()
4. The [performance guide for the regex crate](https://github.com/rust-lang/regex/blob/master/PERFORMANCE.md) is something you should definitely check out.

> although I'm not sure there's much of a penalty to getting captures with Oniguruma

Right. Classical backtracking engines, IME, typically don't impose a penalty for extracting captures.

BurntSushi:
FWIW, there are plans (in my head, anyway) to make capture extraction faster, but I can't commit to a timeline.

robinst:
I [rebased this branch](https://github.com/robinst/syntect/tree/fancy-regex) and have been implementing missing features in fancy-regex, see [my pull requests](https://github.com/google/fancy-regex/pulls?utf8=%E2%9C%93&q=is%3Apr%20author%3Arobinst%20).

Also, the just released [regex 0.2.2](https://github.com/rust-lang/regex/releases/tag/0.2.2) now supports nested character classes and intersections, which means the "Fancy character class syntax `[a-w&&[^c-g]z]`" task is mostly done.

So I think the only task that doesn't have a pull request yet is "Ability to use `\z` in a character class (Using the nonewlines option)", which I haven't looked at yet.

I also found a regex that fancy-regex currently has trouble with (haven't investigated why yet): https://github.com/google/fancy-regex/issues/14

robinst:
I looked into "Ability to use `\z` in a character class (Using the nonewlines option)" now. I'm pretty sure it doesn't work as expected in the current implementation, but it wasn't detected because onig doesn't complain.

With the following:

```rust
let re = onig::Regex::new(r"^a[\z]").unwrap();
println!("{}", re.is_match("a"));
println!("{}", re.is_match("az"));
```

The regex is compiled, but it prints `false`, then `true`. So it's equivalent to `^az`, not `^a\z` (which is what syntect would want).

trishume:
@robinst good catch, I created an issue: #76 and updated the to-do list in this issue.

robinst:
Update: Fixed #76 now. With that, all of the check boxes in the description are done.

I've rebased @trishume's fancy-regex branch here: https://github.com/robinst/syntect/tree/fancy-regex

Now there's the following failing tests left, and they fail on the assertions (instead of panicking while compiling regexes):

    html::tests::strings
    html::tests::tokens
    parsing::parser::tests::can_parse_yaml

The last one I added [here](https://github.com/robinst/syntect/commit/0448adfe2f244c0b986618eae917090b39aba587) and shows that the YAML syntax is not working yet:

```diff
 thread 'parsing::parser::tests::can_parse_yaml' panicked at 'assertion failed: `(left == right)`
-  left: `[(0, Push(<source.yaml>)), (0, Push(<string.unquoted.plain.out.yaml>)), (1, Pop(1)), (1, Push(<string.unquoted.plain.out.yaml>)), (2, Pop(1)), (2, Push(<constant.language.boolean.yaml>)), (3, Pop(1)), (3, Push(<punctuation.separator.key-value.mapping.yaml>)), (4, Pop(1)), (5, Push(<string.unquoted.plain.out.yaml>)), (10, Pop(1))]`,
+ right: `[(0, Push(<source.yaml>)), (0, Push(<string.unquoted.plain.out.yaml>)), (0, Push(<entity.name.tag.yaml>)), (3, Pop(2)), (3, Push(<punctuation.separator.key-value.mapping.yaml>)), (4, Pop(1)), (5, Push(<string.unquoted.plain.out.yaml>)), (10, Pop(1))]`', src/parsing/parser.rs:482:8
```

I had a look at the syntax but it's pretty complex. If someone who knows the syntax wants to track down the problem, that would be cool. (I guess I should learn how to use a debugger for Rust :)).

keith-hall:
I don't know YAML syntax very well, but I cut down the syntax definition to the following, and I think it should still have the same behavior with the `key: value\n` example (it still gets the same scopes in ST as the full YAML syntax def - untested in syntect though, sorry!), so it may help with debugging.

```yaml
%YAML 1.2
---
# See http://www.sublimetext.com/docs/3/syntax.html
scope: source.yaml-test
name: YAML-Test
variables:
  c_indicator: '[-?:,\[\]{}#&*!|>''"%@`]'
  # plain scalar begin and end patterns
  ns_plain_first_plain_out: |- # c=plain-out
    (?x:
        [^\s{{c_indicator}}]
      | [?:-] \S
    )

  _flow_scalar_end_plain_out: |- # kind of the negation of nb-ns-plain-in-line(c) c=plain-out
    (?x:
      (?=
          \s* $
        | \s+ \#
        | \s* : (\s|$)
      )
    )
contexts:
  main:
    - include: block-mapping
    - include: flow-scalar-plain-out

  block-mapping:
    - match: |
        (?x)
        (?=
          {{ns_plain_first_plain_out}}
          (
              [^\s:]
            | : \S
            | \s+ (?![#\s])
          )*
          \s*
          :
          (\s|$)
        )
      push:
        #- include: flow-scalar-plain-out-implicit-type
        - match: '{{_flow_scalar_end_plain_out}}'
          pop: true
        - match: '{{ns_plain_first_plain_out}}'
          set:
            - meta_scope: string.unquoted.plain.out.yaml entity.name.tag.yaml
              meta_include_prototype: false
            - match: '{{_flow_scalar_end_plain_out}}'
              pop: true
    - match: :(?=\s|$)
      scope: punctuation.separator.key-value.mapping.yaml

  flow-scalar-plain-out:
    # http://yaml.org/spec/1.2/spec.html#style/flow/plain
    # ns-plain(n,c) (c=flow-out, c=block-key)
    #- include: flow-scalar-plain-out-implicit-type
    - match: '{{ns_plain_first_plain_out}}'
      push:
        - meta_scope: string.unquoted.plain.out.yaml
          meta_include_prototype: false
        - match: '{{_flow_scalar_end_plain_out}}'
          pop: true

```

robinst:
Thanks @keith-hall! That helped, I've noticed a difference with this pattern (narrowed down):

```rust
let regex = r"(?=\s*$|\s*:(\s|$))";
let s = "key: value";
println!("{:?}", onig::Regex::new(regex).unwrap().find(s));
println!("{:?}", fancy_regex::Regex::new(regex).unwrap().find(s));
```

onig returns `Some((3, 3))`, whereas fancy-regex returns `Some((0, 0))`. Removing the `\s*$|` part of the pattern makes it work. Ran out of time now, but looks like fancy-regex doesn't handle the `$` in the positive lookahead correctly?

robinst:
Ok, tracked down the problem and have a fix here: https://github.com/google/fancy-regex/pull/21

With that fix, `cargo test` works! `syntest` panicks compiling one of the regexes though :).

robinst:
FYI, the regex that currently fails to compile is `(?<=$|.)`. Oniguruma handles it by transforming it to `(?<=$)|(?<=.)`. fancy-regex can do the same. I'm looking into that, but don't have a PR yet.

trishume:
@robinst awesome, great progress, thanks for all the work you've been doing on this. Seems like the goal is in sight.

keith-hall:
Out of interest, why does that cause a panic? Does fancy-regex not support variable-width lookbehind?
I imagine that if I were improving that syntax definition to be compatible with ST's sregex engine, I would transform that pattern to `(?:$|(?!^))` to avoid look-behinds altogether for better performance. But I guess if you need a generic fix for variable-width look-behind then transforming it the way Oniguruma does is the only realistic way to do it.

robinst:
That case is just not implemented yet, yeah. I haven't checked if there's other patterns that have the same problem, but it's something that fancy-regex will need anyway, so I'm looking into adding it.

The pattern comes from here btw (there's another one further down): https://github.com/sublimehq/Packages/blob/master/Clojure/Clojure.sublime-syntax#L834

(And I agree, the pattern can be improved.)

robinst:
PR for supporting look-behind with variable sized alternative in fancy-regex: https://github.com/google/fancy-regex/pull/25

robinst:
So with the latest two fixes (https://github.com/google/fancy-regex/pull/28 and https://github.com/google/fancy-regex/pull/29), all the regexes now compile 🎉.

But, there's still quite a few assertion failures when running syntest, here's a diff of running it on master (old) against fancy-regex (new): https://gist.github.com/robinst/5b18d47cf45697bacd827fee845795bb#file-syntest-onig-to-fancy-diff

So now we have to track down why they fail.

keith-hall:
I had a quick look at the errors from `./testdata/Packages/C#/tests/syntax_test_Generics.cs`, as there aren't many and they all seem related and discovered that the following test, if added to `fancy-regex`, fails:

```rust
assert_match(r"$\n", "\n");
```

robinst:
Thanks Keith, that's very helpful! Will look into those.

robinst:
Had a quick look, both have simple causes:

`$` means end of text (as opposed to end of line) in the regex crate, and in fancy-regex too. We can consider changing it in fancy-regex, but I'm not sure it's a good idea to deviate too much from the regex crate.
We can instead compile regexes in syntect with `(?m)` prepended, to enable multi-line mode.

`[[:alpha:]]` and others are ASCII-only in the regex crate. I'm not sure why they work with onig, as they require Unicode mode, but haven't looked into that yet. 
Anyway, I think they should all be changed in the syntaxes to use `\p` style classes instead, which are unambiguously Unicode in all engines. In the meantime, we could just do replacement in syntect.

robinst:
Ok, I've done the fixes in syntect here: https://github.com/robinst/syntect/tree/fancy-regex

And another fix for fancy-regex: https://github.com/google/fancy-regex/pull/30

Now the diff looks much better: https://gist.github.com/robinst/5b18d47cf45697bacd827fee845795bb

keith-hall:
Excellent work @robinst! I wonder what is causing the remaining differences/failures. It's possible that a lot of them are arising because something earlier on is being matched differently, and thus being in a different context for the later assertions/lines. So I think it's best to try to look at the first failures in each syntax for now to save barking up the wrong tree - I'll try to help investigate if I get some spare time.

robinst:
FYI I've narrowed the failure in Markdown (line 12) to this:

Finding the regex `r"(?m)^\s*$"` in the input `"test\n"`, Oniguruma doesn't match, but fancy-regex returns position 5. Same problem if we simplify the regex to `r"(?m)^$"`.

Note that fancy-regex just delegates to the regex crate for this, so I'll either need to try to fix it there or we work around it. I'll open an issue in the regex crate first.

keith-hall:
In terms of a workaround, would it help if syntect replaces `^` with `\A` and `\n` with `\n\Z`?

robinst:
Yeah. I think the way to work around it would be to replace `^` with `\A` but keep `$` as-is and keep multi-line mode. `\A` should work because we're only matching on individual lines anyway. We can't use `\n\Z` because that wouldn't work on the last lines of files that are not terminated by newlines.

I'll give that a try. It means we'll also have to parse the regex for that case because we don't want to replace all the `^` characters in the pattern.

robinst:
Another fix: https://github.com/google/fancy-regex/pull/36

And with replacing `^` with `\A`, we're now really close :): https://gist.github.com/robinst/5b18d47cf45697bacd827fee845795bb

trishume:
@robinst that's awesome! Maybe if I find myself with some free time I'll try to fix the last few bugs that cause ASP syntax test failures. That should make it easier to figure out which problems are syntect and which are fancy-regex. In the mean time I guess you can look at the Markdown failures.

robinst:
FYI, I've narrowed down the failure to this difference between onig and fancy-regex:

```rust
let regex = r"(?!`(?:[^`]+(?=`)|x)`)";
let s = "`a`";
println!("{:?}", onig::Regex::new(regex).unwrap().find(s));
println!("{:?}", fancy_regex::Regex::new(regex).unwrap().find(s));
```

onig returns `(1, 1)` but fancy-regex returns `(0, 0)`. Haven't tracked down the bug in fancy-regex yet.

robinst:
Found the problem for the above, fix here: https://github.com/google/fancy-regex/pull/42

That fixes the Markdown failures: https://gist.github.com/robinst/5b18d47cf45697bacd827fee845795bb

It's interesting that some tests now have fewer failures with fancy-regex. But that could also be related to the changes in how we transform regexes.

I'll probably look into the Java failures next.

trishume:
@robinst Wow, awesome work! If there's less failures with fancy-regex then Oniguruma now, then that sounds like there's only two things left to do before we can make fancy-regex the default or only option:

- [ ] Do a bunch of benchmarking to make sure it isn't a performance regression. Both posting some results here and if you (or anyone else) have your own corpus/server you'd like to try it out on and you tell me how the speed is I'll take your word for it. I'm willing to merge as long as it isn't a significant regression, if necessary we can work on making it faster later.
- [ ] Test for backtracking explosions that Oniguruma doesn't hit, I seem to recall we found some of these before. Again anyone with a private corpus/service can be helpful here.
- [ ] I review your branch before merging, but that should be fairly easy.

BurntSushi:
If y'all come across performance problems in the `regex` crate, I'm happy to diagnose them if there's a simple example I can run. :-)

keith-hall:
[this regex](https://github.com/sublimehq/Packages/blob/18a2d34340d9a946246ed5baf98dc9b4f390d95b/C%23/C%23.sublime-syntax#L1576) seems to cause catastrophic backtracking with fancy-regex, and takes 30 minutes to find a match against [this line](https://github.com/sublimehq/Packages/blob/18a2d34340d9a946246ed5baf98dc9b4f390d95b/C%23/tests/syntax_test_GeneralStructure.cs#L1123). I wonder how feasible it would be for fancy-regex to automatically unroll the loop - i.e. convert `(?m)([^}"\\]+(\\.)*)+(?="(?!"))` to `(?m)((?:[^}"\\]+)(?:(\\.)+(?:[^}"\\]*))*)(?="(?!"))`, or whether some other mitigation would be a better solution.

robinst:
@keith-hall I tried reproducing that but I can't, it runs fine for me. All of `syntest` completes in reasonable time here. Are you running [my branch](https://github.com/robinst/syntect/tree/fancy-regex)? Did you run `make packs` before? The problem sounds similar to google/fancy-regex#14.

@trishume Yeah that sounds good to me. I'd also like to polish fancy-regex a bit before:

* It needs a good test suite. I was thinking of parsing [Oniguruma's](https://github.com/kkos/oniguruma/blob/master/test/test_utf8.c#L146), but haven't looked into that yet. We'd probably need to ignore some of those tests because of different behavior, but maybe it's a good base. If someone knows a good test suite (that includes lookaround assertions), that'd be interesting.
* The API should be changed to follow the API changes that happened in the regex crate. I think it was modeled on 0.1 regex crate API.
* It needs good documentation for the things where the behavior differs from other engines like Oniguruma. The differences mostly come from being based on the regex crate.

keith-hall:
yes, running your branch - tried `make packs` with the fixed C# syntax (`https://github.com/sublimehq/Packages/pull/1359`) and without and seems to happen in both cases when running `syntest`.

however trying `fancy-regex` directly seems to work fine/fast, so not sure what's going on, maybe something in `syntect` I can try to track down.
`cargo run --example toy run '(?m)([^}"\\]+(\\.)*)+(?="(?!"))' '            Console.WriteLine("{0,-20} {1,5:N1}", names[ctr], hours[ctr]);'`

trishume:
@keith-hall In the past I've inserted code that prints every regex before it tries to search for it, and then when it hangs the last thing printed is the cause.

keith-hall:
it's definitely this regex - I updated the parser to show timings:

```diff
--- a/src/parsing/parser.rs
+++ b/src/parsing/parser.rs
@@ -216,8 +216,13 @@ impl ParseState {
                     } else {
                         match_pat.regex.as_ref().unwrap()
                     };
+                    use std::time::Instant;
+                    let before = Instant::now();
                     // TODO don't panic on regex error
                     let matched = regex.captures_from_pos(line, *start).unwrap();
+                    let duration = Instant::now().duration_since(before);
+                    println!("regex {:?} took {:?} to find {} starting at position {:?}", regex, duration, if matched.is_some() { "a match" } else { "nothing" }, start);
+
                     if let Some(captures) = matched {
                         let match_start = captures.pos(0).unwrap().0;
                         let match_end = captures.pos(0).unwrap().1;

```

and this was the output:

```
regex ([^}"\\]+(\\.)*)+(?="(?!")) took Duration { secs: 1683, nanos: 255688848 } to find nothing starting at position 44
```

which is 28 minutes, and shows that it starts looking for a match at position 44, which is probably why I couldn't reproduce the problem by testing the pattern in `fancy-regex` while passing the whole line to match against.

robinst:
Ok, so I can reproduce it with (note the 44 at the end of the command, to start from that position):

    cargo run --release --example toy run '(?m)([^}"\\]+(\\.)*)+(?="(?!"))' \
    '            Console.WriteLine("{0,-20} {1,5:N1}", names[ctr], hours[ctr]);' 44

or:

    cargo run --release --example toy run '(?m)([^}"\\]+(\\.)*)+(?="(?!"))' ':N1}", names[ctr], hours[ctr]);'

It takes 25 seconds here, which is too slow. Not sure if you were running with `--release` or not, but `--release` makes a much bigger difference with fancy-regex than with onig (because all the regexing happens in debug mode).

robinst:
@keith-hall I've investigated this now, see my comment here: https://github.com/google/fancy-regex/issues/14#issuecomment-364346794

Summary: Oniguruma happens to not run into catastrophic backtracking because it has an optimization which makes it check if the input contains a `"` (from the look-ahead) before it does any matching.

I don't think we'll implement optimizations such as that in fancy-regex in the short term (and it only helps in some cases), although it would be an interesting project.

Can we change the regex instead?

keith-hall:
It'd be great if this PR could be updated with the latest changes (from @robinst's `fancy-regex` branch and the fixes from `master`), if anyone has time to resolve the merge conflicts :)

robinst:
@keith-hall: I rebased my branch here: https://github.com/robinst/syntect/tree/fancy-regex

If @trishume is ok with it, I can also force-push that to the branch that this PR is based on.

trishume:
@robinst yah a force push sounds good

robinst:
Done! Note that you might have to `cargo update` locally before trying to run this (because `Cargo.lock` is not checked in). Updates:

* A [small change](https://github.com/google/fancy-regex/pull/47) to make fancy-regex work with a newer regex crate.
* A [bug fix in syntect](https://github.com/trishume/syntect/pull/34/commits/84059f4b9958e962bcb692d00008bac8a18325a4) for code that I wrote a couple of days ago that fails with fancy-regex (but is fine with Oniguruma). `+= 1` on a string index is almost never the right thing to do. I walked into this one before, but apparently forgot about it again :).

keith-hall:
small note for people [struggling to get `syntect` to build on Windows](https://github.com/trishume/syntect/issues/48#issuecomment-287450655): using the branch from this PR, you can edit `Cargo.toml` to remove the `onig` dependency, and everything will work fine.

Side note: probably this PR should be updated so that `fancy-regex` is an optional dependency as part of the `parsing` feature like `onig` is, no?

```diff
--- Cargo.toml
+++ [new] Cargo.toml
@@ -15,7 +15,7 @@
 
 [dependencies]
 yaml-rust = { version = "0.4", optional = true }
-onig = { version = "3.2.1", optional = true }
+#onig = { version = "3.2.1", optional = true }
 walkdir = "2.0"
 regex-syntax = { version = "0.4", optional = true }
 lazy_static = "1.0"
@@ -25,7 +25,7 @@
 flate2 = { version = "1.0", optional = true, default-features = false }
 fnv = { version = "1.0", optional = true }
 regex = "*"
-fancy-regex = { git = "https://github.com/google/fancy-regex.git" }
+fancy-regex = { git = "https://github.com/google/fancy-regex.git", optional = true }
 serde = { version = "1.0", features = ["rc"] }
 serde_derive = "1.0"
 serde_json = "1.0"
@@ -51,7 +51,7 @@
 # Pure Rust dump creation, worse compressor so produces larger dumps than dump-create
 dump-create-rs = ["flate2/rust_backend", "bincode"]
 
-parsing = ["onig", "regex-syntax", "fnv"]
+parsing = ["fancy-regex", "regex-syntax", "fnv"]
 # The `assets` feature enables inclusion of the default theme and syntax packages.
 # For `assets` to do anything, it requires one of `dump-load-rs` or `dump-load` to be set.
 assets = []
```

robinst:
@keith-hall Pushed a commit with those changes, thanks!

src/parsing/syntax_definition.rs
```diff
@@ -236,20 +249,15 @@ impl MatchPattern {
...
+        fancy_regex::Regex::new(&self.regex_with_substitutes(region, s)).unwrap()
     }
 
     fn compile_regex(&mut self) {
         // TODO don't panic on invalid regex
-        let compiled = Regex::with_options(&self.regex_str,
-                                           RegexOptions::REGEX_OPTION_CAPTURE_GROUP,
-                                           Syntax::default())
-            .unwrap();
+        println!("compiling {:?}", self.regex_str);
```
keith-hall: should this `println` be here, as it generates a lot of noise? if it's useful for debugging, maybe it would be best to hide it behind a feature flag as discussed at https://github.com/trishume/syntect/pull/146#discussion_r183709427

robinst: @keith-hall I pushed a commit that changes this to only print in case it fails.

kornelski:
It would be great if that change landed, because [libgit2 crashes in binaries linking oniguruma](https://github.com/libgit2/libgit2/issues/4738), and I'd like to use both libgit2 and syntect together.

Keats:
Is https://github.com/google/fancy-regex/issues/44 the only blocker left for this?

keith-hall:
I was also wondering about this. If we don't want to default to fancy-regex yet, maybe it would be worth maintaining both the oniguruma and fancy-regex code paths for now and consumers can choose which one to use from a feature flag. It may then be easier to do performance comparisons etc.

src/parsing/syntax_definition.rs
```diff
@@ -5,12 +5,13 @@
 //! into this data structure?
 use std::collections::{BTreeMap, HashMap};
 use std::hash::Hash;
-use onig::{Regex, RegexOptions, Region, Syntax};
+use fancy_regex;
```
pickfire: Why not continue using `fancy_regex::Regex` here?

robinst:
> Is google/fancy-regex#44 the only blocker left for this?

@Keats Yes, unless we find another one.

@keith-hall A feature sounds like a good idea, yeah. Maybe we can abstract the regex compilation and matching parts a bit to make the feature less painful to maintain (so that it's just in one module and not all over the place).

I might have some time this week to work on this.

src/parsing/syntax_definition.rs
robinst: You mean why not write `use fancy_regex::Regex;` here? Yeah there's not really a good reason. I will change it next time I work on this.

src/parsing/yaml_load.rs
```diff
@@ -407,6 +412,76 @@ impl SyntaxDefinition {
     }
 }
 
+
+/// In fancy-regex, POSIX character classes only match ASCII characters.
+/// Sublime's syntaxes expect them to match Unicode characters as well, so transform them to
+/// corresponding Unicode character classes.
+fn replace_posix_char_classes(regex: String) -> String {
```
pickfire: Is there by any chance that we are able to do the sublime syntax replacement before run-time?

OptimisticPeach:
Sorry to poke at an old issue, but is there any news on this? I'd really like to use `syntect` in a WASM project, but according to #135 this needs to land first.

trishume:
Unfortunately this is not quite complete and based on an older release of syntect. It would probably take a fair amount of work to complete. I'm not personally interested in doing that work unfortunately, so unless someone else steps up to do it, it's not on the roadmap.

OptimisticPeach:
Ah that's a bummer, I'm sorry for the letdown but I don't think I'd be able to commit to making it happen right now.

robinst:
Hey, just an update. I'm working on this again, but with a different approach.

I'm moving all the regex usage to a module first, so then we can swap out the implementation using a cargo feature.

I'll have a pull request next week :).

robinst:
Ok, see #270. I think we can close this PR and move discussion to #270.

Issue #36, opened 2017-03-02
Support advanced scope selectors


keith-hall:
Currently, syntect doesn't support grouping parens or `&` operators in scope selectors, which means it doesn't color scopes correctly when the `.tmTheme` file being used contains these more advanced constructs.

See the update under "One last twist" at http://blog.macromates.com/2005/introduction-to-scopes/ for a brief example
> it's also possible to AND, OR, and subtract scope selectors, e.g.: `(a | b) & c - d` would select the scope which is not matched by d, and matched by both c, and a or b.

I've never really found any good documentation on scope selectors, so I tried to write a bit more about them on SO, which I'll link to here in case it's useful: http://stackoverflow.com/a/37407588/4473405

As a simple example, if we imagine the following Python code:

```python
def test:
    global abc
```

then the selector `(storage | entity.name) & meta.function` should match `def` and `test`.

Please would you consider adding support for this to your awesome syntect library.

trishume:
I agree it's a useful and neat feature, but I must admit I don't predict I'll find time or motivation to do it myself. Other functionality like #34 seems more important in that it enables new use cases and prevents basic usage from failing horribly (see #33). This functionality just allows marginally nicer themes.

If someone else (like you) wants to take up the task though I'm happy to provide advice, code review, guidance and information.

FichteFoll:
For the record, it appears that #39 added support for pipes (same as commas) and multiple exclusions. What remains are parentheses for sub-matches and ampersands for AND-connections, as far as I can see. Maybe add a checklist to the OP?

keith-hall:
Correct - I've now updated the OP with the checklist, as suggested.

keith-hall:
Note that the Monokai color scheme that Sublime Text (build 3156) ships with includes the following scope selector:

```json
{
    "name": "markup punctuation",
    "scope": "text & (punctuation.definition.italic | punctuation.definition.bold | punctuation.definition.raw | punctuation.definition.link | punctuation.definition.metadata | punctuation.definition.image | punctuation.separator.table-cell | punctuation.section.table-header | punctuation.definition.constant)",
    "foreground": "color(var(white) alpha(0.67))"
},
```

so if people were to try to use that color scheme in `syntect` (currently impossible, because `syntect` doesn't support the `sublime-color-scheme` format), then their code/text could look different to how it does in Sublime.

brupelo:
[...]

PR #39, opened 2017-03-08
Support more advanced scope selectors

keith-hall:
This PR partly addresses issue #36, and adds support for:
- empty scope stacks in scope selectors, which always match with a score of 1 i.e. a selector of `- a.b` will match any scope stack that doesn't match `a.b`
- using `|` as an alias for `,` when parsing scope selectors from strings
- multiple excludes in a single scope selector i.e. `a.b - c.d - e.f`, which says match `a.b` unless `c.d` matches or `e.f` matches

I plan to build on this in a future PR (unless, of course, someone else beats me to it!), to add an "example" that will execute [ST syntax tests](http://www.sublimetext.com/docs/3/syntax.html#testing), where the above added scope selector features are used often. This could potentially be used to facilitate continuous integration, whereby ST packages that contain syntaxes can run the syntax tests automatically when a PR is made (or updated with new commits).

I deliberately only touched the `highlighting/selector.rs` file, and not `parsing/scope.rs` to hopefully not affect performance too much. This is my first time working with Rust, so if there are better/more performant ways of achieving the same goals, I'd gladly accept any tips.


trishume:
This is awesome! Thanks for getting this done. I definitely like the approach of doing things in `selector.rs`. When I first saw this performance was my main concern, but none of the changes look like they should affect it much. Also the extensive testing is great.

Being able to run syntax tests is a really compelling reason. I didn't know they made use of fancy selectors much. I've thought about how awesome it would be to have `syntect` run Sublime's syntax tests, but never got around to implementing it.

I don't have time to do a proper review of this right now, but cursory review looks good, I'll try and do a full review later today.

trishume:
Looks great as is. 

Your Rust looks great, you even figured out tricky sigil storms like `&[',', '|'][..]`. You seem to have found all the right handy `Vec` and string functions to use. I don't think I would have done anything differently. Impressive for your first time using Rust.

trishume:
I just pushed 1.2.0 to crates.io with this and your other PR.

PR #41, opened 2017-03-09
Generalize gendata to make it more useful to API-users

Byron:
'gendata' now does some trivial argument parsing to support
different source directories to be specified.
That way, customizations can be more easily built.

A Makefile was added to keep track of the arguments needed
to build the default assets.

To prevent regressions, travis also builds assets.

#### Motivation

I want to use syntect in [`tokio-cassandra`s](https://github.com/nhellwig/tokio-cassandra) CQL shell, and want to add just CQL syntax descriptions in the most efficient format possible.

Thus this might be just the first of more PRs as I merely start to discover what this fantastic crate has to offer :)!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,5 +17,6 @@ script:
       export LD_LIBRARY_PATH="${p}:${LD_LIBRARY_PATH}"
     fi
   - cargo test
+  - make assets
   - cargo doc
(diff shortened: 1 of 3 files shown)


trishume:
This is great. Probably useful for all kinds of users. Thank you.

I fixed a merge conflict with your other pr (travis config) and now I'm just waiting for Travis before I merge. Ping me if I forget...

trishume:
I just pushed `1.2.0` to crates.io with this and your other PR.

PR #42, opened 2017-03-09
Add 'html' and 'assets' cargo features

Byron:
That way, users of this library can remove support for highlighting
that is not needed for them.

Binaries will be smaller (e.g. no static asset data), and
compile times will be reduced as well.

This commit is motivated by the `tokio-cassandra` shell, which will
use only use two syntaxes, YAML + SQL/CQL, and a handful of themes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,5 +17,9 @@ script:
       export LD_LIBRARY_PATH="${p}:${LD_LIBRARY_PATH}"
     fi
   - cargo test
+  - rm -Rf examples 
   - cargo doc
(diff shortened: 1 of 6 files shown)


trishume:
Looks good. Thanks for adding testing for it to Travis.

Not necessary but in case you already have it: do you have any data on how much this reduces binary sizes?

Byron:
Thank you!

My guess is that the assets will weigh in the binary as much as they weigh on disk, maybe even more. In number that's ~500kb.

PR #43, opened 2017-03-10
Add "yaml-load" feature flag

Byron:
Default on.
If off, there will be no usage of yaml-rust, which turns out to
be problematic if you are already using it with different feature
flags (i.e. linked hashmap vs. btreemap).

This also means the only way to create a syntax set is using the
functions from `dumps`, and it might be nice to have one of these
directly in the SyntaxSet type itself.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ exclude = [
 ]
 
 [dependencies]
-yaml-rust = "0.3.5"
+yaml-rust = { version = "0.3.5", optional = true }
 onig = "1.2.0"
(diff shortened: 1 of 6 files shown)


Byron:
In case you are interested, this is how it currently looks like when using it from [tokio-cassandra](https://github.com/nhellwig/tokio-cassandra/blob/master/cli/Cargo.toml#L29).

trishume:
Cool, this looks fine to me as is. I don't see too much of a need to replicate the dumps functions onto SyntaxSet, at least for now.

Issue #37, opened 2017-03-06
Panic: Can only iterate patterns after linking: Named("types") with C# file


robinst:
Reproduce on master with 1748efe3685b5f1c91d315c84e9daede559fe9f0:

```
$ cat bug.cs
public class Foo
{
    private static void Bar()
    {
        baz.Sort((a, b) => b.depth.CompareTo(a.depth));
    }
}
```

    $ cargo build --example synstats && RUST_BACKTRACE=1 ./target/debug/examples/synstats bug.cs

Output:

```
################## Files ###################
bug.cs
thread 'main' panicked at 'Can only iterate patterns after linking: Named("types")', src/parsing/syntax_definition.rs:137
stack backtrace:
   1:        0x105c130ea - std::sys::imp::backtrace::tracing::imp::write::hd3b65cdfe843284c
   2:        0x105c14e8f - std::panicking::default_hook::{{closure}}::hf2b7428652613d83
   3:        0x105c14b37 - std::panicking::default_hook::h5da8f27db5582938
   4:        0x105c152f6 - std::panicking::rust_panic_with_hook::hcef1e67c646c6802
   5:        0x105c15194 - std::panicking::begin_panic::hc2e8ca89533cd10d
   6:        0x105c150b2 - std::panicking::begin_panic_fmt::h60990696c3c3a88d
   7:        0x105bd40e8 - <syntect::parsing::syntax_definition::MatchIter as core::iter::iterator::Iterator>::next::h0fc92fde02d6dd7f
   8:        0x105bdc6ea - syntect::parsing::parser::ParseState::parse_next_token::h7b5d8225fe16d480
   9:        0x105bdbf28 - syntect::parsing::parser::ParseState::parse_line::hde79bb6ed3c527e5
  10:        0x105b66cbd - synstats::count::h5c86d3300a80813b
  11:        0x105b6739e - synstats::main::h761e13430343e881
  12:        0x105c160ea - __rust_maybe_catch_panic
  13:        0x105c15696 - std::rt::lang_start::h87cb84a8b6cb187e
  14:        0x105b67839 - main
```

The `Sort` line is the problematic one. The original file had a lot more code in it, but removing that line makes it work.

trishume:
Darn. I think the C# syntax at some point says to push a "types" scope which doesn't exist, and I don't handle that case properly so it panics. It should just silently do its best because syntax highlighting shouldn't crash.

[...]

PR #44, opened 2017-03-10
added syntest example to run ST syntax tests

keith-hall:
This PR adds a new example called "syntest", which will parse and execute ST's "syntax_test_" files.

Currently, there is no way to reference a syntax definition in a SyntaxSet from it's original file path, which is how tests specify which syntax definition to use, so it just lets syntect choose based on the extension of the test file at the moment.

I used the regular "regex" crate to achieve the test line parsing - feel free to replace this with "onig" if you don't want the extra reference - I don't quite have enough experience yet to know how to use it effectively.

Also, if the syntax test files use Windows line endings, it has to replace "\r" with nothing, because otherwise the regular expressions  don't match expressions like "$\n?" properly, but perhaps it would be better to update the main parsing module to fix this?

I've checked and it shows success on the XML, HTML and JSON syntax tests. I also tried modifying them to cause a test to fail, and syntest correctly picked that up.

However, it seems to detect a failure in the Haskell syntax test file, and it looks like some meta scopes are being applied multiple times - so I'd appreciate if you could take a look and see if it's a flaw in my syntest logic somewhere or a bug in the parser. Thanks!

```
cargo run --example syntest testdata/Packages/Haskell/
    Finished debug [unoptimized + debuginfo] target(s) in 0.0 secs
     Running `target/debug/examples/syntest testdata/Packages/Haskell/`
Testing file testdata/Packages/Haskell/syntax_test_haskell.hs
The test file references syntax definition file: Packages/Haskell/Haskell.sublime-syntax
Assertion selector "- comment.line.double-dash.haskell" from line 6 failed on line 5, column range 0-1 (with text ['2']) has scope [<source.haskell>, <comment.line.double-dash.haskell>, <comment.line.double-dash.haskell>, <comment.line.double-dash.haskell>, <comment.line.double-dash.haskell>, <constant.numeric.haskell>]
Ok(FailedAssertions(1, 422))
```


keith-hall:
I vote for something fishy in `easy:ScopeRegionIterator`, because if I do `println!("{:?}", &ops);` after `state.parse_line`, I get what I would expect, but when I `println!("{:?}", op);` inside the loop, I am missing the last `pop`:

```
[(0, Push(<source.haskell>)), (0, Push(<comment.line.double-dash.haskell>)), (0, Push(<punctuation.definition.comment.haskell>)), (2, Pop(1)), (57, Pop(1))]
Noop
Push(<source.haskell>)
Push(<comment.line.double-dash.haskell>)
Push(<punctuation.definition.comment.haskell>)
Pop(1)
[]
```

keith-hall:
The following change seems to help:

```diff
--- syntect-master/src/easy.rs
+++ syntect-syntest/src/easy.rs
@@ -148,10 +148,9 @@
 impl<'a> Iterator for ScopeRegionIterator<'a> {
     type Item = (&'a str, &'a ScopeStackOp);
     fn next(&mut self) -> Option<Self::Item> {
-        let next_str_i = if self.index >= self.ops.len() {
-            if self.last_str_index >= self.line.len() {
-                return None;
-            }
+        let next_str_i = if self.index > self.ops.len() {
+            return None;
+        } else if self.index == self.ops.len() {
             self.line.len()
         } else {
             self.ops[self.index].0
```

with this change, the "Batch File" and "Haskell" syntax tests start working, and the number of failures in "PHP" are reduced significantly.

examples/syntest.rs
```diff
@@ -0,0 +1,247 @@
...
+                    SyntaxTestFileResult::Success(total_assertions)
+                }
+            )
+        },
+        None => Err(SyntaxTestHeaderError::MalformedHeader)
+    }
+}
+
+fn main() {
+    let ss = SyntaxSet::load_defaults_newlines(); // note we load the version with newlines
```
trishume: For this case it might be good to use `SyntaxSet::load_from_folder(argv[0])` so that you can use it for syntax development without recompiling every time. Once we get `syntect` to a point where you can actually rely on it to do things correctly that is...

trishume: Actually since this is using the version with newlines included and `load_from_folder` defaults to without newlines. You can just do what `load_from_folder` does yourself:

```rust
        let mut ss = SyntaxSet::new();
        ss.load_syntaxes(args[0], true).unwrap();
        ss.link_syntaxes();
```

examples/syntest.rs
```diff
@@ -0,0 +1,247 @@
...
+fn recursive_walk(ss: &SyntaxSet, path: &str) -> i32 {
+    let mut exit_code: i32 = 0;
+    let walker = WalkDir::new(path).into_iter();
+    for entry in walker.filter_entry(|e|e.file_type().is_dir() || is_a_syntax_test_file(e)) {
+        let entry = entry.unwrap();
+        if entry.file_type().is_file() {
+            println!("Testing file {}", entry.path().display());
+            let result = test_file(&ss, entry.path(), true);
+            println!("{:?}", result);
+            if let Err(_) = result {
```
trishume: It would be great if there were a flag/argument to make it not die on the first failing file. It would be easier to get a sense of how much was broken then.

Cargo.toml
```diff
@@ -24,6 +24,7 @@ rustc-serialize = "^0.3"
 bincode = "0.6"
 flate2 = "^0.2"
 fnv = "^1.0"
+regex = "0.2.1"
```
trishume: For now the use of `regex` is fine. I may port it to onig before or after it is merged.

However, it should be under `dev-dependencies` since it is only used in an example. See https://github.com/rust-lang/cargo/issues/523

examples/syntest.rs
keith-hall: it currently doesn't die on the first failing file, it just ensures that when it has run through all the files, it will exit with an appropriate exit code :) maybe we could add a flag though so that it optionally can fail fast, to save time (not that these syntax tests take long at all!) when all the caller wants to know is whether there were any failures

trishume: Oh woops. I just skimmed over the code briefly and missed that this didn't actually stop anything.

I think the behaviour you describe is perfect for now. I don't see any pressing need to add a fail-fast flag.

examples/syntest.rs
keith-hall: great idea! however, in my testing using `cargo run --example syntest testdata/Packages/HTML/`, `args[0]` refers to the example in the `target/debug` folder, and it would have to traverse up a few folder levels to find the `testdata/Packages` folder where the syntaxes are, so we will need to come up with a suitable workaround I think.

trishume: Sorry I meant `args[1]` and then you would pass `testdata/Packages` as the first argument and it would run tests for all languages, although it's kinda nice to be able to run tests for only one language I guess. Maybe have the first command line argument be the same as now, and then if there's a second command line argument load from that folder, otherwise just load the defaults.

keith-hall: I was thinking something along those lines too :) thanks, I'll give it a go

keith-hall: done :)

keith-hall:
It seems like the test failures are due to scopes from `meta_content_scope` being missing in the scope stack.
For example, Java:
- https://github.com/sublimehq/Packages/blob/5504b84ee219ad5e2e84523d9affd77fb8935058/Java/Java.sublime-syntax#L231
- https://github.com/sublimehq/Packages/blob/5504b84ee219ad5e2e84523d9affd77fb8935058/Java/syntax_test_java.java#L20

> Testing file testdata/Packages/Java/syntax_test_java.java
The test file references syntax definition file: Packages/Java/Java.sublime-syntax
Assertion selector "meta.class" from line 20 failed on line 19, column range 7-12 (with text ['c', 'l', 'a', 's', 's']) has scope [<source.java>, <storage.type.java>]

trishume:
@keith-hall interesting. Thanks for the diagnosing. I'm going to see if I can review this PR later today and maybe fix some of the bugs it found. If not today I'll try and do it tomorrow.

keith-hall:
looking at the Java example a bit more closely, it might be related to how `set` interacts with meta scopes - in ST, `meta_content_scope` and `meta_scope` apply to the match that `set`s a different context

trishume:
@keith-hall I'm working on reviewing this. Just a heads up I pushed a commit to the branch refactoring the looking up of syntaxes by path (thanks for enabling maintainer commits). You may want to `git pull`.

keith-hall:
@trishume nice one, that's a much better solution than what I had, thanks! :)

trishume:
@keith-hall reviewed the fix to `ScopeRegionIterator` and committed the cleanup I did while doing so. It looks good, thanks for the debugging. You may want to pull.

Now all that's left for me to review is `syntest.rs`.

trishume:
@keith-hall I just reviewed the `syntest.rs` file and it looks good. I pushed a commit to rearrange some error handling code and fix a few small things. I recommend you review the whitespace-insensitive diff to see what actually changed and what was just me changing the indentation of a large segment: https://github.com/trishume/syntect/pull/44/commits/3e95235f4a7335793bfad7d552bb01c57c25f3bb?w=1

I've now reviewed everything in this PR and am pretty happy with it. There's still the bugs that keep the tests from actually passing, but I'm thinking I/we can fix those in separate PRs. I'm ready to merge this now if you are. If you're ready to merge I think I'll do an interactive rebase to remove the dump changes to the first commit that changes them, just to avoid increasing the repo size unnecessarily.

After merging we can create issues for all the bugs we find. The one that makes the C# test suite crash at least already has an issue: #37 

I'm currently looking at why the ASP test has so many failures. I think walkdir might actually be giving us the tests in a different order (possibly because of different OSs, I'm on OSX). Mine runs ASP (with many failures), then Batch (success), then C# (success), then another C# test (crash #37). The ASP failures may or may not be the same underlying bug as you are encountering with Java.

trishume:
Also I notice that Travis is failing, but only the nightly one is failing, the stable one passes. And nightly fails on `cargo --version` so I think it's just Rust nightly that's broken on Travis.

keith-hall:
@trishume yep looks good, I'm ready to merge - I agree that it makes sense to have separate PRs for the remaining issues :) thanks for the cleanup and associated commentary - makes it super easy to follow along :) it's a pleasure working with you!

do you want me to create the issue for the Java failures? I was thinking that it might make sense to tackle that first, then everything related to the same underlying issue will go away and make it easier to investigate what (if anything) else is wrong...

I was curious about those Travis failures - what's the benefit of building against nightly, if I may ask, when it's so unstable? :)

robinst:
Wow, nice change! After all the bugs are fixed, are you planning to run this as part of regular tests?

trishume:
@robinst yah I'm thinking maybe move it to a module and then use it from both the example and `cargo test`.

keith-hall:
@trishume FYI, I've fixed the Java failures over at https://github.com/forkeith/syntect/tree/set, and it reduces the number of ASP test failures quite drastically too. I'm not sure if I will get it to a state where it is Pull Request-able or not, but wanted you to see my progress so we can collaborate together rather than work separately to achieve the same thing.

trishume:
@keith-hall awesome. I'll take a look at it.

Don't worry about duplicating work in the next few days. I'll be busy with school work until this weekend so I won't be doing any work on syntect (other than Githubbing).

Issue #45, opened 2017-03-13
"clear_scopes" restore too early on "pop"


keith-hall:
Imagine the following `clear_scopes_test.sublime-syntax` file:

```yaml
%YAML 1.2
---
# See http://www.sublimetext.com/docs/3/syntax.html
name: Simple Clear Scopes Test
scope: source.example
contexts:
  main:
    - match: \'
      scope: punctuation.definition.string.begin.example
      push: single_quoted_string
  single_quoted_string:
    - meta_scope: string.quoted.single.example
    - match: 'test'
      scope: example.pushes-clear-scopes.example
      push:
        - clear_scopes: 1
        - meta_scope: example.meta-scope.after-clear-scopes.example
        - match: 'foo'
          scope: foo
          push:
            - clear_scopes: 1
            - meta_scope: example.meta-scope.cleared-previous-meta-scope.example
            - match: 'bar'
              scope: bar
              pop: true
        - match: 'test'
          scope: example.pops-clear-scopes.example
          pop: true
    - match: '\\.'
      scope: constant.character.escape.example
    - match: \'
      scope: punctuation.definition.string.end.example
      pop: true
```

and the following `syntax_test_clear_scopes.txt` file:

```
# SYNTAX TEST "clear_scopes_test.sublime-syntax"
'hellotest
# ^^^^ string.quoted.single.example - example
#     ^^^^ source.example example.meta-scope.after-clear-scopes.example example.pushes-clear-scopes.example - string
#         ^ example.meta-scope.after-clear-scopes.example - example.pushes-clear-scopes.example - string
world foo bar
# ^^^^ source.example example.meta-scope.after-clear-scopes.example - example.pushes-clear-scopes - example.meta-scope.cleared-previous-meta-scope.example - string
#     ^^^^^^^ source.example example.meta-scope.cleared-previous-meta-scope.example - example.meta-scope.after-clear-scopes.example - example.pushes-clear-scopes - string
testing123'
# ^^ example.meta-scope.after-clear-scopes.example example.pops-clear-scopes.example - example.meta-scope.cleared-previous-meta-scope.example - string
#   ^^^^^^ string.quoted.single.example - example
#         ^ string.quoted.single.example punctuation.definition.string.end.example - example


```

In Sublime Text, the tests all pass. In syntect 1.2.0, there are some failures - it seems like nested "clear_scopes" are currently unsupported:

```
loading syntax definitions from testdata/Packages/cst/
[...]

PR #47, opened 2017-03-15
Fix some parsing bugs

keith-hall:
This PR fixes a few small bugs in parsing, like #45, and also adds more tests. These tests define their own syntax definitions, so that they won't break (or no longer effectively test what we want them to) when the Default Packages inevitably get updated.

### DIFF
--- a/src/parsing/parser.rs
+++ b/src/parsing/parser.rs
@@ -484,7 +484,25 @@ mod tests {
         // assert!(false);
     }
 
-    fn expect_scope_stack(line: &str, state: &mut ParseState, stack: &mut ScopeStack, expect: &[&str]) {
+    fn expect_scope_stacks(line: &str, expect: &[&str]) {
+        // check that each expected scope stack appears at least once while parsing the given test line
+        
+        //let syntax = SyntaxSet::load_syntax_file("testdata/parser_tests.sublime-syntax", true).unwrap();
+        use std::fs::File;
+        use std::io::Read;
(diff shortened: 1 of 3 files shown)


src/parsing/parser.rs
```diff
@@ -483,4 +483,256 @@ mod tests {
...
+        if let Some(missing) = expect.iter().filter(|e| !criteria_met.contains(&e)).next() {
+            panic!("expected scope stack '{}' missing", missing);
+        }
+    }
+
+    #[test]
+    fn can_parse_non_nested_clear_scopes() {
+        let syntax: SyntaxDefinition =
+            SyntaxDefinition::load_from_str(r#"
+                name: Simple Clear Scopes Test
```
trishume: Since this is almost exactly the same in each test, can you make this a static string instead?

```rust
static TEST_SYNTAX: &'static str = r#"
...
"#;
```

Or alternatively and perhaps even better, can you just commit it to a file in `testdata` and load from a file in the test?

To account for the small differences between the tests, just modify the syntax to look for slightly different things and change the input string that you parse between the tests. For example, instead of one clause matching `test` that changes, have a `#test1`, `#test10` and `#test_nested` match clause that contain the contents of the test from the three different syntaxes. Similar for `#infinite_loop` and `#infinite_seeming_loop` clauses. That way there doesn't need to be as much duplication and it is easier to find what changes between the different tests.

keith-hall: Done :)

trishume:
Looks awesome, thanks.

PR #52, opened 2017-03-23
iterate the stack in top-down (FILO) order to apply the prototypes

keith-hall:
This reduces the number of PHP failures from 104 to 40.
Basically, `with_prototype`s pushed onto the stack later take precedence over those pushed earlier.


trishume:
Awesome thanks.

Issue #51, opened 2017-03-18
Incorrect push/pop order when combining meta scopes, content scopes and pushing multiple things


trishume:
This is the bug that is causing at least some, probably most to all of the ASP tests to fail (cc @keith-hall).

The model of the parser outputting only the pushes and pops doesn't hold up to well in some cases. @keith-hall fixed the case of the weird behaviour of `set`, but I found another.

Consider the following section of the ASP syntax:
```yaml
  class_definitions:
    - match: '\b(?i:Class)\s'
      scope: storage.type.asp
      push: [inside_class, class_name]

  class_name:
    - meta_scope: meta.class.asp meta.class.identifier.asp
    # ...
    - match: '{{identifier}}'
      scope: entity.name.class.asp
      pop: true
    # ...

  inside_class:
    - meta_content_scope: meta.class.asp meta.class.body.asp
```

Sublime Text passes the following test:
```asa
' SYNTAX TEST "Packages/ASP/ASP.sublime-syntax"
 Class TestClass2 Public Sub TestSub () Response.Write("wow") End Sub End Class
'^^^^^ meta.class.asp meta.class.identifier.asp storage.type.asp
'      ^ meta.class.asp meta.class.body.asp meta.class.asp meta.class.identifier.asp
'                ^ meta.class.asp meta.class.body.asp
```

Notice how when the `meta_content_scope` of `inside_class` is introduced after `Class`, it is introduced under `meta.class.asp meta.class.identifier.asp`.

`syntect` does not do this. It pushes the `inside_class` `meta_content_scope` on top, but then after `TestClass2` when it pops the `class_name` context it pops 2 scopes, thinking it is popping the `meta_scope` of `class_name`, but actually the `inside_class` scopes are on top of the stack.

I see two ways of approaching this:
- Make pushes like `set` is in [Keith's set branch](https://github.com/forkeith/syntect/tree/set) where they pop the early meta scopes and then re-apply them in the right order.
- Make a breaking change to the API and scrap the idea of the parser outputting operations. Have stacks be linked lists instead of vectors, and have the parser output regions of text with a scope stack, where the scope stacks share tails. Unfortunately this would require a bunch of reference counting. It may be faster, it may be slower, I'm not sure. It would also be a bunch of work.

I'm inclined to take the first route. My worry is that if done simply it will lead to a bunch of redundant popping and pushing the same thing again in the case when only one item is pushed. It's probably a good idea to have a fast path for that case, although it would probably require a bunch more code.

[...]

PR #55, opened 2017-04-04
Fix bugs exposed by syntax tests

trishume:
Fixes some bugs found by running Sublime's syntax tests. Mostly fixing #51 with a total overhaul of the push/pop/set behaviour. This is a collaboration with @keith-hall that I did a couple weeks ago but haven't finished yet.

- [x] Don't match partial syntax def names @keith-hall
- [x] Fix behaviour with set @keith-hall 
- [x] Overhaul push/pop/set behaviour, fixing #51 @trishume
- [x] Iterate stack in correct order for prototype application @keith-hall 
- [x] Fix top level scopes, fixing #58 @trishume
- [x] Benchmark against master to check extent of performance regression (it didn't regress)
- [x] Add fast path for single-context push/pop/set if performance regressed (it didn't, so none added)


trishume:
I just fixed top level scopes using the reduction method I mentioned in #58.

Now the syntax tests for almost all languages pass perfectly. The following languages still fail:

- ASP: not sure why, think it is some weird thing with `with_prototype`. Speculation in #51 
- LaTeX: mostly works, but one pattern matches the empty string, should be fixed in Sublime Packages. Notes in #51 by @keith-hall 
- Markdown: one test fails on `<a name=\"demo\"></a>`. Not sure why. Might be that https://github.com/sublimehq/Packages/pull/877 doesn't actually work correctly. Not sure if @keith-hall ever ran the syntax tests on it.
- PHP: a number of failures, not sure why. Might be the same thing as ASP.

But things are now looking much much much better. Even the languages with failures produce highlighting that looks mostly correct.

trishume:
I did some benchmarking and didn't see any noticeable regression. There might be one of a percent or two but that doesn't really matter. It's easy to add a fast path later if it is needed.

I'm going to do some rebasing to get rid of redundant changing of the packs, then probably merge the package update and this.

trishume:
I moved my comment about the things that are still failing to #59. Please discuss further failures there.

Issue #46, opened 2017-03-13
The "name" key should be optional in "sublime-syntax" files


keith-hall:
In syntect, trying to parse a `sublime-syntax` file without a `name` top level key causes a panic:

> thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ParseSyntax(MissingMandatoryKey("name"))'

But in Sublime Text, this key is optional (indeed, the "new syntax" template in ST doesn't contain a placeholder for it), so currently syntect cannot parse some valid syntaxes. If it is not provided, ST derives the name of the syntax from the filename - i.e. "Hello World.sublime-syntax" would have a name of "Hello World".

trishume:
Ok interesting I didn't know that. 

Is there anything this is breaking right now? If not I'll probably work on the bugs in parsing found by the syntax tests first.

keith-hall:
No no, this is definitely not urgent :) it just prevents me from running syntax tests against some third party syntaxes, nothing major :)

FichteFoll:
Just wanted to note that I was trying to run CI tests using syntect for my PackageDev third-party package and am bothered by this issue. You can read more about this here: https://github.com/SublimeText/PackageDev/issues/98

trishume:
I just pushed `v1.3.0` to crates.io with the fix for this. Release notes are on the Github releases page.

FichteFoll:
Thanks. You possibly want to keep this issue open because it doesn't do The Right Thing atm, but I don't know if there is anything that depends on a syntax's name to be honest.

PR #61, opened 2017-04-05
if syntax lookup by name key fails, lookup by the base filename

keith-hall:
This emulates ST's behavior, as per https://github.com/trishume/syntect/issues/46:

> in Sublime Text, this key is optional. If it is not provided, ST derives the name of the syntax from the filename - i.e. "Hello World.sublime-syntax" would have a name of "Hello World".

### DIFF
--- a/src/parsing/syntax_set.rs
+++ b/src/parsing/syntax_set.rs
@@ -136,7 +136,13 @@ impl SyntaxSet {
     }
 
     pub fn find_syntax_by_name<'a>(&'a self, name: &str) -> Option<&'a SyntaxDefinition> {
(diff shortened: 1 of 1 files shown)


trishume:
Hmmmm, I'll have to think about this. This is definitely the simplest way of solving the main problem, but the `name` field is public and it doesn't set that correctly, which might be important for use cases like the Syntax selector popup in Sublime's bottom right corner.

I may still merge this, but I think I'd prefer to find a way of passing a `fallbackName: Option<String>` derived from the file path into the YAML loader.

keith-hall:
Good point, that sure sounds like a better solution :)

Issue #65, opened 2017-04-30
Drop rustc-serialize dependency


dtolnay:
It has been deprecated: [announcement](https://users.rust-lang.org/t/deprecation-of-rustc-serialize/10485?u=dtolnay).

Is there any functionality you would require from Serde or another library before this would be possible?

trishume:
Yah, I should probably do this at some point.

robinst:
It looks like all the dependencies now also have versions that depend on serde 1.0: [plist](https://crates.io/crates/plist) and [bincode](https://crates.io/crates/bincode)

I can look at this if you want me to. I'm not sure what you want to do about this though: https://github.com/trishume/syntect/blob/c9b611f122a29238db66bcd11162558ad9cdb4b7/src/dumps.rs#L7-L9

The types from rustc-serialize are part of the API currently.

trishume:
@robinst that might be a good idea. As far as I interpret my contract with Google, if I work on syntect any time in the next 3 months, it all belongs to Google since unlike my other projects, it is clearly in Google's business since it is used in a Google project. I'm not sure how that interacts with it being MIT licensed, but I'd rather play it safe.

 As for API/binary stability, meh. My policy is this: semver stability for the easy/html/core APIs and "meh, whatever" for everything else. I don't think anyone is using these in a way that will break except perhaps @Byron, and now that ping has given him warning. Although perhaps it would be a good idea to serialize a version number into the files before the rest of the contents so that at least the code could error on incompatibly files (and allow them to be regenerated perhaps) instead of just panic-ing or doing the wrong thing when de-serializing.

Byron:
Thanks for the friendly warning :)! When I set things up for tokio-cassandara, I made it so that regenerating our syntaxes and themes would be easy. Whatever you come up with, I am sure I will be fine.

robinst:
> if I work on syntect any time in the next 3 months, it all belongs to Google [...]

I would probably ask for clarification on that, normally companies are ok with it for open source projects and if you ask for permission first.

Thanks for that @Byron, that made it easier. I have made the changes now and will submit a PR soon.

PR #66, opened 2017-05-08
Switch from rustc-serialize to serde (#65)

robinst:
CC @Byron


Cargo.toml
```diff
@@ -19,11 +19,13 @@ walkdir = "^1.0"
...
 lazy_static = "^0.2"
 bitflags = "^0.8"
-plist = "0.1.1"
-rustc-serialize = "^0.3"
-bincode = "0.6"
+plist = "0.2"
+bincode = "0.8"
 flate2 = "^0.2"
 fnv = "^1.0"
+serde = { version = "1.0", features = ["rc"] }
```
robinst: The `rc` feature is necessary because `MatchPattern` has a `ContextPtr` which is `Rc<RefCell<Context>>`. The `Serialize` and `Deserialize` impls for `Rc` and `RefCell` are behind a feature: https://github.com/serde-rs/serde/blob/master/serde/Cargo.toml#L67-L70

src/highlighting/settings.rs
```diff
@@ -32,5 +33,22 @@ impl From<PlistError> for SettingsError {
 }
 
 pub fn read_plist<R: Read + Seek>(reader: R) -> Result<Settings, SettingsError> {
-    Ok(try!(Plist::read(reader)).into_rustc_serialize_json())
+    let plist = Plist::read(reader)?;
+    Ok(to_json(plist))
+}
+
+fn to_json(plist: Plist) -> Settings {
```
robinst: Not quite sure this is the right way to go. This is the reason for the `serde_json` dependency. We could add our own enum with the cases we're interested in or use `Plist::*` directly instead of using `serde_json`. But I'm not sure what the intention of the original code is, to let users load settings from JSON instead?

src/parsing/syntax_definition.rs
```diff
@@ -243,40 +246,6 @@ impl MatchPattern {
     }
 }
 
-/// Only valid to use this on a syntax which hasn't been linked up to other syntaxes yet
-impl Encodable for MatchPattern {
```
robinst: These are not necessary because with serde we can just do `#[serde(skip_serializing, skip_deserializing)]` :).

src/highlighting/settings.rs
trishume: I think the original code just liked the Json data model better, and it was easy. I took the theme stuff from an older project.

This is fine, certainly text editors may want to load themes from JSON instead of plists. I'm fine with this as is.

PR #69, opened 2017-05-22
[WIP] Split syntect into two crates

cmyr:
Okay, here's a first stab at this little refactor. All mostly clean, and the tests are passing. Two little comment-worthy changes:

- `LoadError` no longer has a member for `SyntaxParseError`, and lives in the new `syntect-highlight` crate. `SyntaxParseError` gets an `IoError` member. The error types of a few return values are adjusted to accommodate the change.

- **bigger**: `ThemeSet::load_defaults()` is replaced by a free `load_default_themeset()` function. An alternative would've been to add a Trait. (This is all getting around the 'no impl for type in external crate' rule). I don't like this, and doubt you will? I'm open to whatever you prefer.

This isn't totally cleaned up, because I wanted to wait and see if I was going to be reverting changes before removing unused imports / comments etc.

If these seems sane to you I'll do some cleanup and update docs.


examples/gendata.rs
```diff
@@ -3,8 +3,9 @@
 //! Although it is a valid example for serializing syntaxes, you probably won't need
 //! to do this yourself unless you want to cache your own compiled grammars.
 extern crate syntect;
+extern crate syntect_highlighting as highlighting;
```
trishume: I'd rather not make every other user of the crate's job harder just because Xi wants to use some functionality without depending on `onig`.

You can fix this by adding `pub extern crate syntect_highlighting as highlighting` to `src/lib.rs`.

trishume:
After reviewing, I realized something that I should have though of before. Instead of splitting `syntect` into separate crates, I think this could be done with a Cargo feature like https://github.com/trishume/syntect/pull/42.

The real underlying structure of `syntect` is made up of a whole bunch of things that could theoretically be different crates: scopes & selectors, highlighting, parsing, html, dumps, and an easy end-to-end wrapper. The only reason for the particular crate split in this PR is it's what Xi would use, but it's actually a weird split, using just the parsing but not the highlighting would be a much more common use case.

The only real reason for splitting things up is avoiding certain dependencies and binary size. Xi wants to avoid `onig`. #42 wanted to avoid including the theme/syntax dumps. Some people use just the parsing but don't care about minor unnecessary binary size so they haven't yet agitated for separating that.

Instead of splitting `syntect` up into N different crates in a massive API-breaking conflagration that would be a pain to update and depending on the aggregation, a pain for users, I propose just making liberal use of Cargo features.

Using a Cargo feature to do this would avoid the public API breakage of moving the `Scope` type and changing `ThemeSet::load_defaults()` while still allowing Xi to avoid depending on `onig`.

Sorry for thinking of this only after you did this work and I saw the problems this PR exposed. I'm open to being convinced otherwise but I don't think I'll accept this approach but I would accept a PR that added a `parsing` feature that put `#[cfg(feature = "parsing")]` on a bunch of the lines in `src/parsing/mod.rs` and some other places, like the enum value in the error type. Then modified `Cargo.toml` to add that feature to the default set and tie the inclusion of `onig` to it. See [here](https://doc.rust-lang.org/book/conditional-compilation.html) for how to interact with other features.

Yes it would be a bit weird in Xi's case to have a `parsing` module that only contains scopes, but it's not that weird and I don't mind too much since I expect Xi to be the only thing that uses the feature.

cc @raphlinus who originally proposed separating crates

cmyr:
superseded by #70

PR #70, opened 2017-05-24
Feature flag for parsing

cmyr:
This is a dirty first-pass at putting more of syntect behind feature flags. Specifically, this adds the `parsing` feature, which wraps most of the `parsing` module. 

@trishume I wanted to check in and see what you thought. With `--release --no-features` this compiles to about 1.5M on macOS.  That seems manageable. Is there any low hanging fruit you can think of? I'm looking at `dumps`, but that's probably convenient for us to hold on to for now, and doesn't really produce much savings.

If this is okay with you I'll clean it up a bit and revise docs as needed.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,15 +14,15 @@ exclude = [
 
 [dependencies]
 yaml-rust = { version = "0.3.5", optional = true }
-onig = "^1.2"
+onig = { version = "^1.2", optional = true }
 walkdir = "^1.0"
-regex-syntax = "^0.4"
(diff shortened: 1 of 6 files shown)


cmyr:
@trishume this is ready to go

trishume:
@Byron as I think the only other current user of feature flags, this change in the next release may break your build. All you have to do is add the `parsing` feature flag to your `Cargo.toml`.

Byron:
@trishume Thanks for the heads-up! The upgrade was performed without any issues :)!

PR #72, opened 2017-05-30
Added Highlighter::style_for_stack

cmyr:
I... think this is right? 🤷‍♂️

### DIFF
--- a/src/highlighting/highlighter.rs
+++ b/src/highlighting/highlighter.rs
@@ -244,10 +244,10 @@ impl<'a> Highlighter<'a> {
     ///
     /// This operation is convenient but expensive. For reasonable performance,
     /// the caller should be caching results.
-    pub fn style_for_stack(&self, stack: &ScopeStack) -> Style {
+    pub fn style_for_stack(&self, stack: &[Scope]) -> Style {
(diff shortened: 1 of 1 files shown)


src/highlighting/highlighter.rs
```diff
@@ -239,6 +239,19 @@ impl<'a> Highlighter<'a> {
             font_style: None,
         }
     }
+
+    /// Returns the fully resolved style for the given stack.
+    ///
+    /// This operation is convenient but expensive. For reasonable performance,
+    /// the caller should be caching results.
+    pub fn style_for_stack(&self, stack: &ScopeStack) -> Style {
```
trishume: This should really take a `&[Scope]` like the other `Highlighter` methods, there's no reason for it to take a `ScopeStack`.

cmyr: I made it take a `ScopeStack` so that I could use the `bottom_n` method that is so helpfully defined there.

trishume: With a slice you can use the `[0..n]` syntax which is shorter, and is actually all the `bottom_n` method does.

cmyr:
okay, made that change. 😎

PR #75, opened 2017-06-05
Improved syncat

rdnetto:
This PR adds some niceties to make syncat slightly more useful as a tool in its own right:
* ability to specify the theme at runtime (and cache its parsed representation)
* ability to list supported file types

I apologise if the code isn't idiomatic, as I'm not very familiar with Rust.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,10 +26,10 @@ fnv = { version = "^1.0", optional = true }
 serde = { version = "1.0", features = ["rc"] }
 serde_derive = "1.0"
 serde_json = "1.0"
-getopts = "0.2"
 
 [dev-dependencies]
 regex = "0.2.1"
+getopts = "0.2"
 
 [features]
 static-onig = ["onig/static-libonig"]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,6 +26,7 @@ fnv = { version = "^1.0", optional = true }
 serde = { version = "1.0", features = ["rc"] }
 serde_derive = "1.0"
 serde_json = "1.0"
+getopts = "0.2"
 
 [dev-dependencies]
 regex = "0.2.1"
--- a/examples/syncat.rs
+++ b/examples/syncat.rs
@@ -1,14 +1,25 @@
 extern crate syntect;
+use std::path::Path;
 use syntect::parsing::SyntaxSet;
 use syntect::highlighting::{Theme, ThemeSet, Style};
 use syntect::util::as_24_bit_terminal_escaped;
 use syntect::easy::HighlightFile;
+use syntect::dumps::{from_dump_file, dump_to_file};
 
 use std::io::BufRead;
 
-fn load_theme(path: &String) -> Theme {
-    // TODO: cache this
-    ThemeSet::get_theme(path).unwrap()
+fn load_theme(tm_file: &String) -> Theme {
+    let tm_path = Path::new(tm_file);
+    let tm_cache = tm_path.with_extension("themedump");
+
+    if tm_cache.exists() {
+        from_dump_file(tm_cache).unwrap()
+    } else {
+        let theme = ThemeSet::get_theme(tm_path).unwrap();
+        dump_to_file(&theme, tm_cache).unwrap();
+        theme
+
+    }
 }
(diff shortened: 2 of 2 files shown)


examples/syncat.rs
```diff
@@ -1,22 +1,61 @@
...
+use syntect::highlighting::{Theme, ThemeSet, Style};
 use syntect::util::as_24_bit_terminal_escaped;
 use syntect::easy::HighlightFile;
+use syntect::dumps::{from_dump_file, dump_to_file};
 
 use std::io::BufRead;
 
+fn load_theme(tm_file: &String) -> Theme {
+    let tm_path = Path::new(tm_file);
+    let tm_cache = tm_path.with_extension("themedump");
```
trishume: I'm reluctant to use the same extension for dumps of `ThemeSet` and `Theme`, which are incompatible. Arguably this is my fault since I should have named the dump files for `ThemeSet` `themesetdump` instead of `themedump`. But, that's hard to change now.

How about `cachedtheme` instead? Or something else you think of.

examples/syncat.rs
```diff
@@ -1,22 +1,61 @@
...
+
+    if tm_cache.exists() {
+        from_dump_file(tm_cache).unwrap()
+    } else {
+        let theme = ThemeSet::get_theme(tm_path).unwrap();
+        dump_to_file(&theme, tm_cache).unwrap();
```
trishume: I'm reluctant to have this be the default functionality since one might want to use this (handy!) feature for debugging/writing a theme, and run syncat multiple times as you change the `tmTheme` file and see how the output changes. Automatically creating cache files would stop that from working.

I'd appreciate if you could add an argument like `--cache-theme` to enable create the cache file.

examples/syncat.rs
```diff
@@ -1,22 +1,61 @@
...
+}
+
 fn main() {
     let ss = SyntaxSet::load_defaults_newlines(); // note we load the version with newlines
     let ts = ThemeSet::load_defaults();
 
     let args: Vec<String> = std::env::args().collect();
+    let src;
+    let theme;
+    let theme_ref;
```
trishume: This works, but the idiomatic way to do this in Rust would be to use a `Cow`. You would just have a `theme` variable and in the case of loading it from the `ThemeSet` you would use `Cow::Borrowed(&ts.themes[...])` and in the loading case you would use `Cow::Owned(load_theme(...))` and then you would pass it to `HighlightFile` as `&theme`.

examples/syncat.rs
```diff
@@ -1,22 +1,61 @@
...
+
+    }
+}
+
 fn main() {
     let ss = SyntaxSet::load_defaults_newlines(); // note we load the version with newlines
     let ts = ThemeSet::load_defaults();
 
     let args: Vec<String> = std::env::args().collect();
+    let src;
```
trishume: You could avoid this by putting `let src = args.last()` after the if statements.

examples/syncat.rs
rdnetto: Sure, maybe `.tmdump`? (To be consistent with the dump suffix)

examples/syncat.rs
rdnetto: :+1:

examples/syncat.rs
rdnetto: No problem, thanks for the detailed explanation. :)

examples/syncat.rs
rdnetto: Makes sense. The complexity of the argument parsing is starting to get a bit much though, so I'll rewrite it to use [getopts](https://doc.rust-lang.org/getopts/getopts/index.html).

rdnetto:
Sorry about my delay in replying - I must have missed the email notification somehow.

Cargo.toml
```diff
@@ -26,6 +26,7 @@ fnv = { version = "^1.0", optional = true }
 serde = { version = "1.0", features = ["rc"] }
 serde_derive = "1.0"
 serde_json = "1.0"
+getopts = "0.2"
```
trishume: I think this can be moved to the `[dev-dependencies]` section.

Issue #20, opened 2016-10-07
Enable multi-threaded use


robinst:
Background: I'm trying to write a service that receives source code via HTTP, parses it using syntect and then returns some information about it. I'm using Iron for it, but this issue would also occur with other uses. In Iron, the [`Handler`](http://ironframework.io/doc/iron/middleware/trait.Handler.html) trait which handles requests is required to be `Send` and `Sync`, meaning it can be reused between threads.

So, the simplest implementation is to just call `SyntaxSet::load_defaults_nonewlines()` on every request, then figure out which `SyntaxDefinition` to use, and then parse the text. Unfortunately, this means that the syntax definitions have to be loaded for each request, which is quite a bit of work that could be avoided.

You can see where this is going. I'd like to have a way to only load `SyntaxSet` once at startup and then share it between all the request handlers. Or if that's not possible, at least share the `SyntaxDefinition` between threads (and have a mutex protecting `SyntaxSet`).

If we try to write that with the current syntect, it would look something like this:

``` rust
#[test]
fn threading_with_shared_syntax_set() {
    let syntax_set = SyntaxSet::load_defaults_nonewlines();
    let syntax = syntax_set.find_syntax_by_name("rust").unwrap();

    thread::spawn(|| {
        let parse_state = ParseState::new(&syntax);
        let mut f = File::open("testdata/parser.rs").unwrap();
        let mut reader = BufReader::new(f);

        for line in reader.lines() {
            let line = line.unwrap();
            parse_state.parse_line(&line);
        }
    });
}
```

And it currently fails to compile like this:

```
[101] % cargo test threading
   Compiling syntect v1.0.0 (file:///Users/rstocker/Projects/rust/syntect)
error[E0277]: the trait bound `std::rc::Rc<std::cell::RefCell<syntect::parsing::syntax_definition::Context>>: std::marker::Sync` is not satisfied
  --> tests/threading.rs:14:5
   |
14 |     thread::spawn(|| {
   |     ^^^^^^^^^^^^^
   |
   = note: `std::rc::Rc<std::cell::RefCell<syntect::parsing::syntax_definition::Context>>` cannot be shared between threads safely
   = note: required because it appears within the type `std::option::Option<std::rc::Rc<std::cell::RefCell<syntect::parsing::syntax_definition::Context>>>`
   = note: required because it appears within the type `syntect::parsing::SyntaxDefinition`
   = note: required because it appears within the type `&syntect::parsing::SyntaxDefinition`
[...]

PR #78, opened 2017-06-17
Add pooling for parallel usage of SyntaxSet

williamyaoh:
See #20.

This PR adds a struct `SyntaxSetPool` which pools `SyntaxSet`s for parallel usage with the minimum amount of initialization of said `SyntaxSet`s. By default, it will only initialize as many `SyntaxSet`s as the running machine has CPUs, though this can be specified as well.

Examples of the usage of `SyntaxSetPool` can be found in `examples/parsyncat.rs`, which highlights multiple source files in parallel and prints the results to `stdout`, and `benches/parallel_parsing.rs`.

## Issues

`SyntaxDefinition`'s preexisting `Clone` implementation, which was merely incorrect before (due to `Weak` pointers in the resulting clone), can now cause memory unsafety due to their containing `SyntaxSet` being ferried across thread boundaries. Since this implementation was already incorrect, ideally this `Clone` implementation would be fixed or removed before merging this PR. Alternatively, there might be a way to hide it from library users.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,9 +26,11 @@ fnv = { version = "^1.0", optional = true }
 serde = { version = "1.0", features = ["rc"] }
 serde_derive = "1.0"
 serde_json = "1.0"
+num_cpus = "^1.5"
 
 [dev-dependencies]
 regex = "0.2.1"
+crossbeam = "0.2.0"
 
 [features]
 static-onig = ["onig/static-libonig"]
--- a/src/parsing/syntax_set_pool.rs
+++ b/src/parsing/syntax_set_pool.rs
@@ -1,0 +1,122 @@
+use std::cell::UnsafeCell;
+use std::sync::{Mutex, Condvar};
+
+use num_cpus;
+
+use super::SyntaxSet;
+
+/// This is intentionally not public. The `Send` implementation makes
+/// this unsafe to use outside this module.
+struct LazyInit<T> {
+    inner: UnsafeCell<Option<T>>
+}
+
+/// A pool for parallelizing syntax parsing/highlighting. Constructs
+/// a lazily-initialized pool of `SyntaxSet`s, each initialized the
+/// same way. The first time each `SyntaxSet` gets used, the thread
+/// using it runs its setup code.
+///
+/// Use syntax sets by passing a closure to `with_syntax_set()`. A
+/// `SyntaxSet` will be removed from the pool, and a reference to it
+/// passed to the given closure. Afterwards, the set will automatically
+/// be returned to the pool.
+pub struct SyntaxSetPool<F: Fn() -> SyntaxSet> {
+    /// We intentionally use a *stack* of `SyntaxSet`s so that
+    /// already-initialized syntaxes get reused as much as possible.
+    syntaxes: Mutex<Vec<LazyInit<SyntaxSet>>>,
+    has_syntax: Condvar,
+    init_fn: F
+}
+
+// The unsafety boundary for this `unsafe` is not *this* module, but the
+// *parent* module. With the way we're using this, `SyntaxSet` must not
+// implement either `Clone` or `Copy`.
+unsafe impl<T> Send for LazyInit<T> {}
+
+impl<T> LazyInit<T> {
+    /// Create a `LazyInit` with contents uninitialized.
+    fn new() -> Self {
+        LazyInit { inner: UnsafeCell::new(None) }
+    }
+
+    fn maybe_init<F>(&self, init_fn: F) -> *mut T where F: Fn() -> T {
+        let inner_ptr = self.inner.get();
+        unsafe {
+            if let None = *inner_ptr {
+                *inner_ptr = Some(init_fn());
+            }
+
+            match *inner_ptr {
+                Some(ref mut inner) => inner as *mut T,
+                None => unreachable!()
+            }
+        }
+    }
+
+    #[allow(dead_code)]
+    fn get_or<F>(&self, init_fn: F) -> &T where F: Fn() -> T {
+        unsafe { &*self.maybe_init(init_fn) }
+    }
+
+    fn get_mut_or<F>(&self, init_fn: F) -> &mut T where F: Fn() -> T {
+        unsafe { &mut *self.maybe_init(init_fn) }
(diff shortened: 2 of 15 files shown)


williamyaoh:
I'd like some eyes on the `unsafe` code in `parsing/syntax_set_pool.rs`; from the benchmarks/tests, it seems to be safe the way I'm using it (barring the aforementioned `Clone` problem on `SyntaxDefinition`), but it doesn't hurt to check.

src/parsing/syntax_set_pool.rs
```diff
@@ -0,0 +1,135 @@
...
+/// using it runs its setup code.
+///
+/// Use syntax sets by passing a closure to `with_syntax_set()`. A
+/// `SyntaxSet` will be removed from the pool, and a reference to it
+/// passed to the given closure. Afterwards, the set will automatically
+/// be returned to the pool.
+pub struct SyntaxSetPool<F: Fn() -> SyntaxSet> {
+    /// We intentionally use a *stack* of `SyntaxSet`s so that
+    /// already-initialized syntaxes get reused as much as possible.
+    syntaxes: Mutex<Vec<LazyInit<SyntaxSet>>>,
```
trishume: Instead of using this whole unsafe `LazyInit` business (which I'm very uncomfortable with, especially given that without it `syntect` itself contains no unsafe code), you couldn't you do something like this:

1. Define `type PooledSet = Arc<Mutex<SyntaxSet>>;
1. In SyntaxSetPool use `syntaxes: Mutex<Vec<PooledSet>>`.
1. Add a field `max_pool_size` that you set on initialization to the core count.
1. Add a field `num_in_pool` that starts at `0`.
1. At initialization the `syntaxes` vector is empty.
1. When getting a syntax, if there are none left in `syntaxes`:
    - If `num_in_pool < max_pool_size`, create a new one and bump `num_in_pool`.
    - Else block on `has_syntax`
1. If there is a syntax, pop it, lock it, call `go`, then unlock it, lock the pool and return it.

So at the cost of one extra never-contested lock (which AFAIK should be very cheap) per syntax get, the amount of code required decreases substantially and it no longer requires `unsafe`.

Cargo.toml
```diff
@@ -26,9 +26,11 @@ fnv = { version = "^1.0", optional = true }
 serde = { version = "1.0", features = ["rc"] }
 serde_derive = "1.0"
 serde_json = "1.0"
+num_cpus = "^1.5"
```
trishume: I think this should be behind the `parsing` feature just like a bunch of the other crates. Same for the corresponding `extern crate` line.

src/parsing/syntax_set_pool.rs
```diff
@@ -0,0 +1,135 @@
...
+
+/// A pool for parallelizing syntax parsing/highlighting. Constructs
+/// a lazily-initialized pool of `SyntaxSet`s, each initialized the
+/// same way. The first time each `SyntaxSet` gets used, the thread
+/// using it runs its setup code.
+///
+/// Use syntax sets by passing a closure to `with_syntax_set()`. A
+/// `SyntaxSet` will be removed from the pool, and a reference to it
+/// passed to the given closure. Afterwards, the set will automatically
+/// be returned to the pool.
```
trishume: I would add a note here to avoid using this when you're already using a crate like `rayon` that uses a small thread pool. In that case it is simpler to just use:

```rust
thread_local!(static SYNTAX_SET: SyntaxSet = SyntaxSet::load_defaults_newlines());

// and later:
SYNTAX_SET.with(|syntax_set| ... );
```

src/parsing/syntax_set_pool.rs
williamyaoh: Unfortunately, I don't think that solution would work, since that wouldn't make `SyntaxSetPool` be `Sync`, since `PooledSet` wouldn't be `Sync`. This is the big sticking point for this pull request too, but I don't see a way to transport `SyntaxSet`s across thread boundaries the way it's needed without using `unsafe`.

I agree with you that adding `unsafe` code is unsavoury, so maybe in the end the best way to resolve #20 would be to recommend people use `thread_local!` + `rayon`.

trishume: @williamyaoh you're right, my mistake. Looking into the docs for `Send` and `Sync` to try and understand why the `Mutex` doesn't end up `Sync`, I think the same issue might apply to your unsafe code.

I haven't thought about it very long yet but I think you might be able to pass an initializer function that clones one of the `Rc`s inside `SyntaxSet`, stashes it outside the closure inside a `RefCell` or something and use that to create a data race on the refcount.

Unless one of us thinks of a way to do this without `unsafe`, or perhaps more possibly, using a well-tested established crate (which might use unsafe) to do it, I'm leaning towards just recommending `thread_local!` and `rayon` in the docs like you say.

williamyaoh: I don't believe, with `SyntaxSet` being written the way it is, that there's a way to implement pooling in `syntect` itself using only safe code.

For the purposes of pooling, the only marker we care about is making `SyntaxSet` be `Send` (If it were `Sync`, we could just share it between threads directly, with no need for pooling). My understanding of `Send` is that it denotes a type that's safe to move ownership of across thread boundaries, so making `SyntaxSet` be `Send` would require everything it owns to *also* be `Send`, which clearly can't happen because of the `Rc`s. For the same reason, no external crate could provide a safe interface to transport non-`Send` types across thread boundaries in general.

I'd just go with recommending `rayon`. On reflection, the amount of code within the unsafety boundary caused by unsafely `impl`ementing `Send` on `SyntaxSet` is a bit too large to stomach.

trishume: @williamyaoh that's a good point, you're right. There's basically no way to prove to the compiler that you haven't cloned one of the `Rc`s and are going to access it from another thread. And I'm not even sure there is a way to stop you from doing that, even without proving it to the compiler.

I can update the docs and maybe add an example using Rayon and thread_local!, but it's inconvenient to contribute to syntect right now because I'm working at Google and it's used in a Google project, so I can't claim that it's unrelated and I'd have to go through a process.

So either I can do that in a couple months after my internship or I can accept a PR if you do it.

Issue #80, opened 2017-06-20
Naming confusion in ThemeSettings


cmyr:
Comparing the documentation of `ThemeSettings` in `theme.rs`, and comparing that with the sublimetext documentation [here](https://www.sublimetext.com/docs/3/color_schemes.html#global_settings), I'm confused in a few places.

1. the ST docs describe `selection` and `selectionForeground` properties. The syntect docs describe `selection` and `selectionBackground`. In the actual `InspiredGitHub` theme, as loaded by syntect, `selection` has a value but `selectionBackground` is missing. This is surprising to me: I would expect that the background is more likely to be present than the foreground?

2. Less significantly there are a few fields missing. None of these are critical (to me), but among the missing are `inactiveSelectionForeground`.

3. There are some items for which the syntect docs are either unclear (`foreground`, `selection`) or have errors (`find_highlight_foreground`).

@trishume if you have anything to add here let me know, especially around item 1; then I'll PR as appropriate.

trishume:
I know little to nothing about this because the theme loading is the one part of syntect that I straight-up copy-pasted from https://github.com/defuz/sublimate

It's possible its wrong.

cmyr:
Okay, will investigate.

PR #81, opened 2017-06-20
Fix issues in theme.rs and update themedump

cmyr:
Here's the followup for #80. This fixes the one actual bug (confusion around `selection`, missing `selectionForeground`) improves docs, and adds some missing options. It also marks as deprecated (docs only) two fields: `selection_background`, which was a mistake, and `highlight_foreground` which I couldn't find docs for.


src/highlighting/theme.rs
```diff
@@ -25,16 +25,28 @@ pub struct Theme {
...
 
+    /// The color to use for the squiggly underline drawn under misspelled words.
+    pub misspelling: Option<Color>,
+    /// The color of the border drawn around the viewport area of the minimap.
+    /// Only used when the `draw_minimap_border` setting is enabled.
+    pub minimap_border: Option<Color>,
+    /// A color made available for use by the theme.
+    pub accent: Option<Color>,
+    /// CSS passed to popups.
+    pub popup_css: Option<Color>,
```
keith-hall: The CSS is a string, not a color :)

cmyr: @keith-hall huh yea that makes sense. Are you running this code anywhere? My work in here is mostly thinking about my use case ([xi-editor](https://github.com/google/xi-editor)), where I'm not using many of these properties, hence my being pretty cavalier about testing etc.

keith-hall: I'm not using the color schemes much, atm I mainly just use them when I want to get some code output as a syntax highlighted HTML file, so the CSS isn't too important to me at the moment. But, that said: if we add things, we might as well get them correct, to save potentially breaking the API when it gets used/fixed later ;)

Issue #29, opened 2016-12-31
Use in threaded context seems to require unsafe wrapper struct


ojensen5115:
Documentation states that the SyntaxSet and ThemeSet should be loaded once at program startup (and indeed reloading them whenever highlighting is needed hurts performance). However, SyntaxSet does not appear to be threadsafe, possibly due to the internal use of RefCell in SyntaxDefinition.

As a result, the simplest way of getting it to work in Iron that I have found requires the following (unsafe) dance:

```
struct HighlighterSyntaxSet {
    ss: SyntaxSet,
}
unsafe impl Sync for HighlighterSyntaxSet {}

lazy_static! {
    static ref HL_SYNTAX_SET: HighlighterSyntaxSet = {
        let ss = SyntaxSet::load_defaults_newlines();
        HighlighterSyntaxSet {ss: ss}
    };
}
```

Using it directly without the unsafe wrapper results in the following error:

```
the trait `std::marker::Sync` is not implemented for `std::rc::Rc<std::cell::RefCell<syntect::parsing::syntax_definition::Context>>`
    = note: `std::rc::Rc<std::cell::RefCell<syntect::parsing::syntax_definition::Context>>` cannot be shared between threads safely
    = note: required because it appears within the type `std::option::Option<std::rc::Rc<std::cell::RefCell<syntect::parsing::syntax_definition::Context>>>`
    = note: required because it appears within the type `syntect::parsing::SyntaxDefinition`
    = note: required because of the requirements on the impl of `std::marker::Sync` for `std::ptr::Unique<syntect::parsing::SyntaxDefinition>`
    = note: required because it appears within the type `alloc::raw_vec::RawVec<syntect::parsing::SyntaxDefinition>`
    = note: required because it appears within the type `std::vec::Vec<syntect::parsing::SyntaxDefinition>`
    = note: required because it appears within the type `syntect::parsing::SyntaxSet`

```

The use of lazy_static here is simply to avoid the more complex code necessary for using the `persistent` crate, or `Arc`s directly. I don't believe it affects the issue at hand.

All that said, it is of course entirely possible I'm using this incorrectly -- I'm somewhat new to Rust.

trishume:
So that unsafe wrapper can actually do bad things because the lazy loading of regexes modifies the structure, and you don't want that to happen in two threads at the same time.

See https://github.com/trishume/syntect/issues/20 for a long overview of the issues and some ways around the issue.

ojensen5115:
[...]

PR #82, opened 2017-06-25
Recommend rayon, add examples/documentation of parallel usage

williamyaoh:
See #20.

@trishume It might be good to add the recommendation to #20 and #29, before closing them. I imagine it would carry more weight coming from the project owner, too :)

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -118,7 +118,9 @@ Any time the file is changed the latest cached state is found, the cache is clea
 
 ### Parallelizing
 
-`syntect` doesn't provide any built-in facilities to enable highlighting in parallel. Some of the important data structures are not thread-safe, either, most notably `SyntaxSet`. However, if you find yourself in need of highlighting lots of files in parallel, the recommendation is to use the `thread_local!` macro in `libstd` along with `rayon`. See `examples/parsyncat.rs` for an example of how to do this.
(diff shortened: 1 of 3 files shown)


Readme.md
```diff
@@ -17,7 +17,7 @@ I've spent months working on, tweaking, optimizing, documenting and testing this
 `syntect` is [available on crates.io](https://crates.io/crates/syntect). You can install it by adding this line to your `Cargo.toml`:
 
 ```toml
-syntect = "1.4"
+syntect = "1.6"
```
trishume: Thanks, I always forget to update this.

Readme.md
```diff
@@ -116,6 +116,10 @@ This way from the time the edit happens to the time the new colouring gets rende
 
 Any time the file is changed the latest cached state is found, the cache is cleared after that point, and a background job is started. Any already running jobs are stopped because they would be working on old state. This way you can just have one thread dedicated to highlighting that is always doing the most up-to-date work, or sleeping.
 
+### Parallelizing
+
+`syntect` doesn't provide any built-in facilities to enable highlighting in parallel. Some of the important data structures are not thread-safe, either, most notably `SyntaxSet`. However, if you find yourself in need of highlighting lots of files in parallel, the recommendation is to use the `thread_local!` macro in `libstd` along with `rayon`. See `examples/parsyncat.rs` for an example of how to do this.
+
```
trishume: For bonus points, add another paragraph mentioning that if you'd like to use syntect with a library that uses threads, but too many, like Iron, then you can use a thread pool crate (like https://github.com/reem/rust-scoped-pool) to create a small pool, and possibly also `mpsc` channels to transfer the work/result.

And maybe also link to #20 and #78 for people interested in why that is recommended.

robinst:
Looks good to me too, using `thread_local!` and rayon is exactly what we do.

Readme.md
williamyaoh: The reasoning behind that is because `rayon` *might*, but not necessarily *will* run a parallelizable computation using a separate thread, right? Otherwise, it seems like just configuring `rayon`'s own thread pool size would be enough.

It'd also be good to mention that the reason behind all of that advice is to avoid initializing more `SyntaxSet`s than necessary.

PR #85, opened 2017-07-03
Make syntax set dumps deterministic

robinst:
Because the HashMaps used in SyntaxDefinition have non-deterministic
ordering, loading and dumping a syntax set resulted in different binary
data each time.

Having a deterministic result is nice for some things such as making a
change to a source .sublime-syntax file and seeing if it changes the
resulting dump or not. Or checking that regenerating the dump with some
modifications in the code does not alter the output.


robinst:
No worries :). The migration to serde made it a piece of cake to implement.

PR #92, opened 2017-07-25
Split up can_parse test

robinst:
Also parse the Ruby heredoc as individual lines.

### DIFF
--- a/src/parsing/parser.rs
+++ b/src/parsing/parser.rs
@@ -434,32 +434,21 @@ impl ParseState {
 mod tests {
     use super::*;
     use parsing::{SyntaxSet, Scope, ScopeStack};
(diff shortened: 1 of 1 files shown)


src/parsing/parser.rs
```diff
@@ -486,28 +473,46 @@ mod tests {
...
+        let mut stack = ScopeStack::new();
+        for &(_, ref op) in ops.iter() {
+            stack.apply(op);
         }
-        assert_eq!(test_stack2, test_stack);
+        assert_eq!(stack, test_stack);
+    }
 
-        // for testing backrefs
-        let line4 = "lol = <<-SQL\nwow\nSQL";
```
robinst: Not sure how this worked before, given that it uses a "nonewlines" syntax.

Issue #59, opened 2017-04-04
Fix the last few syntax tests


trishume:
After #55 fixed most syntax test exposed bugs, there's a few left. Now the syntax tests for almost all languages pass perfectly. The following languages still fail:

- [ ] ASP: not sure why, think it is some weird thing with `with_prototype`. Speculation in #51 
- [ ] LaTeX: mostly works, but one pattern matches the empty string, should be fixed in Sublime Packages. Notes in #51 by @keith-hall 
- [ ] Markdown: one test fails on `<a name=\"demo\"></a>`. Not sure why. Might be that https://github.com/sublimehq/Packages/pull/877 doesn't actually work correctly. Not sure if @keith-hall ever ran the syntax tests on it.
- [ ] PHP: a number of failures, not sure why. Might be the same thing as ASP.

If you plan on attempting to diagnose or fix one of these bugs, create an issue for it and reference it here. I will do the same.

I don't foresee myself personally doing much work on this for the next little while, but I will review code and discuss and may get around to it eventually.

keith-hall:
I'm investigating the Markdown test failure under https://github.com/trishume/syntect/issues/62

keith-hall:
I've logged a new issue at https://github.com/trishume/syntect/issues/104 which details why the ASP syntax tests are still failing, and an additional bug which doesn't show up in the official ST default packages repo.

keith-hall:
I think the reason why the LaTeX syntax has a failure in `syntect` is due to how ST and `syntect` optimize the matching - ST seems to check "does this regex pattern match at this point", and if no patterns match, it advances the "point" to the next character and does all the work again, whereas `syntect` caches the *next* match position for each regex. When there are dodgy regex patterns in the syntax definition that consume no characters, `syntect` doesn't see that actually there is a match further along that *will* consume characters, and instead just moves on to the next pattern match that "does something".

i.e. we can see from the operations that the pattern that matches the empty string matches before the close brace pattern that pops the context

<details><summary>

```
blank match `[A-Za-z[:digit:]-]*` at column 27 on line `\usepackage[args]{mypackage, anotherpackage}`
```
</summary>

```
\usepackage[args]{mypackage, anotherpackage}
^ +meta.preamble.usepackage.latex
\usepackage[args]{mypackage, anotherpackage}
^ +keyword.control.preamble.latex
\usepackage[args]{mypackage, anotherpackage}
^ +punctuation.definition.backslash.latex
[...]

PR #103, opened 2017-09-05
Improve with_prototype to better match Sublime. Fixes #101

### LINKED ISSUE #97: Difference in highlighting some TypeScript React files (Sublime vs. Syntect)
Syntect fails to properly syntax highlight some TypeScript React (tsx) files in contrast to Sublime Text 3 which highlights them fine.

To reproduce this:

1. You'll need to first download Microsoft's official [TypeScriptReact.tmLanguage](https://raw.githubusercontent.com/Microsoft/TypeScript-TmLanguage/master/TypeScriptReact.tmLanguage) file and save it somewhere.
2. Convert that `.tmLanguage` file to a `.sublime-syntax` file, using Sublime Text 3's builtin `Plugin Development: Convert Syntax to .sublime-syntax` command (important: only shows up when viewing `.tmLanguage` files)
3. Save the new `.sublime-syntax` file into e.g. `testdata/Packages/TypeScript/TypeScriptReact.sublime-syntax`
4. Run `make assets`.

Alternatively, [my fork's branch here](https://github.com/slimsag/syntect/tree/issue-97) has the above steps done so you don't have to perform them. The `.sublime-syntax` file is also in the root of the repo. 

Once you've done the above or cloned my fork,

1. Save this [BasePicker.tsx](https://raw.githubusercontent.com/OfficeDev/office-ui-fabric-react/5e9fbcd5735cc185c1b2e047beb517cb9f1dbe96/packages/office-ui-fabric-react/src/components/pickers/BasePicker.tsx) test file anywhere.
2. Save [TypeScriptReact.sublime-syntax](https://raw.githubusercontent.com/slimsag/syntect/issue-97/TypeScriptReact.sublime-syntax) into e.g. `~/Library/Application\ Support/Sublime\ Text\ 3/Packages/User/` on Mac (and uninstall the TypeScript Sublime plugin, to prevent any concern over the tmLanguage -> sublime-syntax conversion potentially being the issue).

When the file is viewed in Sublime, it looks like this:

image

In contrast, the HTML rendered by `cargo run --example synhtml ./BasePicker.tsx > out.html` ends up looking like this:

image

Of course, the above two look different due to the theme in use, but notice that even just in the syntect result alone there are two different colors in use for the attributes:

image

I found that changing this line of code to remove the addition operator:

```diff
                suggestedDisplayValue={ suggestedDisplayValue }
-               aria-activedescendant={ 'sug-' + this.suggestionStore.currentIndex }
+               aria-activedescendant={ this.suggestionStore.currentIndex }
                aria-owns='suggestion-list'
```

Fixes the highlighting in Syntect:

image

If it wasn't for Sublime rendering this file with the same sublime-syntax, I would assume the syntax file was broken somehow -- but given the difference between Sublime and Syntect's rendering, I assume this must be a bug in Syntect (but I do not understand Syntect well enough to even begin tracking it down :) )

### LINKED ISSUE #101: with_prototype syntax test failures
I've finally had chance to investigate https://github.com/trishume/syntect/issues/59 a little bit, and found something interesting.

Let's say we have a file `testdata/Packages/ASP/HTML-ASP2.sublime-syntax` with the following contents (it's basically a cut down version of the full HTML-ASP syntax, that doesn't reference ASP itself and can serve as a(n almost minimal) complete verifiable example):

```yaml
%YAML 1.2
---
# http://www.sublimetext.com/docs/3/syntax.html
name: HTML (ASP)2
file_extensions:
  - asp2
scope: text.html.asp2
contexts:
  main:
    - include: html

  asp_punctuation_begin:
    - match: '<%'
      scope: punctuation.section.embedded.begin.asp
      push:
        - match: '(?=\S)'
          set: [close_embedded_asp, begin_embedded_asp]

  html:
    - match: ''
      set:
        - include: scope:text.html.basic
      with_prototype:
        - include: asp_punctuation_begin

  begin_embedded_asp:
    - meta_content_scope: source.asp.embedded.html
    - match: '(?=%>)'
      pop: true

  close_embedded_asp:
    - match: '%>'
      scope: punctuation.section.embedded.end.asp
      pop: true
```

and the following corresponding syntax test (`testdata/Packages/ASP/syntax_test_asp2.asp2`):

```ASP
' SYNTAX TEST "Packages/User/HTML-ASP2.sublime-syntax"
<html><% example <%
' ^^^ meta.tag.structure.any.html entity.name.tag.structure.any.html
'     ^^ punctuation.section.embedded.begin.asp
'        ^^^^^^^^^^^ source.asp.embedded.html - punctuation
```

and execute `syntest` with `RUST_BACKTRACE=1 cargo run --example syntest testdata/Packages/ASP/syntax_test_asp2.asp2 testdata/Packages/`

The output is:

```
loading syntax definitions from testdata/Packages/
Testing file testdata/Packages/ASP/syntax_test_asp2.asp2
The test file references syntax definition file: Packages/ASP/HTML-ASP2.sublime-syntax
  Assertion selector "source.asp.embedded.html - punctuation" from line 5 failed against line 2, column range 17-19 (with text "<%") has scope [<text.html.asp2>, <source.asp.embedded.html>, <punctuation.section.embedded.begin.asp>]
Ok(FailedAssertions(2, 16))
exiting with code 1
```

but in ST, if we look at the scopes, we see the `<%` which had a failure in syntect has no special scopes applied to it.
image

so we can conclude that the `with_prototype` is being applied in syntect to the `begin_embedded_asp` context, but isn't in ST.  I suspect there is a general rule that the `with_prototype` shouldn't apply to nested contexts pushed by matches made in itself, if that makes sense.

i.e. estimated rough sequence of events:

- the anonymous context inside the `html` context is pushed onto the stack, with the `asp_punctuation_begin` context in the `with_prototype`
- eventually the `<%` match pattern from that context is hit, and it pushes a new anonymous context - *this context should no longer have the `asp_punctuation_begin` `with_prototype` applied to it*
- the `begin_embedded_asp` context replaces the anonymous context at the top of the stack
- because syntect is still applying the `with_prototype`, the `<%` after `example` gets matched unexpectedly/incorrectly and scoped.

trishume:
I did a bunch of testing and figured on the details of Sublime's `with_prototype` behaviour around cases like described in #101. 

This PR implements a fix that:
1. Corrects the order I was iterating `with_prototypes` in the stack. It is now bottom to top like Sublime.
2. Doesn't use `with_prototype` contexts from stack frames lower than the last frame pushed by a `with_prototype` rule.

This reduces the number of failing syntax tests to 162 for ASP, 1 for LaTeX, and everything else passes. Previously ASP had 202 failing tests, fixing the order only got that down to 183. PHP also used to have 40 failing tests, although that jumped to 104 when I fixed the order, but that was just further exposing the other bug that when fixed brought it down to 0 failures. So this fixes #101 but not yet #59.

### TODO
- [x] Test what this does to #97. **Edit:** NVM the Typescript syntax mentioned doesn't even use `with_prototype`.
- [x] Benchmark for significant performance regressions. **Edit:** no regression.

cc @keith-hall

### DIFF
--- a/src/parsing/parser.rs
+++ b/src/parsing/parser.rs
@@ -129,7 +129,7 @@ impl ParseState {
                 ctx_ref.prototype.clone()
             };
             let context_chain = self.stack
-                .iter().rev() // iterate the stack in top-down order to apply the prototypes
+                .iter() // Sublime applies with_prototypes from bottom to top
                 .filter_map(|lvl| lvl.prototype.as_ref().cloned())
                 .chain(prototype.into_iter())
                 .chain(Some(cur_level.context.clone()).into_iter());
--- a/src/parsing/parser.rs
+++ b/src/parsing/parser.rs
@@ -27,6 +27,9 @@ use fnv::FnvHasher;
 pub struct ParseState {
     stack: Vec<StateLevel>,
     first_line: bool,
+    // See issue #101. Contains indices of frames pushed by `with_prototype`s.
+    // Doesn't look at `with_prototype`s below top of stack.
+    proto_starts: Vec<usize>,
 }
 
 #[derive(Debug, Clone, PartialEq, Eq)]
@@ -60,6 +63,7 @@ impl ParseState {
         ParseState {
             stack: vec![start_state],
             first_line: true,
+            proto_starts: Vec::new(),
         }
     }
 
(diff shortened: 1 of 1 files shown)


robinst:
Nice! Is there some description of the sublime-syntax format that we can expand to add this information? Would be good to have a better description of how they work, for future us but also other implementers.

trishume:
There's https://www.sublimetext.com/docs/3/syntax.html, but it's written for syntax-writers, not implementers. So it heavily underspecifies a lot of things. It's not easy to submit updates, and if it was it wouldn't be the right place.

Maybe a `.md` file in the Sublime packages repo, or on the wiki of that repo, would be the right place. Or even in syntect's repo or wiki.

Issue #63, opened 2017-04-10
Parsing C# file (RegexCharClass.cs) doesn't finish


robinst:
The [RegexCharClass.cs from the MonoDevelop sources](https://github.com/mono/monodevelop/blob/522676ea6126e743df6b108a194dc155f4e8519e/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor.Highlighting/RegexEngine/RegexCharClass.cs) takes a very long time to parse (killed it after a couple of minutes), with current master (f03fa0f672ab0a25ec93cddba11a8a6b036605f9):

    curl -LO "https://github.com/mono/monodevelop/raw/522676ea6126e743df6b108a194dc155f4e8519e/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor.Highlighting/RegexEngine/RegexCharClass.cs"

    cargo build --release --example syncat

    time ./target/release/examples/syncat RegexCharClass.cs

trishume:
This is probably also more catastrophic backtracking.

I should really look into Oniguruma options to bail out on bad cases.

robinst:
Looks like the lines with a lot of `\u1234` escapes are the problem: https://github.com/mono/monodevelop/blob/522676ea6126e743df6b108a194dc155f4e8519e/main/src/core/MonoDevelop.Ide/MonoDevelop.Ide.Editor.Highlighting/RegexEngine/RegexCharClass.cs#L210

robinst:
This PR should fix it by changing `\\u\h{,4}` to `\\u\h{4}` (which is also in line with the spec): https://github.com/sublimehq/Packages/pull/1132

robinst:
Ah, this has been fixed by the change in Packages. Forgot to close the issue :).

PR #108, opened 2017-09-27
Update pack dumps

robinst:
Includes the [fix](https://github.com/sublimehq/Packages/pull/1132) for #63 and other improvements.

This one is nice to have as well: https://github.com/sublimehq/Packages/pull/1115


trishume:
Thanks!

Side note: someday I'll probably want to find a better solution to the pack files than putting them in the repo, so that cloning doesn't end up taking forever, maybe Git LFS. But for now it's still fine.

robinst:
You mean the `default_newlines.packdump` and `default_nonewlines.packdump` files? The repo is currently 17 MB, so that's still ok.

I would be hesitant to introduce Git LFS because it's a bit painful to set up for people and might be a hurdle for contribution (more so than the initial clone taking a bit longer).

PR #111, opened 2017-09-28
Upgrade bitflags to 1.0

robinst:
bitflags 1.0 requires Rust 1.20 which stabilized associated constants.
This change also makes the black and white Colors consts.

How do you feel about requiring Rust 1.20? If you want to support stable and stable-1, we can wait with this.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ onig = { version = "^1.2", optional = true }
 walkdir = "^1.0"
 regex-syntax = { version = "^0.4", optional = true }
 lazy_static = "^0.2"
-bitflags = "^0.8"
+bitflags = "1.0"
 plist = "0.2"
 bincode = { version = "0.8", optional = true }
 flate2 = { version = "^0.2", optional = true }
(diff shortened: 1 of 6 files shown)


trishume:
Hmmm, I'm inclined to accept this but I haven't made up my mind about any version support policy yet. I know that at the very least https://fuchsia.googlesource.com/ depends on syntect indirectly through Xi, and probably hasn't upgraded enough yet, but I think they're going to be on an old version of Xi anyhow for unrelated reasons for a while, so it probably doesn't matter.

This PR also changes a public API. The `WHITE` and `BLACK` constants probably nobody uses, but I don't know about the font style ones.

I think I'm going to leave this unmerged and sit on it for a while until I either make up my mind, decide to release v2.0 for another breaking change, or it becomes long enough that Rust releases a new version.

robinst:
Upgrading bitflags is already a breaking change because the font constants are no longer defined as top-level items, but as associated constants.

But to avoid breaking API for now we could also define them as top-levels (in addition to the associated consts) I think, same for the colors. Maybe we can mark them as deprecated too. What do you think?

trishume:
@robinst sure, I think I would be willing to merge a version of this PR that required a new version of Rust but didn't have the breaking changes, using deprecated top-level defines.

robinst:
Cool, done. I used "1.8.0" as the next version number, not sure if that's what you wanna bump to (because of the changed packs?).

PR #112, opened 2017-09-28
Switch to docs.rs for documentation (#110)

### LINKED ISSUE #110: Switch docs to use docs.rs
Before I raise any PRs, I was wondering if you're ok with switching the docs to use docs.rs instead of hosting them yourself. That has a couple of advantages:

* It makes the build simpler (we can just remove that step)
* It means the docs are available for older versions as well
* We can add a badge for it

There's disadvantages as well:

* It's run by someone other than you :). But given that so many crates use it (including crates by the rust-lang team), the Rust community has a pretty big interest in keeping the site up.

You can check out the docs here: https://docs.rs/syntect/

robinst:
Also add `html_root_url` attribute so that dependent crates get correct
links, see [section in Rust API guidelines](https://rust-lang-nursery.github.io/api-guidelines/documentation.html#crate-sets-html_root_url-attribute-c-html-root), quote:

> It should point to "https://docs.rs/CRATE/MAJOR.MINOR.PATCH", assuming the crate uses docs.rs for its primary API documentation.
> 
> The html_root_url attribute tells rustdoc how to create URLs to items in the crate when compiling downstream crates. Without it, links in the documentation of crates that depend on your crate will be incorrect.

(Fixes #110)


Cargo.toml
```diff
@@ -1,13 +1,13 @@
...
 description = "library for high quality syntax highlighting and code intelligence using Sublime Text's grammars"
-documentation = "http://thume.ca/rustdoc/syntect/syntect/"
+documentation = "https://docs.rs/syntect"
 repository = "https://github.com/trishume/syntect"
 keywords = ["syntax", "highlighting", "highlighter", "colouring", "parsing"]
 categories = ["parser-implementations", "parsing", "text-processing"]
 readme = "Readme.md"
 license = "MIT"
-version = "1.7.3"
+version = "1.7.3" # remember to update html_root_url
```
trishume: Thanks for adding the comment, I was just about to remark that I'll have to remember to do that

Readme.md
```diff
@@ -10,8 +13,6 @@ I won't be committing much anymore because the marginal return on additional wor
 I'll also quite possibly still fix issues and definitely offer advice and knowledge on how the library works. Basically I'll be maintaining the library but not developing it further.
 I've spent months working on, tweaking, optimizing, documenting and testing this library. If you still have any reasons you don't think it fits your needs, file an issue or email me.
 
-### Rendered docs: <http://thume.ca/rustdoc/syntect/syntect/>
```
trishume: It's not a huge deal, but I think I'd prefer to keep a big obvious header with a link to the docs, since they're super important and unlike some other crates syntect actually has some.

robinst: Sure, done

PR #114, opened 2017-10-12
Bincode no longer exposes ErrorKind::IoError.

brandonedens:


TyOverby:
This was caused by a breaking change that was accidentally published.  I've yanked that version of the crate, so the build should work again.

cmyr:
@TyOverby Okay that makes sense, I was very confused there for a day. Thank you for the quick fix.

trishume:
@TyOverby do you recommend I still merge this PR or not? It seems that this PR just switches to relying on `bincode` providing a `From` impl for `io::Error` instead of using an explicit variant. What do you recommend we do?

TyOverby:
The PR is no longer necessary, but I think that the code looks nicer 👍

Bincode will not remove the From implementation.

PR #116, opened 2017-10-15
update the InspiredGitHub theme

eyelash:
This fixes Markdown punctuations.


trishume:
Thanks! Note that the tests are failing since one of the HTML tests relies on that theme. You can fix the test by just copy-pasting the output into the source file, then opening it in your browser to check it looks right.

eyelash:
fixed

PR #119, opened 2017-10-18
README: Add codecov badge

robinst:


robinst:
Oh, I didn't realize that I pushed the branch to this repo. Would you prefer branches to be in my fork or is that ok?

trishume:
Either one works for me. If you push to this repo I'll delete your branches when I merge, if you use your fork, you can do it on your own time, your choice which you prefer.

PR #122, opened 2017-11-11
Bump onig to 2.0

est31:
The bump is needed so that I can get my documentation failure fix https://github.com/rust-onig/rust-onig/pull/61 . The minimum required Rust version is still 1.20 so no major bump is needed. I'd really love a minor or patch version bump though!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ exclude = [
 
 [dependencies]
 yaml-rust = { version = "0.3.5", optional = true }
-onig = { version = "^1.2", optional = true }
+onig = { version = "^2.0", optional = true }
 walkdir = "^1.0"
 regex-syntax = { version = "^0.4", optional = true }
 lazy_static = "^0.2"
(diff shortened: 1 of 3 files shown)


trishume:
Thanks!

trishume:
I pushed v1.8.2 with this change

Cargo.toml
```diff
@@ -49,7 +49,6 @@ dump-create = ["flate2", "bincode"]
 # Pure Rust dump creation, worse compressor so produces larger dumps than dump-create
 dump-create-rs = ["libflate", "bincode"]
 
-static-onig = ["onig/static-libonig"]
```
epage: Isn't this a breaking change?

epage: For anyone else wondering, the removal of this doesn't seem to be because `onig` static-links by default but because `onig` now only supports static linking through environment variable.

See https://github.com/rust-onig/rust-onig/commit/679348dc2a1966e0d748378a8ef365e2597e538a#diff-80398c5faae3c069e4e6aa2ed11b28c0

est31: @epage sadly you are right, it is a breaking change technically. I guess if you want to be super exact about this you'd want to yank the 1.8.2 release and make a 2.0.0 release, or alternatively yanking 1.8.2 as well and geting onig to re-add a cargo feature and then releasing a 1.8.3 version depending on the onig that has the cargo feature again, but no idea how @trishume feels about that.

I have just wanted to bump onig because documentation *failed* with it on rust nightly so I sent them a patch and it only got added to the 2.0 branch...

epage: Yeah, I caught this when trying to do `cargo update` on a project where we aim for a single-distributable binary.

trishume: @epage @est31 So I yanked 1.8.2 pending decision on what to do. I think the cargo feature works better for projects that want to build static binaries, and I imagine @epage would agree. So I'm inclined to raise an issue on `onig` asking them to re-add the cargo feature.

PR #123, opened 2017-11-11
Greedily merge consecutive spans on the same line

est31:
This merges two or more consecutive spans of the same color
that are on the same line. If the span on the right is
a whitespace, and the background colours are identical,
a merger happens as well.

This allows us to save about 50% of character count in the HTML
output.


est31:
The algorithm is very simple, yet it obtains very good results!

It does two things:

a) merge consecutive spans of the same style
b) merge consecutive spans if the background colours match and the second span only consists of whitespaces.

I haven't optimized b) further for checking whether the first span only consists of whitespaces, as then you'd have to wait with writing the span start until you know the full length of the span. This would make the algorithm a bit more complicated :)

The merging strategy can be still be improved:

* by merging across lines, but this would require me to break the way the functions call each other
* by creating a tree-like structure, but this would require me to come up with a less simple algorithm.

The 50% claim comes from the following Rust code snippet:

```rust
extern crate piston_window;

use piston_window::*;

fn main() {
    let mut window: PistonWindow =
        WindowSettings::new("Hello Piston!", [640, 480])
        .exit_on_esc(true).build().unwrap();
    while let Some(event) = window.next() {
        window.draw_2d(&event, |context, graphics| {
            clear([1.0; 4], graphics);
            rectangle([1.0, 0.0, 0.0, 1.0], // red
                      [0.0, 0.0, 100.0, 100.0],
                      context.transform,
                      graphics);
        });
    }
}
```

Without this PR, syntect needs 6196 characters to display the snippet. With this PR applied, syntect needs 3136 characters. With only strategy a) applied, it needs 3388 characters.

trishume:
Thanks for fixing the debug print diffs in the test. I know CI will definitely pass but I'll wait for it anyway and then merge.

Let me know whenever you'd like me to release a change immediately as opposed to waiting to batch it with other changes.

est31:
@trishume thanks for reviewing my PR! Feel free to batch the PR with future changes. I have copied the code locally and will most likely implement cross-line merging or further optimisations.

PR #125, opened 2017-11-19
Update dependencies

est31:
Most importantly, it updates the onig dependency to 3.0 [so that it links onig statically per default](https://github.com/rust-onig/rust-onig/issues/65).

I've tested using `cargo +1.20.0 test --no-default-features --features "dump-create-rs dump-load-rs parsing assets html yaml-load"` and `cargo +1.20.0 test`.

cc @epage

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,13 +15,13 @@ exclude = [
 
 [dependencies]
 yaml-rust = { version = "0.3.5", optional = true }
-onig = { version = "2.0", optional = true }
-walkdir = "1.0"
+onig = { version = "3.0", optional = true }
+walkdir = "2.0"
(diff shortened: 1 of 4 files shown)


est31:
@trishume right now I can rely on 1.8.2 in my Cargo.lock, so I don't need an immediate version bump after this PR got merged, so you can e.g. wait for #124 to be included into the new release (would probably still technically be a breaking change as it removes the cargo feature, so 2.0.0 instead of 1.9 or 1.8.3).

trishume:
Thanks!

epage:
Will major be bumped at some point?

trishume:
@epage Probably in early December, see my comment on #124

PR #128, opened 2017-12-09
Update lazy_static and yaml-rust

est31:


est31:
With this PR, all dependencies should be up to date once we have the 2.0 release of syntect.

trishume:
Thanks!

PR #130, opened 2018-01-02
Bump to v2.0.0 with cleanup

trishume:
Breaking changes:
- The `static-onig` feature was removed, static linking is now the default
- Font styles and color constants now use associated consts because of bitflags upgrade
- SyntaxDefinition::load_from_str now has an extra parameter

Other notable changes:
- Support for new `embed` syntax #124
- Updates to many dependencies
- Updated dumps
- More compact HTML output

As well as bumping the version this commit does some cleaning up for the breaking changes.

Thanks to @Keats @keith-hall and @est31 for your significant contributions to this release!

### DIFF
--- a/src/parsing/syntax_set.rs
+++ b/src/parsing/syntax_set.rs
@@ -13,7 +13,11 @@ use std::fs::File;
 use std::ops::DerefMut;
 use std::mem;
 use std::rc::Rc;
+
+// TODO this is unused in nightly but used in stable, remove eventually
+#[allow(unused_imports)]
 use std::ascii::AsciiExt;
+
 use std::sync::Mutex;
 use onig::Regex;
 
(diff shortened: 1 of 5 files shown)


src/parsing/syntax_set.rs
```diff
@@ -13,7 +13,11 @@ use std::fs::File;
 use std::ops::DerefMut;
 use std::mem;
 use std::rc::Rc;
+
+// TODO this is unused in nightly but used in stable, remove eventually
```
est31: More precisely, in 1.23 (current beta, reaching stable soon, on Jan 4th) it is unused, and in 1.22.1 (current stable) it is used.

trishume:
I published a new release to crates.io and added some release notes to a Github Release.

PR #132, opened 2018-01-10
add example for scoring scope selectors

keith-hall:
This PR adds an example for scoring scope selectors, to make it easier to debug color schemes etc. when multiple selectors match a scope stack, and we want to see which selector has a higher precedence/score.

### DIFF
--- a/examples/score_selector.rs
+++ b/examples/score_selector.rs
@@ -7,6 +7,10 @@ use std::str::FromStr;
 
 fn main() {
     let args: Vec<String> = std::env::args().collect();
+    if args.len() == 1 {
+        println!("Expected use: {} \"scope stack\" \"scope selector\"", &args[0]);
+        return;
+    }
     // format the scope selector to include a space at the beginning, because, currently, ScopeSelector expects excludes to begin with " -"
     // but somebody might just type "-punctuation" on the commandline, for example
     let selector = ScopeSelectors::from_str(&format!(" {}", &args[2])).expect("Unable to parse scope selector");
--- a/examples/score_selector.rs
+++ b/examples/score_selector.rs
@@ -1,0 +1,22 @@
+//! An example of using syntect for comparing scope stacks against selectors.
+//! Useful for debugging color schemes etc. to see which selector has a higher precedence when multiple selectors match the same text/scope stack.
+extern crate syntect;
+use syntect::highlighting::ScopeSelectors;
+use syntect::parsing::ScopeStack;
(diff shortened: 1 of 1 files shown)


keith-hall:
Example use: `cargo run --example score_selector "text.plain" "text"`
Output:
> Scoring selector "text" against scope stack [<text.plain>]...
MatchPower(1)

or `cargo run --example score_selector -- "source.rust comment.line.documentation.rust" "-comment"
`
> Scoring selector "-comment" against scope stack [<source.rust>, <comment.line.documentation.rust>]...
The selector does not match the scope stack.

keith-hall:
this also addresses a bug whereby `syntect` wasn't scoring the best match when the scope stack contained multiple scopes whose prefixes would match a scope in the selector. As mentioned at https://manual.macromates.com/en/scope_selectors#ranking_matches, deeper scopes should give higher scores, but in `syntect`'s previous implementation, it would work from left to right and so it would match a scope less deep and score that, rather than looking for the deepest matching scope. (Note that Sublime is also affected by this, see https://github.com/SublimeTextIssues/Core/issues/2152.)

trishume:
Code looks good, and good idea, thanks!

However, this brings up the problem of whether I want to be bug-for-bug compatible with Sublime. So far I've tried to match Sublime's behaviour, and that's possibly even why that code is the way it is in syntect (although I may have just messed up in the same way). I've never had to rule on whether to replicate a bug before. I'd definitely also feel better if there was an official ruling on that issue that it really is a bug they intend to fix soon.

I also wonder if changing this would break compatibility with any themes. My head's not in the game enough right now to tell.

brupelo:
@keith-hall From this thread and this [one](https://github.com/SublimeTextIssues/Core/issues/2152) I also assume you're also interested on this subject... I've made some tests [here](https://github.com/trishume/syntect/issues/241) in order to figure out `score_selector` works internally but I haven't been able to figure out yet, do you know something about it?

keith-hall:
Nope, sorry, can't help with that. For my purposes, as long as it returns the correct results, how it does so is just an implementation detail that could change any time without really affecting me. If you do find out, it would be of academic interest to me ofc, but I'm happy enough knowing how it works in `syntect`.

Enselic:
I'm closing this PR for now as the open questions in https://github.com/trishume/syntect/pull/132#issuecomment-356791838 seems unaddressed, and since it is nice to keep the PR inbox clean, but of course feel very free to reopen this if I am closing it prematurely.

PR #140, opened 2018-03-11
Update link to rust-scoped-pool in Readme.md

beta-phenylethylamine:
Previous link was changed and just went to a 404 page.


trishume:
Thanks! Good catch.

PR #141, opened 2018-03-22
Replace libflate with rust_backend feature of flate2

messense:
flate2 already has an [pure Rust backend](https://github.com/alexcrichton/flate2-rs/pull/128).


trishume:
Thanks, I don't think that was the case when I originally used `libflate`.

Do you happen to know if flate2's Rust-based loading is faster or slower than `libflate`? It's not super important.

trishume:
Awesome, so it's actually faster than C miniz now. That's cool, thanks for the PR!

Issue #64, opened 2017-04-10
Parsing simple Groovy file takes 10 s


robinst:
The [UnmanagedNodeBackedModelMapTest.groovy file from the gradle sources](https://github.com/gradle/gradle/blob/aa8c200808824c61f2c46ed1eb948c759c9483c2/subprojects/model-core/src/test/groovy/org/gradle/model/UnmanagedNodeBackedModelMapTest.groovy) takes 10 seconds to parse, on current master (f03fa0f672ab0a25ec93cddba11a8a6b036605f9):

```
$ curl -LO "https://github.com/gradle/gradle/raw/aa8c200808824c61f2c46ed1eb948c759c9483c2/subprojects/model-core/src/test/groovy/org/gradle/model/UnmanagedNodeBackedModelMapTest.groovy"
$ cargo build --release --example syncat
$ time ./target/release/examples/syncat UnmanagedNodeBackedModelMapTest.groovy
...
    9.72s user 0.04s system 99% cpu 9.811 total
```

trishume:
Darn, I bet this is more catastrophic backtracking.

keith-hall:
This is interesting because there have basically been no changes to the Groovy syntax definition since the Sublime Text developer took it from TextMate, so I wonder about the performance in TextMate (and ST before the custom regex engine was implemented): https://github.com/sublimehq/Packages/commits/master/Groovy

blach:
FWIW, I had the same problem in my own Sublime Text syntax parser I'm currently working on.
I could fix the problem by enabling `USE_COMBINATION_EXPLOSION_CHECK` in `regint.h` in Onigmo/Oniguruma.

robinst:
Fixed in 2.0.1 by using a newer Onig which has the explosion check enabled by default, and handling it when we do matching.

PR #142, opened 2018-03-23
Recover from catastrophic backtracking

robinst:
Oniguruma 6.8.0 [enabled USE_TRY_IN_MATCH_LIMIT by default](https://github.com/kkos/oniguruma/blob/v6.8.0/HISTORY#L14), with a limit
of 10_000_000. That means after that number of failed matches, it will
stop trying to match and return an error instead.

We can gracefully handle that error by using the onig crate's new
`search_with_param` method which returns a Result instead of panicking.

I tried this on the file from #64 and the parse time goes from 10
seconds to 200 ms.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ exclude = [
 
 [dependencies]
 yaml-rust = { version = "0.4", optional = true }
-onig = { version = "3.0", optional = true }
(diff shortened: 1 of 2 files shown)


robinst:
@trishume onig has been released with the change and I updated this PR. What do you think?

Issue #127, opened 2017-12-01
syntect handles push & immediately pop non-consuming patterns differently to ST


keith-hall:
In ST, the following syntax:

```yaml
%YAML 1.2
---
# See http://www.sublimetext.com/docs/3/syntax.html
name: push_pop_test
scope: source.push_pop_test
contexts:
  main:
    - match: (?=hello)
      push: test
  test:
    - match: (?!world)
      pop: true
    - match: \w+
      scope: comment.matched
```

accompanied by the following test passes:

```
# SYNTAX TEST "push_pop_test.sublime-syntax"
hello
# ^^^ comment.matched
```

in syntect, it doesn't:

```
loading syntax definitions from testdata/push_pop_test.sublime-syntax
Testing file testdata/syntax_test_push_pop_test.txt
The test file references syntax definition file: push_pop_test.sublime-syntax
-- debugging line 1 -- scope stack: ScopeStack { clear_stack: [], scopes: [] }
# SYNTAX TEST "push_pop_test.sublime-syntax"
^ +source.push_pop_test
-- debugging line 2 -- scope stack: ScopeStack { clear_stack: [], scopes: [<source.push_pop_test>] }
no operations for this line...
  Assertion selector "comment.matched" from line 3 failed against line 2, column range 2-5 (with text "llo") has scope [<source.push_pop_test>]
Ok(FailedAssertions(3, 3))
exiting with code 1
```

I believe this is due to how ST handles a match that:
- consumes no characters and
- pushes one context
- where that context has a match which immediately pops without consuming any characters

In ST, it ignores the pop until at least one character has been consumed. Currently, syntect seems to push, then pop, then ignore the push again until at least one char has been consumed.

Related reading: https://github.com/SublimeTextIssues/Core/issues/1906#issuecomment-329536967

This is the cause for most of the C# syntax test failures in syntect alluded to in https://github.com/trishume/syntect/pull/124#issuecomment-348206852 since https://github.com/sublimehq/Packages/commit/e29d674d4f7ac1834c51777394a712964da992f3#diff-9b5e53d4e14d87efc8f78017464d6b94.

keith-hall:
I've made a PR to the Packages repo that fixes the C# syntax so that all syntax tests pass in syntect: https://github.com/sublimehq/Packages/pull/1359

trishume:
[...]

PR #144, opened 2018-04-18
Extract parts of parse_next_token into new method

robinst:
It was hard to figure out what was going on with that much indentation
and vertical scrolling. Extracting the search code into a new method
helped me understand it a bit better. It also means we can use `return`
instead of `continue` in the for loop.

### DIFF
--- a/src/parsing/parser.rs
+++ b/src/parsing/parser.rs
@@ -169,88 +169,27 @@ impl ParseState {
             // println!("token at {} on {}", start, line.trim_right());
 
             let mut min_start = usize::MAX;
-            let mut match_from_with_proto = false;
             let mut cur_match: Option<RegexMatch> = None;
+            let mut match_from_with_proto = false;
 
             for (from_with_proto, ctx, captures) in context_chain {
                 for (pat_context_ptr, pat_index) in context_iter(ctx) {
                     let mut pat_context = pat_context_ptr.borrow_mut();
                     let match_pat = pat_context.match_at_mut(pat_index);
-                    // println!("{} - {:?} - {:?}", match_pat.regex_str, match_pat.has_captures, cur_level.captures.is_some());
-                    let match_ptr = match_pat as *const MatchPattern;
 
-                    // Avoid matching the same pattern twice in the same place, causing an infinite loop
-                    if matched.contains(&match_ptr) {
-                        continue;
-                    }
-
-                    if let Some(maybe_region) =
-                           search_cache.get(&match_ptr) {
-                        let mut valid_entry = true;
-                        if let Some(ref region) = *maybe_region {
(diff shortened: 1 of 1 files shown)


src/parsing/parser.rs
```diff
@@ -275,6 +214,79 @@ impl ParseState {
...
+        if let Ok(Some(match_start)) = matched {
+            let match_end = regions.pos(0).unwrap().1;
+            // this is necessary to avoid infinite looping on dumb patterns
+            let does_something = match match_pat.operation {
+                MatchOperation::None => match_start != match_end,
+                _ => true,
+            };
+            if refs_regex.is_none() && does_something {
+                search_cache.insert(match_pat, Some(regions.clone()));
+            }
```
robinst: What's the reason for only inserting it into the cache if `does_something` is true? I kept the current logic, but I'm not sure I understand.

robinst:
(I've run syntest before and after these changes and it's unchanged.)

src/parsing/parser.rs
keith-hall: I would guess that sometimes people write syntax definitions with silly rules like a regex that can match nothing (and doesn't change the context stack):

```yaml
- match: a*
  scope: some.scope
```

and if it cached places where it matched nothing, it'd never find the places where it matches something, i.e. it needs to repeat the match at a different position to find a real non-zero length match.

robinst: It makes sense not to use that match, but caching it should be ok. The way it works is that if there's a cached match, it is only used if the start position of it is `>=` the current position where we're trying to search. Otherwise, we search again using that pattern.

Just now, I tried removing `&& does_something` from the condition and both tests and syntest still pass. But maybe there's no test for this case (with caching)?

trishume:
Actually can you run the jQuery benchmark on old and new and make sure this doesn't regress performance? And if it does regress performance a `#[inline]` might fix it.

src/parsing/parser.rs
trishume: On second look maybe the `does_something` check is unnecessary now. I think I might have added it before the more general infinite loop protection, and Sublime may just have the more general protection.

robinst: Yeah. But I think how this works might change with the fix for #127 anyway (maybe).

robinst:
I got this result for a run:

```
test bench_highlighting_jquery  ... bench: 798,202,387 ns/iter (+/- 366,799,381)
```

But the +/- numbers there don't make me confident that I'll get a meaningful result. I'm looking into converting the benchmarks to use [criterion.rs](https://github.com/japaric/criterion.rs) instead :).

robinst:
I benchmarked this with criterion and it doesn't show a performance change.

Haven't had time to raise a PR for using criterion, will do tomorrow.

PR #145, opened 2018-04-21
Switch benchmarks to criterion

robinst:
Advantages over the previous system:

* Allows running on stable Rust
* Compares benchmarks between runs using statistics

See https://github.com/japaric/criterion.rs for more.

### DIFF
--- a/benches/highlighting.rs
+++ b/benches/highlighting.rs
@@ -21,13 +21,23 @@ fn do_highlight(s: &str, syntax: &SyntaxDefinition, theme: &Theme) -> usize {
     count
 }
 
-fn highlight_file(b: &mut Bencher, path_s: &str) {
+fn highlight_file(b: &mut Bencher, file: &str) {
+    let path = match file {
+        "highlight_test.erb" => "testdata/highlight_test.erb",
+        "InspiredGitHub.tmTheme" => "testdata/InspiredGitHub.tmtheme/InspiredGitHub.tmTheme",
+        "Ruby.sublime-syntax" => "testdata/Packages/Ruby/Ruby.sublime-syntax",
+        "jquery.js" => "testdata/jquery.js",
+        "parser.rs" => "testdata/parser.rs",
+        "scope.rs" => "src/parsing/scope.rs",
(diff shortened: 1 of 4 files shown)


benches/highlighting.rs
```diff
@@ -29,49 +31,40 @@ fn highlight_file(b: &mut Bencher, path_s: &str) {
...
+            "testdata/jquery.js",
+            "testdata/parser.rs",
+            "src/parsing/scope.rs",
+        ]
+    );
+}
+
+criterion_group!{
+    name = benches;
+    config = Criterion::default().sample_size(10);
```
robinst: Note that more samples is better, but it takes a much longer time to run. But this can be adjusted if the measured runtime varies too much between runs of `cargo bench`.

benches/loading.rs
```diff
@@ -1,49 +1,56 @@
...
+    c.bench_function("load_internal_dump", bench_load_internal_dump);
+    c.bench_function("load_internal_themes", bench_load_internal_themes);
+    c.bench_function("load_theme", bench_load_theme);
+    c.bench_function("load_syntaxes", bench_load_syntaxes);
+    c.bench_function("link_syntaxes", bench_link_syntaxes);
+}
+
+criterion_group! {
+    name = benches;
+    config = Criterion::default().sample_size(100);
```
trishume: This sample size is too high, the `load_syntaxes` benchmark has an estimated time of 10 minutes, and others are pretty long.

robinst: Fixed!

robinst:
Good pickup with the results (I only looked at the output, not the files).

I'll look into it tomorrow and either raise an issue or a fix. To work around it in syntect we could instead use numerical indexes or simpler strings and then look up the test file in a Map. What do you think?

trishume:
Yah a string map seems like a fine workaround, numerical indexes would probably lead to the files being numbers and thus harder to navigate.

robinst:
Done using a match, and raised an issue for criterion: https://github.com/japaric/criterion.rs/issues/142

Issue #83, opened 2017-06-27
Refactor to use an arena


trishume:
This would be a large project, and may never happen, but the architecture would work much better with an arena of `Context`s instead of using `Rc<RefCell<>>` everywhere. Not only would the code be cleaner, but the lifetimes of `SyntaxDefinition` and `SyntaxSet` would be more correct, avoiding the footgun of dropping a `SyntaxSet` and keeping a cloned `SyntaxDefinition` that no longer has references to nested languages.

This would look something like using a `Vec<Context>` and indices everywhere.

It may also be easier to parallelize and make thread-safe, but that still requires extra work because of lazy regex compilation. See extensive discussion in #20.

williamyaoh:
Re: lazy regex compilation: Just the change to an arena of some sort should be enough to make `SyntaxSet` be `Send`; a `Vec<RefCell<Context>>` would do the trick without needing to change the existing *logic* that much. This also opens up the possibility of storing `&RefCell<Context>` after linking syntaxes instead of indices, avoiding adding an extra layer of indirection on each `Context` link from the current implementation. 

It would be a bit too much to hope to combine `Send` with lazy regex compilation; that would pretty much require synchronizing thread-shared mutability on `Context`s, so using `Mutex`/`RwLock` or friends.

williamyaoh:
Disregard that part about using `&RefCell` instead of numerical indices; it would be implementable, but would preclude a `Send` implementation. However, using a `Vec<RefCell<Context>>` to store a given `SyntaxSet`'s regexes + numerical indices as context links would still be enough to make it `Send`.

trishume:
You're right, it would have the advantage of making it `Send`. I have a plan that could make it even `Sync` in #20: separate the compiled Regexes into a separate hash table and automatically pool those hash tables but access everything else between threads immutably. This is similar to what the Regex library does, I think I describe it more in #20.

robinst:
Hey, looked into this a bit. I have some thoughts and wanted to check if they make sense:

Should we split the types so that "syntax definition read from yaml" and "linked syntax definition" are different types? Let's call the first one `SyntaxDefinition` and the second `SyntaxReference`. We could then have a `SyntaxSetBuilder` where you can add `SyntaxDefinition`, but when you call
`build`, you get a `SyntaxSet` (where all the contexts are linked). `SyntaxSet` and `SyntaxReference` is what you would pass to `ParseState`.

[...]

PR #146, opened 2018-04-24
Change loop detection to work like Sublime Text

robinst:
This fixes issue #127. There's now 575 fewer failing assertions when
running syntest. The following test files no longer have any failing
assertions:

* syntax_test_bash.sh
* syntax_test_c#.cs
* syntax_test_C#7.cs
* syntax_test_GeneralStructure.cs
* syntax_test_Generics.cs
* syntax_test_Operators.cs
* syntax_test_Using.cs

See the added comment before ParseState for more details.

### DIFF
--- a/src/parsing/parser.rs
+++ b/src/parsing/parser.rs
@@ -2,7 +2,7 @@ use super::syntax_definition::*;
 use super::scope::*;
 use onig::{MatchParam, Region, SearchOptions};
 use std::usize;
-use std::collections::{HashMap, HashSet};
+use std::collections::HashMap;
 use std::i32;
 use std::hash::BuildHasherDefault;
 use std::ptr;
@@ -27,10 +27,11 @@ use fnv::FnvHasher;
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct ParseState {
     stack: Vec<StateLevel>,
-    first_line: bool,
     // See issue #101. Contains indices of frames pushed by `with_prototype`s.
     // Doesn't look at `with_prototype`s below top of stack.
     proto_starts: Vec<usize>,
+    // The line being parsed (starting at 0)
+    line: usize,
 }
 
 #[derive(Debug, Clone)]
@@ -38,6 +39,7 @@ struct StateLevel {
     context: ContextPtr,
     prototype: Option<ContextPtr>,
     captures: Option<(Region, String)>,
+    non_consuming_push_at: Option<(usize, usize)>,
 }
 
 fn context_ptr_eq(a: &ContextPtr, b: &ContextPtr) -> bool {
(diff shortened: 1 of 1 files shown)


robinst:
Here's a gist with before/after and the diff for syntest 🎉 : https://gist.github.com/robinst/df79211f81694dce8a913fd5990a3b51

src/parsing/parser.rs
```diff
@@ -171,33 +257,74 @@ impl ParseState {
...
                             });
                             match_from_with_proto = from_with_proto;
+                            pop_would_loop = check_pop_loop && !consuming && match match_pat.operation {
+                                MatchOperation::Pop => true,
+                                _ => false,
+                            };
                         }
                     }
                 }
             }
```
robinst: Unrelated to this pull request, but I think there's some potential for avoiding some unnecessary matching here. If we found a match with `match_start == *start` (and it's not a looping pop), we can stop trying more patterns, as we'll not find a better match. Or am I missing something? I haven't tried it yet but I'll experiment with it and see if it makes a difference.

src/parsing/parser.rs
```diff
@@ -171,33 +257,74 @@ impl ParseState {
...
+            if pop_would_loop {
+                // A push that doesn't consume anything (a regex that resulted
+                // in an empty match at the current position) can not be
+                // followed by a non-consuming pop. Otherwise we're back where
+                // we started and would try the same sequence of matches again,
+                // resulting in an infinite loop. In this case, Sublime Text
+                // advances one character and tries again, thus preventing the
+                // loop.
+
+                // println!("pop_would_loop for match {:?}, start {}", reg_match, *start);
```
robinst: This and other printlns helped me understand why some things didn't work on the way to this solution. But I wonder if we should use something else here that's simpler to toggle, such as logging that can be enabled/disabled at compile-time (so there's no runtime overhead).

Also, it would help a lot if the context structs remembered their names (if they have one), that would make it much easier to understand where a match is coming from.. Was that considered or is the thinking that it would blow up the size of the structs for something that's only used when debugging?

src/parsing/parser.rs
```diff
@@ -220,24 +348,18 @@ impl ParseState {
               match_pat: &mut MatchPattern,
               captures: Option<&(Region, String)>,
               search_cache: &mut SearchCache,
-              matched: &mut MatchedPatterns,
               regions: &mut Region)
```
robinst: Is it so expensive to create a new `Region` that it's worth it to pass in an existing one to use? In case there was a match, it gets cloned later anyway.

src/parsing/parser.rs
```diff
@@ -701,6 +889,320 @@ contexts:
         expect_scope_stacks(&line, &expect, syntax);
     }
 
+    #[test]
+    fn can_parse_issue120() {
```
robinst: This one was just moved from the bottom.

src/parsing/parser.rs
```diff
@@ -701,6 +889,320 @@ contexts:
...
+        let expect2 = [
+            "<meta.tag.style.begin.html>, <punctuation.definition.tag.end.html>",
+            "<source.css.embedded.html>, <test.embedded>",
+            "<top-level.test>",
+        ];
+        expect_scope_stacks_with_syntax(&line2, &expect2, syntax.to_owned());
+    }
+
+    #[test]
+    fn can_parse_non_consuming_pop_that_would_loop() {
```
robinst: This might seem like a lot of test cases, but most of them came out of running syntest on Sublime's syntax tests, looking at a failure, finding the problem and reducing it to a minimal test case in here.

@keith-hall: I wonder if some of these would be useful to have in [sublimehq/Packages](https://github.com/sublimehq/Packages/) somewhere, to help people understand how matching works?

robinst:
@keith-hall It would be cool if you could have a look to see if the changes make sense. It seems like you have the most experience when it comes to understanding how Sublime Text's matching works :).

keith-hall:
Looks good to me, nice work @robinst ! You even added test cases with multiple pushes - when I logged the issue reporting the differences, my biggest concern was that this could get broken when our implementation changes, but you've covered it perfectly :)

src/parsing/parser.rs
keith-hall: Probably it would make sense for SublimeHQ to document it at http://www.sublimetext.com/docs/3/syntax.html - or, hopefully, they will open source some documentation pages soon for us to submit PRs to.

src/parsing/parser.rs
```diff
@@ -63,11 +65,105 @@ struct RegexMatch {
...
+// we go into `test` and stay at position 0. Now, the best match is the rule
+// with "pop". But if we used that rule, we'd pop back to `main` and would still
+// be at the same position we started at! So this would be an infinite loop,
+// which we don't want.
+//
+// So what Sublime Text does in case a looping rule "won":
+//
+// * If there's another rule that matches at the same position and does not
+//   result in a loop, use that instead.
+// * Otherwise, go to the next position and continue matching in the current
```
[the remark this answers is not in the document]
robinst: Maybe what I wrote "continue matching in the current context" is confusing, I meant "go through the rules in the current context again". So yes, if we get to the same "pop" at that point, it's no longer looping and we use it as normal.

I'll amend the text to make that more clear.

src/parsing/parser.rs
```diff
@@ -27,17 +27,19 @@ use fnv::FnvHasher;
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct ParseState {
     stack: Vec<StateLevel>,
-    first_line: bool,
     // See issue #101. Contains indices of frames pushed by `with_prototype`s.
     // Doesn't look at `with_prototype`s below top of stack.
     proto_starts: Vec<usize>,
+    // The line being parsed (starting at 0)
+    line: usize,
```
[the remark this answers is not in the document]
robinst: You're right! I actually had that at some point while trying to figure this thing out, but it didn't work because I didn't get it quite right (and I didn't fully understand how ST worked yet). But I went back now and redid it like that, and it's even simpler now :).

src/parsing/parser.rs
robinst: Yeah, thanks for your thoughts.. `Option<String>` also occurred to me, might be a nice way to do it. Regarding the context ID, I have the feeling that this might be the straightforward way to do it if we had #83.

src/parsing/parser.rs
robinst: So I tried this and it improves `time cargo run --release --example syntest > /dev/null` significantly:

```diff
-        5.59 real         5.30 user         0.24 sys
+        2.21 real         1.92 user         0.24 sys
```

cargo bench is less clear, but I guess it depends heavily on how the syntax definition is written, e.g. for jquery.js:

```
jquery.js               time:   [653.05 ms 667.12 ms 681.12 ms]
                        change: [-13.724% -11.616% -9.5689%] (p = 0.00 < 0.05)
                        Performance has improved.
```

It would be very interesting to have timings per syntax test in syntest, then we could see improvements per language.

I'm gonna raise another PR with this change after this one is merged.

src/parsing/parser.rs
```diff
@@ -321,11 +317,31 @@ impl ParseState {
...
             }
 
             let match_end = reg_match.regions.pos(0).unwrap().1;
-            *start = match_end;
+            let consuming = match_end > *start;
+            if consuming {
+                // The match consumes some characters. So update the position
+                // and clear state we use for checking for loops.
+                *start = match_end;
+                *non_consuming_push_at = None;
```
trishume: I don't think this is strictly necessary, since any future checks against it won't succeed since it will have an earlier start position. But it's totally fine to leave it, maybe makes it easier to think about.

I also realized that you can even avoid the logic of it being an `Option` because `(0,0)` is a starting state that should act correctly. Unsure of whether I prefer the cleaner logic of doing that or the perhaps more conceptually nice way of using an option.

src/parsing/parser.rs
```diff
@@ -321,11 +317,31 @@ impl ParseState {
...
+            } else {
+                // The match doesn't consume any characters. If this is a
+                // "push", remember the position and stack size so that we can
+                // check the next "pop" for loops. Otherwise leave the state,
+                // because a non-consuming "set" could also result in a loop.
+                let context = reg_match.context.borrow();
+                let match_pattern = context.match_at(reg_match.pat_index);
+                match match_pattern.operation {
```
trishume: This could be an `if let`

src/parsing/parser.rs
```diff
@@ -321,11 +317,31 @@ impl ParseState {
...
                 return true;
             }
 
             let match_end = reg_match.regions.pos(0).unwrap().1;
-            *start = match_end;
+            let consuming = match_end > *start;
+            if consuming {
+                // The match consumes some characters. So update the position
+                // and clear state we use for checking for loops.
+                *start = match_end;
```
trishume: This can be done unconditionally, combined with the next comment, this branch of the if is unnecessary. If you think it's easier to understand this way then I'm fine with it though.

src/parsing/parser.rs
```diff
@@ -321,11 +317,31 @@ impl ParseState {
                     return false;
                 }
                 *start += 1;
+                *non_consuming_push_at = None;
```
trishume: This is also unnecessary, see other similar comment for more.

src/parsing/parser.rs
robinst: Yes, all good ideas :). Done!

robinst:
All done!

robinst:
Cool! Would you mind creating a release with this and the other changes?

robinst:
Actually, let's first get the change in to abort early on matching, I'll raise a PR.

src/parsing/parser.rs
keith-hall: Thinking about it, we could maybe post some test cases at https://github.com/sublimehq/Packages/issues/757 to help people understand how the matching works, or depending on the outcome of https://github.com/sublimehq/Packages/issues/1522, on the wiki pages of that repo. My understanding is that, because Sublime Text directly bundles the contents of that repository with it's releases, the test cases would be less welcome in the repository code base directly, but I could be wrong.

Issue #94, opened 2017-08-12
Implement Error trait for error types


jmacdonald:
Implementing [Rust's canonical `Error` trait](https://doc.rust-lang.org/std/error/trait.Error.html) would allow errors in syntect to be chained with application-level errors, via [error-chain](https://github.com/rust-lang-nursery/error-chain).

Enselic:
For the record and FWIW: the error-chain crate has been deprecated. Thew new de-facto standard for libraries seems to be the thiserror crate. We recently replaced error-chain with thiserror in bat. See https://github.com/sharkdp/bat/pull/1820

Enselic:
I'd like to take a shot at this for 5.0.0.

PR #153, opened 2018-04-30
Implement Error for LoadingError

sharkdp:
Implements `std::error::Error` for `LoadingError`.

This is a first step towards #94. Things could definitely be improved by also implementing `Error` for the other syntect-specific errors.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -112,11 +112,7 @@ impl fmt::Display for LoadingError {
         match *self {
             WalkDir(ref error) => error.fmt(f),
             Io(ref error) => error.fmt(f),
-            #[cfg(feature = "yaml-load")]
-            ParseSyntax(_) => write!(f, "Invalid syntax file"),
-            ParseTheme(_) => write!(f, "Invalid syntax theme"),
-            ReadSettings(_) => write!(f, "Invalid syntax theme settings"),
-            BadPath => write!(f, "Invalid path"),
+            _ => write!(f, "{}", self.description()),
         }
     }
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -55,6 +55,8 @@ pub mod html;
 mod escape;
 
 use std::io::Error as IoError;
+use std::error::Error;
+use std::fmt;
 #[cfg(all(feature = "yaml-load", feature = "parsing"))]
 use parsing::ParseSyntaxError;
 use highlighting::{ParseThemeError, SettingsError};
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -102,3 +104,45 @@ impl From<ParseSyntaxError> for LoadingError {
...
+
+impl fmt::Display for LoadingError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        use LoadingError::*;
+
+        match *self {
+            WalkDir(ref error) => error.fmt(f),
+            Io(ref error) => error.fmt(f),
+            #[cfg(feature = "yaml-load")]
+            ParseSyntax(_) => write!(f, "Invalid syntax file"),
```
trishume: These cases could be implemented as `_ => write!(f, self.description())` instead of duplicating the messages.

sharkdp: For some reason, I didn't think it would work in this direction (because `Error` depends on `fmt::Display`).. but that's much better, of course. Updated.

sharkdp:
> Maybe some day syntect should switch to failure or something.

Yes, a library that removes some of the boilerplate involved in error handling would certainly be a good idea. `error-chain` seems to be another alternative.


By the way: thank you for this amazing library!

src/lib.rs
```diff
@@ -113,10 +113,10 @@ impl fmt::Display for LoadingError {
             WalkDir(ref error) => error.fmt(f),
             Io(ref error) => error.fmt(f),
             #[cfg(feature = "yaml-load")]
-            ParseSyntax(_) => write!(f, "Invalid syntax file"),
-            ParseTheme(_) => write!(f, "Invalid syntax theme"),
-            ReadSettings(_) => write!(f, "Invalid syntax theme settings"),
-            BadPath => write!(f, "Invalid path"),
+            ParseSyntax(_) => write!(f, "{}", self.description()),
```
trishume: Still not a fan of this duplication. I'd prefer either a `_` pattern or if you really want the exhaustiveness checking (I don't think it's important in this case since description is a fine default) you can use a separate item for `ParseSyntax` and then `ParseTheme(_) | ReadSettings(_) | BadPath =>` for the rest.

src/lib.rs
```diff
@@ -102,3 +104,41 @@ impl From<ParseSyntaxError> for LoadingError {
...
 }
+
+impl fmt::Display for LoadingError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        use LoadingError::*;
+
+        match *self {
+            WalkDir(ref error) => error.fmt(f),
+            Io(ref error) => error.fmt(f),
+            _ => write!(f, "{}", self.description()),
```
messense: I think this should not rely on `Error::description()` since it is [going to be deprecated soon](https://github.com/rust-lang/rust/pull/50163)?

PR #162, opened 2018-05-11
Allow syntest to work with pack dumps generated on Windows

keith-hall:
Sublime Text internally deals with Unix style paths, which is why syntax tests point to the syntax definition using `Packages/PackageName/SyntaxName.sublime-syntax`. Unfortunately, that currently fails in `syntect` with pack dumps generated on Windows, because it stores the paths in Windows format and thus can't find them when it comes time to look it up using Unix path separators. This PR fixes that by converting `\` to `/` when storing the path the syntax was loaded from. I'm not sure if this is the best solution, so feel free to suggest problems with it and/or something better. :)

### DIFF
--- a/src/parsing/syntax_set.rs
+++ b/src/parsing/syntax_set.rs
@@ -101,7 +101,7 @@ impl SyntaxSet {
             if entry.path().extension().map_or(false, |e| e == "sublime-syntax") {
                 let syntax = load_syntax_file(entry.path(), lines_include_newline)?;
                 if let Some(path_str) = entry.path().to_str() {
-                    self.path_syntaxes.push((path_str.to_string(), self.syntaxes.len()));
+                    self.path_syntaxes.push((path_str.replace("\\", "/").to_string(), self.syntaxes.len()));
                 }
                 self.syntaxes.push(syntax);
             }


src/parsing/syntax_set.rs
```diff
@@ -101,7 +101,7 @@ impl SyntaxSet {
             if entry.path().extension().map_or(false, |e| e == "sublime-syntax") {
                 let syntax = load_syntax_file(entry.path(), lines_include_newline)?;
                 if let Some(path_str) = entry.path().to_str() {
-                    self.path_syntaxes.push((path_str.to_string(), self.syntaxes.len()));
+                    self.path_syntaxes.push((path_str.replace("\\", "/").to_string(), self.syntaxes.len()));
```
robinst: Instead of this, you could use the components of `Path` to construct a slash-separated String, something like this: https://play.rust-lang.org/?gist=0ae1aa5f9f1a9c5db2ba41f83bb13922&version=stable&mode=debug

keith-hall: Great idea, thanks - I will take a look at that when I get chance

keith-hall:
it's possible for different syntax definitions (in different folders) to have the same file name and both be usable independently, so it may not be enough to ignore the path and just check the file name. It's not necessarily just used for `syntest`, it's a public API to retrieve a definition by path. In Sublime land, its very common, so if someone was working on something cross compatible somehow they would likely want the current path behavior to stay.

src/parsing/syntax_set.rs
keith-hall: I've pushed a new commit addressing this :)

PR #163, opened 2018-05-13
Find syntax by filename as well as file extension

sharkdp:
Some sublime-syntax files include explicit filenames in their list of supported "extensions", like `.bashrc` or `.gitignore`. This PR adds a feature to find the right syntax definition for these files.

Note: I'm not sure if this is the right place to introduce this feature.


sharkdp:
> I think it would be worth adding a test case for this, to ensure it is not regressed in the future

Yes, good point. I missed the tests at the bottom of the file. Updated.

Issue #164, opened 2018-05-25
Panic due to erroneous regex-rewriting


sharkdp:
Running `syncat` in no-newlines mode fails for the following file:

`test.rd` (R documentation)
``` rd
\usage{
add(x, y)
}
```

```
> cargo run --example syncat -- -n test.rd
[..]
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value:
Error(-114, target of repeat operator is invalid)', libcore/result.rs:945:5
```

This is apparently caused by rewriting the following regex:
```
((\\)(?:usage))(\{)(?:\n)?
```
from [`testdata/Packages/R/Rd (R Documentation).sublime-syntax` line 92](https://github.com/sublimehq/Packages/blob/fa6b8629c95041bf262d4c1dab95c456a0530122/R/Rd%20(R%20Documentation).sublime-syntax#L92) in a way that it includes `(?:$)?` at the end (causing the "target of repeat operator is invalid" error).

I found this while working on a fix for #98 (See #165).

trishume:
Darn. The nonewlines rewriting is always very fraught and causes us a lot of trouble. Not sure how to fix this. One option is to just add a special case for this instead of solving it generally, which I've done before.

sharkdp:
> The nonewlines rewriting is always very fraught and causes us a lot of trouble. Not sure how to fix this. 

I quickly tried to patch this by keeping track of whether we are in a group `(..)` or not, but this attempt immediately failed on the regex above due to the `(\\)` part which is hard to understand without a real parser (the closing parens is *not* escaped). Also, just ignoring the `$`-rewrite in groups is probably not a good idea anyways.

It would be great if we could use the regex engine to parse the regex first and perform the transformation on the AST. However, `onig` doesn't seem to support this(?).

> One option is to just add a special case for this instead of solving it generally, which I've done before.

That sounds like a valid option for now. How would this work? Would the special case be the full regex `((\\)(?:usage))(\{)(?:\n)?` or just the `(?:\n)?` part? There are two regexes in the `Rd` sublime-syntax using this pattern.

trishume:
Probably just the `(?:\n)?` or even just the `(?:\n)` part as a string replace before the rewriter in `rewrite_regex`. With a comment saying it's a hack to fix the `Rd` syntax.

robinst:
> I quickly tried to patch this by keeping track of whether we are in a group (..) or not

[...]

Issue #98, opened 2017-08-31
Replace calls to panic! with Result types


jmacdonald:
After doing a quick search, there are 8 occurrences of `panic!` in syntect. It'd be great to have these errors bubbled up using `Result` types with custom errors, instead. With the existing implementation, it's awkward to handle these, and because these methods don't signal the possibility of these errors, you won't realize there's an issue until you hit a runtime panic and have to work backwards to find the source.

trishume:
Agreed, it'll be unfortunate that parsing will sometimes fail with regex compilation errors, but better than probably better than panicking. I'm unlikely to work on this myself any time soon though.

jmacdonald:
Totally fine. Custom `error-chain` error setup is still fresh in my mind, so I may tackle both this and #94 shortly.

robinst:
From discussion on https://github.com/trishume/syntect/issues/156#issuecomment-386162742:

For checking that regexes compile, we might want to do that when we load the syntaxes from the yaml files (before linking them). For patterns with placeholders, we'd just use a dummy value (`a` or whatever) before compiling them.

This should not slow down normal use cases where loading from yaml happens once, is serialized to a binary dump, and then later that dump is loaded before matching. Loading the dump would not compile the regexes until they are needed, just like the current behavior.

That gives us a much earlier (and more useful) check if there's a problematic regex in a syntax definition. Note that compiling could still fail at runtime if a dump from a previous version of syntect is loaded, and something about the regex engine changed in syntect, but that should be minimal compared to what we have now. And we could still make matching work with Results later.

Enselic:
I plan on looking into this for 5.0. This clippy command can be used to find potential panics:

```cargo clippy -- --deny clippy::panic --deny clippy::unwrap_used --deny clippy::expect_used
```

there are currently 38 of these, of which 5 are direct calls to `panic!()`.

Canop:
Related: I didn't want to use the version 5 now and I had to fix some panics because it crashed broot, so I converted them to errors in [this commit](https://github.com/Canop/syntect/commit/5c9840eab9477c3964b1a310ede4f5e3070ccc67). This commit isn't directly usable as I had to start from an old commit but I don't think it would be hard to do the same on the whole up to date codebase.

Enselic:
@Canop Thanks for sharing! I will make sure to cover all cases of `Result` that you add in that commit before I close this issue.

I will not use your API verbatim however. For example, I'm not a fan of the `CrashError` type, because:
[...]

PR #165, opened 2018-05-25
Check regular expressions for compile errors

sharkdp:
This PR implements a basic version of regex syntax checking as proposed in #98 by @robinst. This would hopefully allow us to catch a lot of possible errors at `SyntaxSet`-generation time (and prevent `panic!`s at syntax-highlighting time).

I have also implemented `Error` for `ParseSyntaxError` (which includes `RegexCompileError`). This allows user code to display nice error messages when loading syntax sets. For example, I get the following error when I try to load [`CMake.sublime-syntax`](https://github.com/zyxar/Sublime-CMakeLists):

```
Error while compiling regex '(?=\b@?[A-Z_][A-Z0-9_]*(?=[^\w-<>=\$]))':
unmatched range specifier in char-class
```
(this has already been addressed upstream: https://github.com/zyxar/Sublime-CMakeLists/pull/32)

Note that the tests currently do not pass due to #164 (I believe this is a good thing!), so #164 should be fixed before merging this. All tests pass if `Rd (R Documentation).sublime-syntax` is removed.

---

The backreference "rewriting" is very immature (and maybe even "unsafe"?). I ended up using this simple search-and-replace after trying two other approaches first:

- Simply ignore regex compile errors with code `-208` (`onig_sys::ONIGERR_INVALID_BACKREF`).
  However, using this approach, we wouldn't catch other syntax errors in these regexes.

- Use `onig`s `SyntaxBehavior` to disable backreference checking (because I thought that
  SYNTAX_BEHAVIOR_STRICT_CHECK_BACKREF might help - it did not).

### DIFF
--- a/src/parsing/yaml_load.rs
+++ b/src/parsing/yaml_load.rs
@@ -247,6 +247,21 @@ impl SyntaxDefinition {
         })
     }
 
+    fn try_compile_regex(regex_str: &str) -> Result<(), ParseSyntaxError> {
+        // Replace backreferences with a dummy placeholder value
+        let mut regex_str = String::from(regex_str);
+        for i in 0..10 {
+            regex_str = regex_str.replace(&format!("\\{}", i), "placeholder");
+        }
+
+        match Regex::new(&regex_str) {
+            Err(onig_error) => {
+                Err(ParseSyntaxError::RegexCompileError(onig_error))
+            },
+            _ => Ok(())
+        }
+    }
+
     fn parse_match_pattern(map: &Hash,
                            state: &mut ParserState)
                            -> Result<MatchPattern, ParseSyntaxError> {
@@ -259,6 +274,8 @@ impl SyntaxDefinition {
             rewrite_regex(regex_str_1)
         };
         // println!("{:?}", regex_str);
+
+        Self::try_compile_regex(&regex_str)?;
 
         let scope = get_key(map, "scope", |x| x.as_str())
             .ok()
@@ -701,6 +718,24 @@ mod tests {
         match def.unwrap_err() {
             ParseSyntaxError::MissingMandatoryKey(key) => assert_eq!(key, "escape"),
             _ => assert!(false, "Got unexpected ParseSyntaxError"),
+        }
+    }
+
+    #[test]
+    fn errors_on_regex_compile_error() {
+        let def = SyntaxDefinition::load_from_str(r#"
+        name: C
+        scope: source.c
+        file_extensions: [test]
+        contexts:
+          main:
+            - match: '[a'
+              scope: keyword.name
+        "#,false, None);
+        assert!(def.is_err());
+        match def.unwrap_err() {
+            ParseSyntaxError::RegexCompileError(_) => {},
+            _ => assert!(false, "Got unexpedted ParseSyntaxError"),
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/parsing/yaml_load.rs
```diff
@@ -704,6 +763,24 @@ mod tests {
...
+        file_extensions: [test]
+        contexts:
+          main:
+            - match: '[a'
+              scope: keyword.name
+        "#,false, None);
+        assert!(def.is_err());
+        match def.unwrap_err() {
+            ParseSyntaxError::RegexCompileError(ref regex, _) => assert_eq!("[a", regex),
+            _ => assert!(false, "Got unexpedted ParseSyntaxError"),
```
trishume: s/unexpedted/unexpected/

src/parsing/yaml_load.rs
```diff
@@ -247,6 +289,21 @@ impl SyntaxDefinition {
...
     }
 
+    fn try_compile_regex(regex_str: &str) -> Result<(), ParseSyntaxError> {
+        // Replace backreferences with a dummy placeholder value
+        let mut regex_str = String::from(regex_str);
+        for i in 0..10 {
+            regex_str = regex_str.replace(&format!("\\{}", i), "placeholder");
+        }
+
+        match Regex::new(&regex_str) {
```
trishume: I think this may need to be a `with_options` with `ONIG_OPTION_CAPTURE_GROUP` like the other compiles in the library. From some Googling it appears this is the standard for textmate-like things and some syntaxes may rely on it.

sharkdp:
> This is technically a breaking API change

Oh, right - due to the `RegexCompileError(onig::Error)` => `RegexCompileError(String, onig::Error)` change.

> but it is to a previously unused error, and searching all of Github brings up no uses of it, so I think it's fine to change without a major version bump.

:+1: Sounds good to me.

src/parsing/yaml_load.rs
sharkdp: Oh, good point. I saw `Regex::new` somewhere and didn't check at the actually relevant places.

src/parsing/yaml_load.rs
```diff
@@ -247,6 +289,24 @@ impl SyntaxDefinition {
         })
     }
 
+    fn try_compile_regex(regex_str: &str) -> Result<(), ParseSyntaxError> {
+        // Replace backreferences with a dummy placeholder value
+        let mut regex_str = String::from(regex_str);
+        for i in 0..10 {
+            regex_str = regex_str.replace(&format!("\\{}", i), "placeholder");
+        }
```
robinst: Can we make the `regex_with_substitutes` function a bit more abstract (or extract another function), so that we can use the same logic here?: https://github.com/trishume/syntect/blob/af810cdb4333106dd93a298982fe945150b5da4f/src/parsing/syntax_definition.rs#L214

trishume:
@sharkdp I realized that @robinst's suggestion was important not just for cleanliness but also to handle the case of `\\1` correctly. I thought there might be hairy lifetime problems so I wanted to take a crack at it myself, there ended up not being any problems, but I'd already figured it out. 

I pushed my changes to the branch, they look okay? @sharkdp @robinst

sharkdp:
Looks great!

sharkdp:
Thank you!

Very much looking forward to the next release :smile:.

PR #166, opened 2018-05-27
Update submodules

sharkdp:
This supersedes #154. 

- All tests should pass now
- The modifications in sublimehq/Packages should now be regex syntax-checked.

PR #167, opened 2018-05-27
Syntest reports

trishume:
This adds syntax tests to CI and makes it easier to check for regressions locally using them!

I was about to spend some time debugging syntax test regressions in #166 but then I realized I still hadn't gotten around to making regressions be caught automatically and be easy to digest.

I switched `syntest` to use `getopts` since `syncat` was already using it, and then added a `--summary` option which only prints failures and failure counts. I then added a `make syntest` command to diff against known issues and fail if there's a difference, and a `make update-known-failures` to easily update the counts on submodule bump PRs and things, or when we fix some.

I don't print the total number of assertions in summaries so that adding new succeeding tests doesn't cause CI failures.

cc @sharkdp

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,6 +34,7 @@ script:
     fi
   - cargo test
   - make assets
+  - make syntest
   - rm -Rf examples
   - cargo doc
   # default features are required for examples to build - so remove them from sight.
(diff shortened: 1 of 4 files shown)


trishume:
@sharkdp As soon as I submitted the CI run I guessed that might be a problem. Walkdir iterates in sorted order only on macOS apparently.

I fixed it by making `syntest` output in sorted order instead, which properly preserves messages at the start and end and also leads to consistent syntest output for other scripts.

sharkdp:
I can rebase my branch in #166 on master after this is merged - to test if CI fails.

PR #170, opened 2018-05-29
Search for filename before file extension

### LINKED ISSUE #169: Unable to find correct syntax for "CMakeLists.txt"
This [CMake.sublime-syntax](https://github.com/zyxar/Sublime-CMakeLists/blob/ff9a800a4ca942edd095de553ca05fba03b02275/CMake.sublime-syntax#L7) file specifies `CMakeLists.txt` as a "file extension", but `syntect` is currently not able to find the correct syntax, because it searches for the actual file extension first (`txt`) and finds the "Plain Text" syntax.

sharkdp:
Some files like `CMakeLists.txt` have a file extension that does not
specify the correct syntax.

This commit changes the order of the syntax lookup such that we
search for the whole filename first and fall back to the file
extension if there were no matches.

closes #169

### DIFF
--- a/src/parsing/syntax_set.rs
+++ b/src/parsing/syntax_set.rs
@@ -209,10 +209,10 @@ impl SyntaxSet {
                                                 path_obj: P)
                                                 -> io::Result<Option<&SyntaxDefinition>> {
         let path: &Path = path_obj.as_ref();
+        let file_name = path.file_name().and_then(|n| n.to_str()).unwrap_or("");
         let extension = path.extension().and_then(|x| x.to_str()).unwrap_or("");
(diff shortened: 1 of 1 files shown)


mk12:
I just noticed I got C syntax highlighting for a file called `c`. While not a super common use case (I was just naming temporary files `a`, `b`, `c`, etc.), I think it would be more correct to have separate maps for filenames and extensions.

keith-hall:
Agreed, but we are dependent on Sublime HQ updating the `sublime-syntax` file format really, see https://github.com/sublimehq/sublime_text/issues/1353#issuecomment-246982762

PR #173, opened 2018-06-01
trim dots from the right of the scope (selector)

keith-hall:
This PR tweaks the way scopes and scope selectors are created from strings, to more closely match how Sublime Text does it. Specifically, Sublime ignores trailing `.` characters.

In Sublime:

    sublime.score_selector('comment.line.cs', 'comment.line...')

returns 16. The same as with no trailing dots.

[This difference was causing a `syntest` failure](https://travis-ci.org/trishume/syntect/jobs/384849350#L676) that doesn't occur in Sublime for Makefile -
 https://github.com/sublimehq/Packages/blob/22c230deb432ee09b22de0b962ee94567e08b866/Makefile/syntax_test_makefile.mak#L251

### DIFF
--- a/src/parsing/scope.rs
+++ b/src/parsing/scope.rs
@@ -136,7 +136,7 @@ impl ScopeRepository {
         if s.is_empty() {
             return Ok(Scope { a: 0, b: 0 });
         }
-        let parts: Vec<usize> = s.split('.').map(|a| self.atom_to_index(a)).collect();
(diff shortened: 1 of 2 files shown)


keith-hall:
interestingly, on the `fancy-regex` branch, I only get the 1 assertion failure for `Makefile` before this PR. I wonder where the others come from.

PR #174, opened 2018-06-02
Remove unused import that caused the warning

vors:
Fix this warning.

```
warning: use of deprecated item 'std::ascii::AsciiExt': use inherent methods instead
  --> src/parsing/syntax_set.rs:19:5
   |
19 | use std::ascii::AsciiExt;
   |     ^^^^^^^^^^^^^^^^^^^^
   |
   = note: #[warn(deprecated)] on by default
```

My understanding is that comment was referring to the stable toolchain. I tested it with the stable rustup toolchain and it compiles fine, so I assume it's safe to remove.

### DIFF
--- a/src/parsing/syntax_set.rs
+++ b/src/parsing/syntax_set.rs
@@ -14,10 +14,6 @@ use std::ops::DerefMut;
 use std::mem;
 use std::rc::Rc;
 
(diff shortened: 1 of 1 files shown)


trishume:
This'll probably force people to upgrade to the latest stable Rust, but that's fine.

Issue #133, opened 2018-01-11
Color Scheme rules applied differently to Sublime


keith-hall:
In Sublime Text, given a simple color scheme like:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
    <key>name</key>
    <string>Test</string>
    <key>settings</key>
    <array>
        <dict>
            <key>settings</key>
            <dict>
                <key>background</key>
                <string>#2b303b</string>
                <key>foreground</key>
                <string>#c0c5ce</string>
            </dict>
        </dict>
        <dict>
            <key>scope</key>
            <string>comment</string>
            <key>settings</key>
            <dict>
                <key>background</key>
                <string>#FD971F</string>
                <key>foreground</key>
                <string>#AAAAAA</string>
            </dict>
        </dict>
        <dict>
            <key>scope</key>
            <string>comment.line</string>
            <key>settings</key>
            <dict>
                <key>foreground</key>
                <string>#FFFCFF</string>
            </dict>
        </dict>
    </array>
</dict>
</plist>
```

and some sample Python text like `# test`, the `background` rule from the `comment` scope selector applies, and the `foreground` rule from the `comment.line` scope selector applies, as can be seen by:

```python
>>> view.style_for_scope('comment.line')
{'source_column': -1, 'source_file': 'Packages/User/test.tmTheme', 'background': '#fd971f', 'foreground': '#fffcff', 'italic': False, 'source_line': -1, 'bold': False}
```

This is because Sublime applies rules from all scope selectors in the color scheme that match the scope stack, in ascending order of their score, so that the rules with a higher score override those with a lower score.
In `syntect`, currently it just uses the rule with the highest matching score, and so the `background` from `comment` isn't applied.

```
<pre style="background-color:#2b303b;">
<span style="color:#fffcff;"># test</span><!-- missing `background-color:#fd971f` -->
</pre>
```

PR #179, opened 2018-06-27
WIP: support for .tmPreferences features (indentation, comment toggling etc)

cmyr:
**this is not ready, the code is bad, do not merge, etc**

I'd been intending to write up a proposal for this but I was feeling motivated on the train yesterday and ended up just hacking together a proof of concept. This includes a new example, `indent_toy`, which you can use like so:

```bash
cargo run --example=indent_toy -- --extra-syntaxes=testdata/Packages --spaces=4 FILE
```

this will print out a bunch of errors, but will then also reindent files in a bunch of languages. Some work better than others; there's lots of stuff missing from the implementation.

The main question: @trishume, do you have any interest in possibly including these features (starting with indentation) in syntect? I think I should be able to keep this stuff pretty isolated (behind a new cfg flag), i.e. this shouldn't have any impact on other consumers of syntect.

If the answer to that is yes, I'll open a tracking issue and figure out what needs to happen to get this mergeable.

### DIFF
--- a/examples/indent_toy.rs
+++ b/examples/indent_toy.rs
@@ -1,0 +1,49 @@
+extern crate syntect;
+extern crate getopts;
+
+use getopts::Options;
+use syntect::easy::IndentFile;
+use syntect::parsing::SyntaxSet;
+
+fn main() -> Result<(), std::io::Error> {
+
+
+    let args: Vec<String> = std::env::args().collect();
+    let mut opts = Options::new();
+    opts.optflag("t", "tabs", "reindent using tabs");
+    opts.optopt("s", "spaces", "reindent using spaces", "NUM_SPACES");
+    opts.reqopt("p", "extra-syntaxes", "SYNTAX_FOLDER", "Additional folder to search for .sublime-syntax files in.");
+
+    let matches = match opts.parse(&args[1..]) {
+        Ok(m) => { m }
+        Err(f) => { panic!(f.to_string()) }
+    };
+
+    if matches.opt_present("tabs") && matches.opt_present("spaces") {
+        panic!("tabs or spaces? (not both)");
+    }
+
+    let tab_text = if matches.opt_present("tabs") {
+        "\t"
+    } else {
+        let num_spaces = matches.opt_str("spaces")
+        .map(|s| s.parse::<usize>().expect("spaces argument must be an integer"))
+        .unwrap_or(4);
+        &"                                            "[..num_spaces]
+    };
+
+    let mut syntax_set = SyntaxSet::new();
+    syntax_set.load_syntaxes(matches.opt_str("extra-syntaxes").unwrap(), false);
+
+    for src in &matches.free[..] {
+        if matches.free.len() > 1 {
+            println!("==> {} <==", src);
+        }
+
+        let mut indenter = IndentFile::new(src, &syntax_set, tab_text)?;
+        for line in indenter {
+            println!("{}", line);
+        }
+    }
+    Ok(())
+}
--- a/src/easy.rs
+++ b/src/easy.rs
@@ -3,8 +3,10 @@
 //! and caching.
 
 use parsing::{ScopeStack, ParseState, SyntaxDefinition, SyntaxSet, ScopeStackOp};
+use parsing::metadata::{IndentationState};
 use highlighting::{Highlighter, HighlightState, HighlightIterator, Theme, Style};
-use std::io::{self, BufReader};
+
+use std::io::{self, BufRead, BufReader, Lines};
 use std::fs::File;
 use std::path::Path;
 // use util::debug_print_ops;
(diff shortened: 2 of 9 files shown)


keith-hall:
My two cents:

I think it would be great to have support for metadata in syntect. I vote for a feature flag for metadata in general, and maybe one more for indentation on top of that, because I know of some use cases where metadata would be useful, but indentation behavior could be (external and) based on something more modern than the legacy TextMate indentation engine/preferences.

Hopefully you are aware that every token could have different metadata (files) apply to them based on their scopes? (This allows things like commenting JavaScript code inside HTML `script` tags etc.) Sublime Text uses the scope of the `\n` character at the end of the line to determine which indentation metadata to use for the line, probably we will want to replicate that behavior for consistency. And the same [logic as for color schemes should apply](https://github.com/trishume/syntect/issues/133), in that some metadata files might define some rules that override other metadata rules with less specific scopes.

One thing to consider is that, were syntect to be used as part of an editor, it would be great to have some idea how automatic indentation handling could be integrated without duplicating work - for example, in Sublime Text, indentation isn't adjusted automatically on a line if you have removed some leading whitespace manually, even when you continue to type something that would normally cause the indentation of the current line to change. I would guess that this sort of logic should be in the text editor itself, to decide when to query syntect for indentation behavior, so that different editors can choose their own rules about when to apply indentation automatically.

I believe that to support metadata properly, finishing support for all types of scope selectors will be necessary - see https://github.com/trishume/syntect/issues/36 for details.

Lastly, it may be useful to define an easy way to test indentation for a file. A friend of mine put together a proof of concept plugin for Sublime Text at https://github.com/STealthy-and-haSTy/SublimeScraps/tree/036a7df124e94d7b0f89e0d4f664db53dd6698d0/indent_test, but it currently doesn't cater for user preferences like `indent_to_bracket` etc. so maybe we can come up with something more robust wrt individual customization.

cmyr:
@keith-hall Thanks for the input, I was definitely going to ping you about this if the process got any further along.

I did intentionally avoid using full scope selectors for the time being, instead just grabbing the top level scope. This was mostly for expediency, in that I didn't want to have to figure out the whole scope matching mechanism for this first pass.

Thanks for pointing out #133; this might explain one of my confusing cases, which is that sublime increases the indent after the opening `(` in the first line of the following rust code, but that this behaviour isn't defined in any rust metadata I could find:

```rust
fn my_multi_arg_function(
    arg1,
    arg2, // etc
```

I guess this might be inherited from a top level `source`-scoped `increaseIndentPattern`?

My interest here _is_ as part of using syntect in an editor, specifically [xi](https://github.com/google/xi-editor). For us all syntax highlighting and language support is provided by plugins, with syntect being bundled to provide broad initial support. As you suggest, I think the responsibility for things like disabling this behaviour on a given line in response to user action falls to the library consumer.

Longer term, ideally language-specific plugins that are performing real syntactic analysis would be providing indentation and related features; what I'd like to get from syntect is a set of reasonable defaults.

re: testing: my initial goal with this work is specifically to provide information about how indentation should be adjusted when a newline is inserted. Support for real whole-file reindentation is definitely trickier, and some kind of testing harness would be helpful if we go that route. I _hope_ that for single-line behaviour, unit tests will work well enough?

keith-hall:
Keeping it simple for now makes sense, I just wanted to clarify whether it was intentional (in that there might be a lot of refactoring work later to get it to work like Sublime does). :+1:  I will take this opportunity to say thanks for putting this proof of concept together and seeking feedback. :) I'd love to see this done right :tm:, so apologies if my expectations seem a little high at the moment, I do realize we have to start somewhere, but I wanted to make sure we're all on the same page ;)

Sublime indenting code after an unclosed open paren is related to the `indentParens` metadata boolean, set in `Packages/Default/Indentation.tmPreferences`. This behavior is a bit odd, ([I covered it in more detail on the ST forums](https://forum.sublimetext.com/t/everything-you-n-ever-wanted-to-know-about-indentation-in-st3/26207),) so I don't know if it is worth replicating in syntect or not. So you are correct about inheritance being the culprit. Putting `<key>indentParens</key><false/>` in your `Rust/Indentation Rules.tmPreferences` file will prevent this behavior in Sublime, in case it helps.

I have no problem with unit tests for single line indentation. I think that where I was mainly coming from was thinking that it may be simpler to just have one "indentation test suite" that will be usable both for a line by line basis (by collecting all the lines that were tested) and for whole file reindentation, in the sense that the whole file approach would have to work line by line anyway to get the final result. (I believe I had the idea that simply specifying the final result of multiple auto-indented lines could be clearer and more concise than lots of individual unit tests.) But I can see that just working with small unit tests for now seems easier.

cmyr:
@keith-hall If you're happy to provide the domain knowledge that would let us Do It Right, then I would be happy to work on the implementation. 😎


I'm confused about what it would take to **actually use scope selectors:** how is this supposed to work? my current best understanding (and ignoring any possible optimizations):
    
- parse metadata files. For each file, extract the scope selector and the settings blob; store all pairs of `(Selector, Settings)` in (say) a `Vec`.

- make all of this metadata available where needed, (for instance) by keeping it in some shared pointer type, and sharing that between all the various `SyntaxDefinitions`.

- At EOL, take the trailing scope (or scope stack?) and match it against the selector of each loaded metadata collection.

- Rank the matches, and then for each setting key we're interested in (e.g. `increaseIndentPattern`, check for the existence of that setting key in each metadata collection, in best-rank order; if it exists, test the pattern, returning on a match (not testing further) and continuing if a match fails.

- Use these resolved match results to determine the indent behaviour.

Does that sound reasonable? The devil is definitely in the details/optimizations, but would this at least produce the correct behaviour?


**reindent whole file** the main reason I've put this out of mind for the time being is that I don't want to think about how to handle syntaxes with significant whitespace.

**testing**: I'm definitely open to suggestions as to what the best approach would be, you've certainly thought about this particular problem more than I have.

trishume:
I'm definitely on board with having code for making use of Sublime's metadata in syntect. I do want syntect to be a one stop shop for dealing with all the stuff in Sublime packages so each text editor doesn't have to do look into the packages themselves. Having as much as possible under an off-by-default feature would be ideal though.

@cmyr re your last comment about a metadata store. That seems like maybe something that could go in some kind of `Settings` object under the SyntaxSet, where when you load whole packages into the set it also loads the accompanying settings.

cmyr:
@trishume okay, great. A non-default feature sounds good, I'll keep noodling away on this as I have time.

cmyr:
@keith-hall @trishume can one of you sanity check my 'how this should work' outline?

trishume:
@cmyr Broad approach looks good, with the `SyntaxSet` being the place you should store the object that keeps track of all the settings.

I'm not sure if that's the right semantics for the indent patterns though. I suspect that unlike #133 settings might actually just be based on the top match at the settings-file level, because for other settings it might only make sense to have one canonical value and then the settings system could work with one value uniformly.

You can check which one is correct by creating two indent pattern settings files in your User package in Sublime and checking if the behaviour is consistent with only the matchiest file-selector's indent pattern being applied, or if it is consistent with the behaviour you describe.

cmyr:
Okay, so we only test against the best existing match. That sounds better already. Will update if I get particularly confused.

keith-hall:
btw, one thing that may not be documented anywhere is that the patterns are implicitly anchored at BOL (`^`).

so, `<key>increaseIndentPattern</key><string>hello</string>` will not match a line like `oh hello` in Sublime Text, or if there is any whitespace before `hello` etc.

PR #181, opened 2018-06-28
Upgrade plist to 0.3 and take advantage of serde integration

robinst:
I noticed that plist comes with serde integration now, so we can just directly deserialize into the serde_json Value types using the magic of serde :).

Output of `cargo update`:

```
    Updating registry `https://github.com/rust-lang/crates.io-index`
    Updating base64 v0.8.0 -> v0.9.2
      Adding humantime v1.1.1
    Updating plist v0.2.4 -> v0.3.0
```

### DIFF
--- a/src/highlighting/settings.rs
+++ b/src/highlighting/settings.rs
@@ -2,8 +2,8 @@
 /// released under the MIT license by @defuz
 
 use std::io::{Read, Seek};
-use plist::{Plist, Error as PlistError};
-use serde_json::Number;
+use plist::{Error as PlistError};
+use plist::serde::deserialize;
 
 pub use serde_json::Value as Settings;
 pub use serde_json::Value::Array as SettingsArray;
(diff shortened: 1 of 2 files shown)


src/highlighting/settings.rs
```diff
@@ -33,22 +33,6 @@ impl From<PlistError> for SettingsError {
...
-            SettingsObject(entries.into_iter().map(|(k, v)| (k, to_json(v))).collect()),
-        Plist::Boolean(value) => Settings::Bool(value),
-        Plist::Data(bytes) => Settings::Array(bytes.into_iter().map(|b| b.into()).collect()),
-        Plist::Date(value) => Settings::String(value.to_string()),
-        Plist::Real(value) =>
-            Settings::Number(Number::from_f64(value).expect("Error converting plist real value to JSON number")),
-        Plist::Integer(value) => Settings::Number(value.into()),
-        Plist::String(s) => Settings::String(s),
-    }
+    let settings = deserialize(reader)?;
```
cmyr: or just:

```rust
Ok(deserialize(reader)?)
```

(okay either way obviously, just riffing)

robinst: Yeah, had that before, but then decided that having an extra variable reads a bit nicer because it removes the `)?)` 🤷‍♂️.

PR #182, opened 2018-06-28
Use arena for contexts

robinst:
This still needs more work and there's tons of TODOs to clean up, but I have some questions which will affect the direction.

The high-level overview of changes so far:

* `contexts` on `SyntaxDefinition` changed from a `HashMap<String, ContextPtr>` to a `Vec<Context>`
* `Context` has a new `name` field, and the `contexts` are sorted by that. The idea is that during linking, we can do a binary search to find contexts. After linking, we have a `usize` index to look up the context quickly.
* Nested contexts (e.g. anonymous contexts that are pushed) are no longer directly included in the parent context, but instead they are pulled out into the flat context list. This means the contexts are no longer a tree, which simplified satisfying the borrow checker a bit. Having said that, I'm not sure it's strictly needed, we can revisit that.

High-level TODOs:

* [x] Make cloning of `SyntaxDefinition` safe or prevent it
* [x] Add regex cache: see #186
* [x] Fix `FirstLineCache`
* [x] See if `ParseState` can be cached (try out in xi-editor [here](https://github.com/google/xi-editor/blob/master/rust/syntect-plugin/src/main.rs))
* [x] Name anonymous contexts the same way as Sublime
* [x] Look at storing contexts as a single `Vec<Context>` in `SyntaxSet`: see #187 
* [x] Decide if `ParseState` should reference `SyntaxSet` and `Context`s directly or just keep `ContextId`s
* [x] Update documentation
* [x] Change `SyntaxSetBuilder` API as recommended by guidelines: https://rust-lang-nursery.github.io/api-guidelines/type-safety.html#consuming-builders (also go over method names) -> some methods return `Result`, so can't make chaining work for all methods. Probably best to leave as is.


src/parsing/yaml_load.rs
```diff
@@ -412,19 +465,24 @@ impl SyntaxDefinition {
...
         // check for a push of multiple items
         if y.as_vec().map_or(false, |v| !v.is_empty() && (v[0].as_str().is_some() || (v[0].as_vec().is_some() && v[0].as_vec().unwrap()[0].as_hash().is_some()))) {
             // this works because Result implements FromIterator to handle the errors
             y.as_vec()
                 .unwrap()
                 .iter()
-                .map(|x| SyntaxDefinition::parse_reference(x, state))
+                .enumerate()
+                .map(|(i, x)| SyntaxDefinition::parse_reference(
+                    &format!("{}.{}", name, i), x, state, contexts))
```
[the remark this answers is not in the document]
keith-hall: it used to be possible to find how ST names these contexts by checking the ST console when an anonymous context included a `meta_scope` and was subject to a `with_prototype` - see https://github.com/SublimeTextIssues/Core/issues/1683. This is the closest I have to a reference or a way to check the names (with ST build 3153 or older).

It seems like the naming convention is `#anon_` followed by the named context, followed by `_` and the index of the anonymous context within that named context (regardless of nesting level, in an XML-like "document order").

This was my test syntax definition:

```yaml
%YAML 1.2
---
scope: source.example-anon
contexts:
  main:
    - include: c
  c:
    - meta_scope: test_c
    - match: c_test
      scope: c_test
  d:
    - meta_scope: test_d
    - match: d_example
      scope: d_example
      push: #anonymous context 0 inside `d` context
        - meta_scope: test_e
        - match: e
          scope: e
          push: # anonymous context 1 inside `d` context
            - meta_scope: testing123
            - match: f
              push: # anonymous context 2 inside `d` context
                - match: g
              with_prototype:
                - match: stop_already...
          with_prototype:
            - match: nope
      with_prototype:
        - match: this
    - match: another
      push: # anonymous context 3 inside `d` context
        - meta_scope: blah
        - match: something
          push: # anonymous context 4 inside `d` context
            - meta_scope: foobar
            - match: hello
          with_prototype:
            - match: me

```

and the ST console output:

>
```
generating syntax summary
rule 719396458c7046797910f3c1990b81d6::#anon_d_1 has a scope name, but is unreachable, so the name will never be used
rule c has a scope name, but is unreachable, so the name will never be used
rule d has a scope name, but is unreachable, so the name will never be used
rule #anon_d_0 has a scope name, but is unreachable, so the name will never be used
rule #anon_d_1 has a scope name, but is unreachable, so the name will never be used
rule #anon_d_4 has a scope name, but is unreachable, so the name will never be used
```

I believe that `719396458c7046797910f3c1990b81d6` can be ignored for our purposes, I think it's some kind of unique identifier ST gave when creating new anonymous contexts with the `with_prototype` applied, which is not something we do in syntect.

src/parsing/syntax_set.rs
```diff
@@ -248,122 +294,147 @@ impl SyntaxSet {
...
         // 2 loops necessary to satisfy borrow checker :-(
-        for syntax in &mut self.syntaxes {
-            if let Some(proto_ptr) = syntax.contexts.get("prototype") {
-                Self::recursively_mark_no_prototype(syntax, proto_ptr.clone());
-                syntax.prototype = Some((*proto_ptr).clone());
+        for (syntax_index, syntax) in self.syntaxes.iter_mut().enumerate() {
+            if let Some(context_index) = syntax.find_context_index_by_name("prototype") {
+//                let prototype = &mut syntax.contexts[context_index];
+                // TODO: this works but is it nice? Would it be nicer to make
+                // recursively_mark_no_prototype a method on SyntaxDefinition?
```
trishume: This makes me wonder if this is different than how Sublime does it. Now I'm wondering if in Sublime I include a named context in the prototype, and then push that named context from the main section, will it use the prototype or not. This also makes linking not idempotent since if a prototype includes something from another syntax (don't think this ever happens in practice) the first time it'll work fine, but on second link it can interfere with the prototype of another syntax.

I think Sublime may use a model of duplicating contexts when they have a prototype. Although I'm not sure if this ever comes up in practice.

Anyhow yah maybe this would be better on `SyntaxDefinition`, although I think ideally all contexts get moved into a single `Vec` in the `SyntaxSet` and then there's some kind of recursive context walking helper method that this can use or something.

src/parsing/yaml_load.rs
```diff
@@ -128,6 +126,21 @@ impl SyntaxDefinition {
         SyntaxDefinition::parse_top_level(doc, scope_repo.deref_mut(), lines_include_newline, fallback_name)
     }
 
+    // TODO: visibility. also, is this the right file to have these?
```
trishume: No, note that this whole module isn't included if the yaml loading feature isn't included, and I think you use this in a test outside this module.

src/parsing/syntax_set.rs
```diff
@@ -57,6 +58,34 @@ impl Default for SyntaxSet {
     }
 }
 
+struct ContextMap {
```
trishume: I feel like there must be a better way of doing linking, although it may require restructuring some data.

Actually I just realize I don't even understand why this is a thing, why not just iterate over the list of syntaxes directly to find the right index instead of constructing all of these extra vectors with the same indexes and information?

src/parsing/syntax_set.rs
robinst: > Now I'm wondering if in Sublime I include a named context in the prototype, and then push that named context from the main section, will it use the prototype or not.

I tested that as part of #180 and the answer is: No, it also won't use the prototype in that case. So the way we currently do it seems to be correct.

src/parsing/syntax_set.rs
robinst: Because at the point we want to find a syntax/context, we're within a method that has a mutable borrow of a `Context` or some component of it.

I tried what you suggested but I couldn't get it past the borrow checker. If you have any ideas that would be good. We could probably do something with unsafe, because all we want to do is mutate `ContextReference`, not changing any of the structure apart from that.

src/parsing/parser.rs
```diff
@@ -24,44 +24,42 @@ use fnv::FnvHasher;
 ///
 /// **Note:** Caching is for advanced users who have tons of time to maximize performance or want to do so eventually.
 /// It is not recommended that you try caching the first time you implement highlighting.
-#[derive(Debug, Clone, PartialEq, Eq)]
-pub struct ParseState {
-    stack: Vec<StateLevel>,
+// TODO: removed PartialEq, Eq from this because SyntaxSet is not
+#[derive(Debug, Clone)]
+pub struct ParseState<'a> {
+    syntax_set: &'a SyntaxSet,
```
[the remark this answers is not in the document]
robinst: I'll try out the changes in xi-editor.

Passing it in `parse_line` seems like you could accidentally pass in a different one and get all kinds of messed up results.

The way it's currently written with the lifetime specifier reflects the reality nicely: The `ParseState` is valid for as long as the corresponding `SyntaxSet` exists. If that goes away or is changed, a `ParseState` instance should no longer be used.

src/parsing/syntax_set.rs
trishume: Ah okay makes sense.

One way around this might be to switch to having a flat `Vec<Context>` with `SyntaxDefinition`s just having references into it. The YAML loader would take a mutable ref to a `SyntaxSet` (or the underlying `Vec`) that it would shove new contexts into. Then `SyntaxDefinition` would have something like a `HashMap<String, ContextId>` to represent the names. 

That would accomplish moving to a flat system (which I suspect will work better/faster in other ways, see other comments) and also separate the lookup data (and move back to an efficient map lookup for names).

It would sacrifice being able to pick up syntaxes and drop them in different sets, but I don't think that's that useful. The one test that does that can just ask the SyntaxSet to load the file it wants directly instead of loading the entire directory and picking one out.

robinst: This is an interesting idea. I think it's a bigger change than what I have here, but I'd like to try it out.

> It would sacrifice being able to pick up syntaxes and drop them in different sets

Depending on how we implement it, I think we could still make that possible. We could retrieve the original contexts based on the `HashMap<String, ContextId>`.

src/parsing/syntax_definition.rs
```diff
@@ -81,31 +92,26 @@ pub enum Pattern {
...
-#[derive(Debug, Eq, PartialEq, Serialize, Deserialize)]
+#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
 pub struct MatchPattern {
     pub has_captures: bool,
     pub regex_str: String,
-    #[serde(skip_serializing, skip_deserializing)]
-    pub regex: Option<Regex>,
+    // #[serde(skip_serializing, skip_deserializing)]
+    // TODO: cache these somewhere?
+    // pub regex: Option<Regex>,
```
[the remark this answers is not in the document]
robinst: I looked at this and noticed that onig's Regex is actually `Send` and `Sync` now: https://github.com/rust-onig/rust-onig/pull/47

So it looks like all we need is lazy initialization, not a thread-local cache. I've raised a PR here: https://github.com/trishume/syntect/pull/186

src/parsing/syntax_set.rs
robinst: Ok, so I've:

* Made the change to introduce a `SyntaxSetBuilder` on this branch
* Created #187 for the flat `Vec<Context>` approach on top of this branch

robinst:
@trishume I've ported the syntect-plugin, see here: https://github.com/google/xi-editor/compare/master...robinst:syntect-plugin-arena

The lifetimes are not that nice, but it works. I had to re-add the implementation of `PartialEq` and `Eq` for `ParseState` in this PR.

The plugin using `==` to compare `ParseState`s might be a good reason to switch it to use `ContextId` instead of keeping `&Context` and `&SyntaxSet`, as you mentioned earlier. It would mean old contexts would have to be looked up via the id every time they're used, but not sure how much slower that would be. It's your call, I don't mind changing it.

I'll look into the other remaining things for now :).

trishume:
Sorry for leaving this to sit longer than usual, I've been really busy lately with [a school project](http://thume.ca/ray-tracer-site/). I'll try and take a look at what the plugin rewrite looks like this weekend, but if you have time to test switching to ContextID and the benchmarks don't regress much, as I suspect they may not due to things staying in cache, then that may be a good idea.

robinst:
No worries! I've now finished all the big outstanding TODOs. Now the only thing that's left is documentation and some organizational things such as method visibility/naming and where to put them.

I've also added some tests for `SyntaxSet` and the builder for adding syntaxes, cloning, and multi-threaded usage :).

The change for `ParseState` is in a separate PR for now: #190

robinst:
So I've merged the other PR into this one, and updated the docs and the README.

The only thing left that I'd like to look at is:

1. I've kept the `ContextId` APIs `pub(crate)` for now. Do you think e.g. the `contexts` field of `SyntaxReference` should be public, and `get_context` as well?
2. While we're breaking the API, do you want to clean up the methods in `SyntaxSetBuilder` a bit? I'm thinking `load_syntaxes` should be renamed to `load_from_folder` maybe. I'm also not sure we need `load_from_folder` on `SyntaxSet` anymore. And maybe we should rename `add_syntax` to just `add`, or rename `load_from_folder` to `load_syntaxes_from_folder` to make the names consistent.

robinst:
Actually let's do the API renames in a new PR, makes it possible to review.

Should I click the Merge button? :)

trishume:
Nice, a separate PR for renames sounds good.

I normally merge but you should indeed have the satisfaction of hitting merge on this enormous amount of awesome work, go ahead!

After the renames I'll look at drafting a 3.0 release with notes on the breaking and other changes.

robinst:
Merged! 🎉

I'll prepare the other PR in the next few days.

Issue #175, opened 2018-06-04
`syntest` and assertions that extend over the end of the line


keith-hall:
While looking at the CI failures from https://github.com/trishume/syntect/pull/166, I identified the reason why [`syntest` reports failures for the `Git Rebase` syntax](https://travis-ci.org/trishume/syntect/jobs/387417448#L828):

> `+FAILED testdata/Packages/Git Formats/syntax_test_git_rebase: 3`

[The syntax test file contains assertions that extend beyond the `\n` character](https://github.com/sublimehq/Packages/blob/182966ad7bd71ff8a61327a0b114e3759fe615d6/Git%20Formats/syntax_test_git_rebase#L6). `syntest` compares these assertion scope selectors against the scope stack at the new line character. But in Sublime Text, the following syntax test passes:

```
# SYNTAX TEST "Git Rebase.sublime-syntax"
# <- text.git.rebase comment.line punctuation.definition.comment
pick d284bb2 Initial commit
# <- meta.commit
#^^^^^^^^^^^^^^^^^^^^^^^^^^^ meta.commit
#                           ^ text.git.rebase comment.line.git.rebase punctuation.definition.comment.git.rebase -meta.commit
#                            ^ comment.line.git - punctuation
```

which means, in Sublime Text, the assertions wrap onto the next line. Probably we should adapt `syntest` to also work this way.

keith-hall:
I had a go at this, with keeping the "read and process one line at a time" behavior (as opposed to Sublime's "read the whole syntax test file into memory first" implementation) but my Rust skills aren't good enough yet - I'm fighting the borrow checker it seems. My idea was to queue up any test assertions that belong to a line we haven't read yet. Here's the diff in case it helps someone else patch it up. I'm creating the queue with an initial capacity of 2 - I think no syntax test files reference more than one line below at the moment.

```diff
--- syntect/master/examples/syntest.rs
+++ syntect/syntest/examples/syntest.rs
@@ -25,6 +25,7 @@
 use std::cmp::{min, max};
 use std::time::Instant;
 use std::str::FromStr;
+use std::collections::VecDeque;
 
 use getopts::Options;
 use regex::Regex;
@@ -64,6 +65,7 @@
 
 #[derive(Debug)]
 struct AssertionRange<'a> {
+    line_number: usize,
     begin_char: usize,
     end_char: usize,
     scope_selector_text: &'a str,
@@ -84,7 +86,7 @@
     success: bool,
 }
 
[...]

PR #185, opened 2018-07-02
initial rework of syntest to be usable internally

keith-hall:
this is an initial rework of the `syntest` example, to make it usable internally, as it would be useful for parser tests (see https://github.com/trishume/syntect/pull/180#discussion_r198685953). I think it will need some more work to properly support assertions that extend over the end of the line (https://github.com/trishume/syntect/issues/175) and spill over onto not only the next line, but also the line afterwards etc. although such a scenario may never come up in real life use.

### DIFF
--- a/examples/syntest.rs
+++ b/examples/syntest.rs
@@ -15,16 +15,14 @@ extern crate regex;
 extern crate getopts;
 
 //extern crate onig;
-use syntect::parsing::{SyntaxSet, ParseState, ScopeStack, Scope};
-use syntect::highlighting::ScopeSelectors;
-use syntect::easy::{ScopeRegionIterator};
+use syntect::parsing::{SyntaxSet};
+use syntect::easy::{SyntaxTestFileResult, SyntaxTestOutputOptions, process_syntax_test_assertions};
 
 use std::path::Path;
+use std::io::prelude::*;
 use std::io::{BufRead, BufReader};
 use std::fs::File;
-use std::cmp::{min, max};
 use std::time::Instant;
-use std::str::FromStr;
 
 use getopts::Options;
 use regex::Regex;
@@ -36,12 +34,6 @@ pub enum SyntaxTestHeaderError {
     SyntaxDefinitionNotFound,
 }
 
-#[derive(Debug, Clone, PartialEq, Eq)]
-pub enum SyntaxTestFileResult {
-    FailedAssertions(usize, usize),
-    Success(usize),
-}
-
 lazy_static! {
     pub static ref SYNTAX_TEST_HEADER_PATTERN: Regex = Regex::new(r#"(?xm)
             ^(?P<testtoken_start>\s*\S+)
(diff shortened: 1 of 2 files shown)


examples/syntest.rs
```diff
@@ -15,16 +15,14 @@ extern crate regex;
 extern crate getopts;
 
 //extern crate onig;
-use syntect::parsing::{SyntaxSet, ParseState, ScopeStack, Scope};
-use syntect::highlighting::ScopeSelectors;
-use syntect::easy::{ScopeRegionIterator};
+use syntect::parsing::{SyntaxSet};
+use syntect::easy::{SyntaxTestFileResult, SyntaxTestOutputOptions, process_syntax_test_assertions};
```
trishume: I don't think this should go in the `easy` module. 

That's for simple wrappers for common use cases. Syntax tests are neither a wrapper around a more advanced API, nor a common use case.

They should probably go in their own `syntax_tests` module or something

src/easy.rs
```diff
@@ -174,6 +174,236 @@ impl<'a> Iterator for ScopeRegionIterator<'a> {
...
+    pub summary: bool,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum SyntaxTestFileResult {
+    FailedAssertions(usize, usize),
+    Success(usize),
+}
+
+pub fn process_syntax_test_assertions(syntax: &SyntaxDefinition, text: &str, testtoken_start: &str, testtoken_end: Option<&str>, out_opts: &SyntaxTestOutputOptions) -> SyntaxTestFileResult {
```
trishume: If this is now going to be a public API, could you add a doc comment with short descriptions of what the parameters do? Also useful for our own reference.

src/easy.rs
```diff
@@ -174,6 +174,236 @@ impl<'a> Iterator for ScopeRegionIterator<'a> {
     }
 }
 
+#[derive(Clone, Copy)]
+pub struct SyntaxTestOutputOptions {
```
trishume: Now that this is an API, it seems a bit weird for it to just print things to stdout.

Maybe you could add a lifetime parameter to this struct and add a `output: &'a io::Write` member. Alternatively you could add an extra parameter for the output to the process function.

Then the function then uses `writeln!` instead of `println!` to write to, and the process function would probably need to return an `io::Result`. Then `syntest` and possibly also unit tests could just pass stdout.

keith-hall:
I have moved it to a new `syntax_tests` module and added a doc comment :+1:
The main function *has* been modified; it used to read line by line, parsing as it went - now it reads all the syntax test assertion positions/details first, then parses line by line afterwards as a separate step and compares the results.
In terms of using `io::Write`, great idea. I think it's a little beyond my skillset and free time at the moment, so it'd be great if someone else wouldn't mind contributing this. It will be necessary to also modify `Util::debug_print_ops` - perhaps it'd make sense to rename it `debug_write_ops`, take an `io::Write` as a parameter, and create a new `debug_print_ops` which will just pass in `stdout()` (or https://doc.rust-lang.org/std/io/struct.Stdout.html#method.lock (I'm not sure which is better)).

robinst:
Nice, thanks for picking this up :)!

In the current form, I don't think we should make it a public API. An API with the only output being some printed text doesn't seem very useful generally.

So I would do either one of these two things:

* Change it to `pub(crate)`
* Move the printing parts back to the syntest example, and change the API to return plain structs with the information about the assertions instead.

(With the latter, we could do some nice things such as formatting as HTML so you can more easily see the actual scopes, show the results in a text editor instead, etc.)

keith-hall:
Printed text isn't the *only* output - the API returns the successful/failed assertion count too ;)

I would guess that changing the API to return structs with the information about the assertions, and printing them from the `syntest` example would mean that we would lose the "instant" nature of the results. i.e. currently, one would see in the output about failures at the top of the syntax test file without having to wait for the whole rest of the file to be evaluated. I'd be keen to keep this, if possible.
I like the idea of formatting the output, although I don't have any concrete ideas of how it could look at the moment. (Currently I just pipe the output of `syntest` into a file and review it in Sublime.) Maybe outputting to HTML (like `synhtml`), using a specified color scheme, and then showing the failures on top of that somehow?

robinst:
Hehe :). Is it possible to make it return an `Iterator` with results? Then you could just iterate over that and print as the assertions come back.

trishume:
Okay I'll try and give the new implementation a real review then, likely this weekend. If you'd like I can probably also do the `io::Write` stuff after I do that.

@robinst I agree it's not the greatest thing to have as a public API but:

- there's unfortunately no `pub(examples)`
- making it be a nice proper API may be a decent amount of work.
- I've never been too stingy with public APIs in syntect and in this case if we implement a better API it should be pretty easy to reimplement it in terms of the new API, or do a major release and remove the old version (I doubt anyone will use it)

trishume:
Sorry I'm so late on reviewing this, I wanted to use my limited syntect bandwidth for stuff that needed to be in the 3.0 release. This is still on my TODO list.

keith-hall:
No worries, the work you did on the color schemes was awesome and indeed more important ;)

Enselic:
@keith-hall Is this PR something we still want to try to get in?

In general I think it is undesired to add public API only for the purpose of testing, because each increase in API surface makes all subsequent bugfixes and features more complicated to do due to the mere fact that there are more things to take into account, so in general a minimal API surface is desirable.

Couldn't we simply keep the tests as mod tests that have access to internal interfaces?

I apologize in advance for any ignorance on my part.

Enselic:
Thanks for the clarification, that makes sense. I will try to find the time to help review this in a not too distant future.

PR #186, opened 2018-07-03
Cache compiled regexes with lazycell

robinst:
With that, `MatchPattern` does not have to be mutably borrowed to do matching anymore.

It removes one obstacle for allowing `SyntaxSet` to be used by multiple threads, while still keeping regex compilation lazy.

This is one of the prerequisites for #182. I've created the PR against master so that we can benchmark the impact more directly. But I can integrate it into #182 if you prefer to not put it into master just yet.

### DIFF
--- a/src/parsing/parser.rs
+++ b/src/parsing/parser.rs
@@ -326,8 +326,8 @@ impl ParseState {
 
         for (from_with_proto, ctx, captures) in context_chain {
             for (pat_context_ptr, pat_index) in context_iter(ctx) {
-                let mut pat_context = pat_context_ptr.borrow_mut();
-                let match_pat = pat_context.match_at_mut(pat_index);
+                let pat_context = pat_context_ptr.borrow();
+                let match_pat = pat_context.match_at(pat_index);
 
                 if let Some(match_region) = self.search(
                     line, start, match_pat, captures, search_cache, regions
@@ -374,7 +374,7 @@ impl ParseState {
     fn search(&self,
               line: &str,
               start: usize,
-              match_pat: &mut MatchPattern,
+              match_pat: &MatchPattern,
               captures: Option<&(Region, String)>,
               search_cache: &mut SearchCache,
               regions: &mut Region)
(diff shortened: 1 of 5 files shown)


robinst:
I've run `cargo bench` against master and then against this branch. Did the whole thing twice and these are the results: https://gist.github.com/robinst/ec776e1b9d642c04467d8fd939bf60d4

As you can see, the results are not very reliable.

src/parsing/parser.rs
```diff
@@ -396,24 +396,30 @@ impl ParseState {
...
+                line.len(),
+                SearchOptions::SEARCH_OPTION_NONE,
+                Some(regions),
+                MatchParam::default(),
+            );
+            (matched, false)
         } else {
-            None
+            let regex = match_pat.regex();
+            let matched = regex.search_with_param(
```
trishume: You can return a `Cow` out of the if statement to avoid the duplication of the search.

robinst: Hm, doesn't work because `Regex` is not `Clone`:

```
error[E0277]: the trait bound `onig::Regex: std::clone::Clone` is not satisfied
   --> src/parsing/parser.rs:405:14
    |
405 |             (Cow::Borrowed(match_pat.regex()), true)
    |              ^^^^^^^^^^^^^ the trait `std::clone::Clone` is not implemented for `onig::Regex`
    |
    = note: required because of the requirements on the impl of `std::borrow::ToOwned` for `onig::Regex`
    = note: required by `std::borrow::Cow::Borrowed`
```

The following works, but not sure if it's worth it?:

```rust
        let do_search = |regex: &Regex, regions: &mut Region| {
            regex.search_with_param(
                line,
                start,
                line.len(),
                SearchOptions::SEARCH_OPTION_NONE,
                Some(regions),
                MatchParam::default(),
            )
        };

        let (matched, can_cache) = if match_pat.has_captures && captures.is_some() {
            let &(ref region, ref s) = captures.unwrap();
            let regex = match_pat.regex_with_refs(region, s);
            (do_search(&regex, regions), false)
        } else {
            (do_search(match_pat.regex(), regions), true)
        };
```

trishume: Oh darn. Apparently what we need is a ✨ *[supercow](https://docs.rs/supercow/0.1.0/supercow/)* ✨ but it would be kind of dumb to add that library just for this instance.

That closure solution looks fine to me although I'm also fine with merging as is, don't have a strong preference between them, up to you.

robinst: Lol 😆. Ok, I'll merge it like this.

PR #187, opened 2018-07-03
Use a single Vec<Context> in SyntaxSet after linking

robinst:
This means resolving a `ContextId` is now a single lookup.

Having the `SyntaxSetBuilder` (introduced in a previous commit) helped make this change.

This is a separate PR to hopefully make it easier to review, but it's part of the larger work in #182.

### DIFF
--- a/examples/syncat.rs
+++ b/examples/syncat.rs
@@ -48,7 +48,7 @@ fn main() {
     };
 
     let no_newlines = matches.opt_present("no-newlines");
-    let mut ss = if matches.opt_present("no-default-syntaxes") {
+    let ss = if matches.opt_present("no-default-syntaxes") {
         SyntaxSet::new()
     } else if no_newlines {
         SyntaxSet::load_defaults_nonewlines()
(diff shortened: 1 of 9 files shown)


examples/syncat.rs
```diff
@@ -57,9 +57,10 @@ fn main() {
     };
 
     if let Some(folder) = matches.opt_str("extra-syntaxes") {
-        let mut builder = ss.builder();
-        builder.load_syntaxes(folder, !no_newlines).unwrap();
-        ss = builder.build();
+        // TODO: no way to go back to builder anymore :/
+        //let mut builder = ss.builder();
```
robinst: I think we can bring this back by recreating all the `SyntaxDefinition`s from the `SyntaxReference`s and their contexts, but haven't looked into this yet.

trishume: Hmmm that may be an issue, as I mentioned in the other PR, it would be nice for editors to be able to load new syntaxes on the fly. Mostly thinking of Xi.

Three options:

- Just make editors reload everything on adding a new syntax, which may be fine because they might want to create their own dumps for fast startup, and reloading shouldn't be that slow.
- Maybe have the builder also have context/references `Vec`s that are only used when creating from an existing set. This might be pretty easy.
- What you said, recreate the original structures.

src/parsing/syntax_set.rs
```diff
@@ -311,62 +293,83 @@ impl SyntaxSetBuilder {
...
+            if let Some(prototype_id) = prototype {
+                // TODO: We could do this after parsing YAML, instead of here?
+                Self::recursively_mark_no_prototype(syntax, prototype_id.index(), &contexts, &mut no_prototype);
             }
-            context_names.push(contexts);
-        }
 
-        let context_map = ContextMap::new(scopes, syntax_names, context_names);
-
-        // 2 loops necessary to satisfy borrow checker :-(
```
trishume: `:-)`

robinst:
Rebased both branches and incorporated this into the other PR.

PR #190, opened 2018-07-27
Store ContextId in ParseState instead of &Context

robinst:
This means ParseState doesn't contain a reference to SyntaxSet anymore
and doesn't need a lifetime.

This makes it simpler for code that wants to cache it, but it also means
parse_line needs to be called with the SyntaxSet as an argument so that
it can look up contexts.

### DIFF
--- a/src/parsing/parser.rs
+++ b/src/parsing/parser.rs
@@ -24,37 +24,21 @@ use parsing::syntax_set::{SyntaxSet, SyntaxReference};
 ///
 /// **Note:** Caching is for advanced users who have tons of time to maximize performance or want to do so eventually.
 /// It is not recommended that you try caching the first time you implement highlighting.
-// TODO: removed PartialEq, Eq from this because SyntaxSet is not
-#[derive(Debug, Clone)]
-pub struct ParseState<'a> {
-    syntax_set: &'a SyntaxSet,
-    stack: Vec<StateLevel<'a>>,
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct ParseState {
+    stack: Vec<StateLevel>,
     first_line: bool,
     // See issue #101. Contains indices of frames pushed by `with_prototype`s.
     // Doesn't look at `with_prototype`s below top of stack.
     proto_starts: Vec<usize>,
 }
 
-#[derive(Debug, Clone)]
-struct StateLevel<'a> {
-    context: &'a Context,
-    prototype: Option<&'a Context>,
+#[derive(Debug, Clone, Eq, PartialEq)]
+struct StateLevel {
+    context: ContextId,
+    prototype: Option<ContextId>,
     captures: Option<(Region, String)>,
 }
-
-impl<'a> PartialEq for StateLevel<'a> {
-    fn eq(&self, other: &StateLevel) -> bool {
-        self.context == other.context &&
-        match (&self.prototype, &other.prototype) {
-            (&Some(ref a), &Some(ref b)) => a == b,
-            (&None, &None) => true,
-            _ => false,
-        } &&
-        self.captures == other.captures
-    }
-}
-
-impl<'a> Eq for StateLevel<'a> {}
 
 #[derive(Debug)]
 struct RegexMatch<'a> {
@@ -160,17 +144,16 @@ type SearchCache = HashMap<*const MatchPattern, Option<Region>, BuildHasherDefau
 // So in our input string, we'd skip one character and try to match the rules
 // again. This time, the "\w+" wins because it comes first.
 
-impl<'a> ParseState<'a> {
+impl ParseState {
     /// Create a state from a syntax, keeps its own reference counted
     /// pointer to the main context of the syntax.
-    pub fn new(syntax_set: &'a SyntaxSet, syntax: &'a SyntaxReference) -> ParseState<'a> {
+    pub fn new(syntax: &SyntaxReference) -> ParseState {
         let start_state = StateLevel {
-            context: syntax_set.get_context(&syntax.contexts["__start"]),
+            context: syntax.contexts["__start"].clone(),
             prototype: None,
             captures: None,
         };
         ParseState {
-            syntax_set,
             stack: vec![start_state],
             first_line: true,
             proto_starts: Vec::new(),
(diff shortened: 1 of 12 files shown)


robinst:
@trishume Separate PR to make it easier to review. Feel free to merge into other PR if you decide to go ahead with this :). I'm now also in favor of doing this.

robinst:
(Haven't had a chance to benchmark yet, ran into an issue with criterion: https://github.com/japaric/criterion.rs/issues/182)

src/parsing/parser.rs
```diff
@@ -185,17 +168,23 @@ impl<'a> ParseState<'a> {
...
     /// Look at the code in `highlighter.rs` for an example of doing this for highlighting purposes.
     ///
-    /// The vector is in order both by index to apply at (the `usize`) and also by order to apply them at a
+    /// The returned vector is in order both by index to apply at (the `usize`) and also by order to apply them at a
     /// given index (e.g popping old scopes before pushing new scopes).
-    pub fn parse_line(&mut self, line: &str) -> Vec<(usize, ScopeStackOp)> {
+    ///
+    /// The `SyntaxSet` has to be the one that contained the syntax that was
+    /// used to construct this `ParseState`, or an extended version of it.
+    /// Otherwise the parsing would return the wrong result or even panic. The
```
trishume: Thanks for the docs!

One way we could mitigate this is by creating a "token" for each syntax set builder as well as a generation that gets bumped on each syntax added. The `SyntaxReference` would also keep the token and generation from its creation. `parse_line` would then `assert!` that the token of the `SyntaxSet` matches and the generation was greater or equal. This would lead to better error messages and never mysterious incorrect results, and the cost should be negligible.

The token could either be a random number (requiring a `rand` dependency) or a `Box<u8>` in the `SyntaxSet` and a `*const u8` in the `SyntaxReference` because comparing pointers is safe and doesn't need lifetimes, and the box guarantees uniqueness and non-movingness. The `u8` is just because I'm unsure if `Box<()>` actually allocates a different address for each one.

You very much don't have to do this. I think the current state of dropping the `SyntaxSet` causing panics is easier to get into and worse than this. Just bonus if you feel like it.

src/parsing/parser.rs
```diff
@@ -208,25 +197,29 @@ impl<'a> ParseState<'a> {
...
 
-        while self.parse_next_token(line,
-                                    &mut match_start,
-                                    &mut search_cache,
-                                    &mut regions,
-                                    &mut non_consuming_push_at,
-                                    &mut res) {
-        }
+        while self.parse_next_token(
+            line,
```
trishume: 👍 I also prefer this style, but only because my editor isn't set up to do the deep indents properly.

src/parsing/parser.rs
```diff
@@ -811,11 +822,26 @@ mod tests {
...
+            ParseState::new(syntax)
         };
 
         // test fix for issue #25
-        assert_eq!(ops("struct{estruct", &mut state).len(), 10);
+        assert_eq!(ops(&mut state, "struct{estruct", &ss).len(), 10);
+    }
+
+    #[test]
+    fn can_compare_parse_states() {
```
trishume: ❤️

src/parsing/parser.rs
robinst: Yeah I think this is something we can easily do after we merge the big PR :).

As for the implementation, why wouldn't we just store a copy of the same `u8` on the `SyntaxReference` directly?

src/parsing/parser.rs
robinst: Yeah, that's the new standard rustfmt style too. I think we should run `rustfmt` on the whole codebase once the big PR is in.

robinst:
So I ran the benchmark with the base branch and this, here's the results:

```
parse/"highlight_test.erb"
                        time:   [1.4563 ms 1.4663 ms 1.4828 ms]
                        change: [-29.224% -18.677% -7.9199%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 20 measurements (10.00%)
  1 (5.00%) high mild
  1 (5.00%) high severe
parse/"InspiredGitHub.tmTheme"
                        time:   [17.825 ms 18.104 ms 18.336 ms]
                        change: [-12.909% -10.030% -7.3582%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 20 measurements (5.00%)
  1 (5.00%) high mild
parse/"Ruby.sublime-syntax"
                        time:   [73.421 ms 77.776 ms 82.579 ms]
                        change: [-16.879% -11.979% -6.6834%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 20 measurements (5.00%)
  1 (5.00%) high severe
parse/"jquery.js"       time:   [605.69 ms 667.48 ms 756.20 ms]
                        change: [+1.7598% +11.291% +22.238%] (p = 0.02 < 0.05)
                        Performance has regressed.
Found 2 outliers among 20 measurements (10.00%)
  1 (5.00%) high mild
  1 (5.00%) high severe
parse/"parser.rs"       time:   [363.98 ms 374.33 ms 388.59 ms]
                        change: [+4.2024% +7.3198% +11.463%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 2 outliers among 20 measurements (10.00%)
  1 (5.00%) high mild
  1 (5.00%) high severe
parse/"scope.rs"        time:   [31.097 ms 31.270 ms 31.466 ms]
                        change: [-9.1224% -4.2079% -0.7480%] (p = 0.06 > 0.05)
                        No change in performance detected.
Found 1 outliers among 20 measurements (5.00%)
  1 (5.00%) high severe
```

Note that I ran the jquery one again (so both "before" and "after" is from this branch and it changed):

```
parse/"jquery.js"       time:   [554.73 ms 557.23 ms 560.57 ms]
                        change: [-19.530% -12.552% -6.4291%] (p = 0.00 < 0.05)
                        Performance has improved.
```

So not very conclusive unfortunately.

But yeah, given that the overwhelming amount of time is spent doing regex matching, I'm not concerned either. Will merge this!

src/parsing/parser.rs
trishume: Because the `u8` would actually always be `0`, it's just there to make it nonzero sized. The `Box` would be doing all the work by using the system allocator to allocate a globally unique value. Actually a static `AtomicUsize` would also work for allocating unique tokens and probably make more sense.

PR #191, opened 2018-07-28
Update oniguruma

kornelski:
Disables oniguruma's feature of replacing POSIX regex APIs (https://github.com/rust-onig/rust-onig/pull/85) which make libgit2 crash when used in the same executable as syntect (https://github.com/libgit2/libgit2/issues/4738).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ exclude = [
 
 [dependencies]
 yaml-rust = { version = "0.4", optional = true }
-onig = { version = "3.2.1", optional = true }
+onig = { version = "4.0.0", default-features = false, optional = true }
 walkdir = "2.0"
 regex-syntax = { version = "0.4", optional = true }
 lazy_static = "1.0"


trishume:
Thanks!

trishume:
Well frick, this broke CI on master despite passing on the branch, not sure why yet. Looks like linking is failing with a bunch of undefined references for some reason.

Gonna revert, investigate, then re-merge when I figure out why.

trishume:
I can reproduce this locally (just by building the failing commits, no special effort) but it may take me a while to get around to motivating myself to delve into linker hell. In the mean time if anybody has any ideas as to what the problem might be I'm all ears.

robinst:
Maybe @iwillspeak can help?

iwillspeak:
Ok. I'm seeing the same problem on *macOS*. Ony with Syntect though, the unit tests for `onig` itself seem to link fine.

The build script output on my machine ends with the following:

```
l/Repositories/syntect/target/debug/build/onig_sys-2855c31893d9fb7d/out/oniguruma/src/unicode_fold3_key.o"
exit code: 0
TARGET = Some("x86_64-apple-darwin")
cargo:rustc-link-lib=static=onig
cargo:rustc-link-search=native=/Users/will/Repositories/syntect/target/debug/build/onig_sys-2855c31893d9fb7d/out
```

Which looks correct. If I check that directory there is a `libonig.a` file. Not sure why it isn't being linked.

kornelski:
Change

```
onig = { version = "4.0.0", default-features = false, optional = true }
```

to

```toml
onig = { version = "4.0.0", optional = true }
```

I incorrectly assumed syntect doesn't use regexec.c that's for replacing POSIX regex APIs.

iwillspeak:
Should be fixed in `onig` now. Try updating to version `4.1`

PR #194, opened 2018-07-31
Use bufread version of ZlibDecoder

robinst:
For the default dumps, we have a `&[u8]`. `BufRead` is implemented for
that already, so we can use that to avoid another buffer.

See flate2's code here, the `read` version of `ZlibDecoder` delegates to the `bufread` one and wraps the input in a `BufReader`: https://github.com/alexcrichton/flate2-rs/blob/master/src/zlib/read.rs#L184

### DIFF
--- a/src/dumps.rs
+++ b/src/dumps.rs
@@ -14,7 +14,7 @@ use bincode::deserialize_from;
 use bincode::serialize_into;
 use std::fs::File;
 #[cfg(any(feature = "dump-load", feature = "dump-load-rs"))]
-use std::io::{BufReader, Read};
+use std::io::{BufRead, BufReader};
 #[cfg(any(feature = "dump-create", feature = "dump-create-rs"))]
 use std::io::{BufWriter, Write};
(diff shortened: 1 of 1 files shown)


robinst:
Note that loading benchmark doesn't show a change in performance, but should not hurt to avoid some copying anyway.

robinst:
Not directly related to this, but I've benchmarked what effect removing compression has for the dumps and it's a big difference:

```
load_internal_dump      time:   [17.695 ms 18.144 ms 18.652 ms]
                        change: [-48.328% -44.551% -41.800%] (p = 0.00 < 0.05)
                        Performance has improved.
```

It's probably not something that we want to do for the shipped dumps because they increase to 1.8 MB, but it might be worth noting in the README?

trishume:
Looks good thanks!

I didn't know the compression perf hit was that large, but indeed I'll still take the compression for shipped dumps. It probably doesn't matter much since it's pretty fast, especially now that people won't have to do thread local loading anymore, but if you want to note it in the docs go ahead.

PR #195, opened 2018-08-02
Rename methods on SyntaxSetBuilder

robinst:
While we're breaking API anyway, I think this makes them a bit clearer:

* `add_syntax` -> `add`
* `load_syntaxes` -> `load_from_folder`
* `load_plain_text_syntax` -> `add_plain_text_syntax`

### DIFF
--- a/src/parsing/syntax_set.rs
+++ b/src/parsing/syntax_set.rs
@@ -99,13 +99,16 @@ impl SyntaxSet {
         SyntaxSet::default()
     }
 
-    /// Convenience constructor calling `new` and then `load_syntaxes` on the resulting set
-    /// defaults to lines given not including newline characters, see the
-    /// `load_syntaxes` method docs for an explanation as to why this might not be the best.
+    /// Convenience constructor for creating a builder, then loading syntax
+    /// definitions from a folder and then building the syntax set.
+    ///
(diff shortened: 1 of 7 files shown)


src/parsing/syntax_set.rs
```diff
@@ -292,22 +295,22 @@ impl SyntaxSetBuilder {
...
     /// Add a syntax to the set.
-    pub fn add_syntax(&mut self, syntax: SyntaxDefinition) {
+    pub fn add(&mut self, syntax: SyntaxDefinition) {
         self.syntaxes.push(syntax);
     }
 
     /// Rarely useful method that loads in a syntax with no highlighting rules for plain text.
     /// Exists mainly for adding the plain text syntax to syntax set dumps, because for some
     /// reason the default Sublime plain text syntax is still in `.tmLanguage` format.
     #[cfg(feature = "yaml-load")]
```
robinst: I noticed that this needs YAML, but we could change it to instead just create the structs directly. Any thoughts?

trishume: Sure if you want to do the translation, submit a separate PR and I'll approve.

trishume:
I have a breaking change I may want to make to the `html` API, and then may review #185, then I'll look at drafting 3.0 in a few days ✨.

PR #199, opened 2018-08-15
 Add filename to ParseSyntax error & update dumps

Keats:
When loading from a folder with dozens of syntaxes, it's nice to know immediately which one failed loading


Keats:
I removed the packages update

PR #200, opened 2018-08-20
Add `ThemeSet::load_themes` method

sharkdp:
This adds a new `ThemeSet::load_themes` method (in analogy to `SyntaxSet::load_syntaxes`) which can be used to extend an existing `ThemeSet` by loading all files from a given folder into the set.

It also adds a `ThemeSet::new` function to make it possible to start from an empty set.

It is possible to [implement these methods](https://github.com/sharkdp/bat/blob/052425b12f79aa42b1ca6453725b501be87dcc8d/src/assets.rs#L197-L209) without adding them to syntect (because everything in `ThemeSet` is public), but I thought these methods might be useful for others.

### DIFF
--- a/src/highlighting/theme_set.rs
+++ b/src/highlighting/theme_set.rs
@@ -46,12 +46,12 @@ impl ThemeSet {
     /// Generate a `ThemeSet` from all themes in a folder
     pub fn load_from_folder<P: AsRef<Path>>(folder: P) -> Result<ThemeSet, LoadingError> {
         let mut theme_set = Self::new();
-        theme_set.load_themes(folder)?;
+        theme_set.add_from_folder(folder)?;
         Ok(theme_set)
     }
 
     /// Load all the themes in the folder into this `ThemeSet`
-    pub fn load_themes<P: AsRef<Path>>(&mut self, folder: P) -> Result<(), LoadingError> {
+    pub fn add_from_folder<P: AsRef<Path>>(&mut self, folder: P) -> Result<(), LoadingError> {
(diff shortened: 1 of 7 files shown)


src/highlighting/theme_set.rs
```diff
@@ -60,6 +72,8 @@ mod tests {
     fn can_parse_common_themes() {
         let themes = ThemeSet::load_from_folder("testdata").unwrap();
         let all_themes: Vec<&str> = themes.themes.keys().map(|x| &**x).collect();
+        assert!(all_themes.contains(&"base16-ocean.dark"));
```
sharkdp: Not sure if this should be added, but the existing `load_from_folder` tests would simply print out something, so I thought this would be helpful.

trishume:
I don't think `ThemeSet` needs a builder, that's mostly needed because syntaxes can reference each other, which even the new ST3 theme format we don't support doesn't do.

I don't think the inconsistency is big enough to warrant using worse names on `SyntaxSet(Builder)`. This is somewhat consistent in that both have `load_from_folder` as the primary way of loading from a folder.

robinst:
Hmm yeah. What about if we change the methods that add additional things `add_from_folder`? With "load" it's not super clear that it adds (as opposed to replacing).

trishume:
So changing both the `ThemeSet` and `SyntaxSetBuilder` methods to `add`? That sounds good to me.

sharkdp:
I've done the renaming (hoping I understood you correctly).

PR #207, opened 2018-09-09
Use newlines syntaxes in HTML and docs, fixes #196

### LINKED ISSUE #196: Encourage use of lines with newline characters
I'd like to change the HTML APIs and the doc examples to use the `newlines` version of the syntax sets. The `nonewlines` version is a hack and sometimes provides incorrect results.

Here's my proposed plan:
- [x] @robinst can you contribute your iterator from [your Stack Overflow post](https://stackoverflow.com/questions/40455997/iterate-over-lines-in-a-string-including-the-newline-characters) to syntect under the `easy` or `util` module with a doc comment explaining how it lets you use the `newlines` version with strings? Stack Overflow's default license isn't compatible with me copy-pasting it AFAIK.
- [ ] Use that iterator in the `html::highlighted_snippet_for_string` and the IO with newline reader loop in `highlighted_snippet_for_file`.
- [ ] Rename both of those methods, not because they particularly need renaming but so that people upgrading don't miss the breaking change and get subtly broken behaviour.
- [ ] Change all the examples to use the `newlines` version of the syntax sets as required

trishume:
This PR switches the HTML helpers to use newlines syntaxes so that they are faster and more correct and so fixes #196. It also refactors them to not allocate intermediate strings and share more code.

It renames the methods which had silent requirements changes so that people
upgrading get compile errors and hopefully then look at the release notes.

There are two other functionality changes: the position of the newline characters in the resulting HTML changed, in a way I think is reasonable and doesn't change the rendered output. And the `start_coloured_html_snippet` function got a semicolon added and now returns the background.

I also changed a number of doc tests to use newlines syntaxes so that the slightly more complex things you have to do to use the newlines syntaxes are demonstrated.

This is the last set of breaking API changes I wanted to do before the 3.0 release, so after this is done with I can work on getting the release out. Although it's plausible #203 might require breaking API changes, I should take a look before releasing.

### DIFF
--- a/src/html.rs
+++ b/src/html.rs
@@ -145,7 +145,7 @@ fn write_css_color(s: &mut String, c: Color) {
 /// use syntect::html::{styles_to_coloured_html, IncludeBackground};
 ///
 /// // Load these once at the start of your program
-/// let ps = SyntaxSet::load_defaults_nonewlines();
+/// let ps = SyntaxSet::load_defaults_newlines();
 /// let ts = ThemeSet::load_defaults();
 ///
 /// let syntax = ps.find_syntax_by_name("Ruby").unwrap();
--- a/src/html.rs
+++ b/src/html.rs
@@ -32,7 +32,7 @@ fn scope_to_classes(s: &mut String, scope: Scope, style: ClassStyle) {
     }
 }
 
-/// Convenience method that combines `start_coloured_html_snippet`, `styles_to_coloured_html`
+/// Convenience method that combines `start_highlighted_html_snippet`, `styled_line_to_highlighted_html`
 /// and `HighlightLines` from `syntect::easy` to create a full highlighted HTML snippet for
 /// a string (which can contain many lines).
 ///
@@ -40,17 +40,17 @@ fn scope_to_classes(s: &mut String, scope: Scope, style: ClassStyle) {
 /// This is easy to get with `SyntaxSet::load_defaults_newlines()`. (Note: this was different before v3.0)
 pub fn highlighted_html_for_string(s: &str, ss: &SyntaxSet, syntax: &SyntaxReference, theme: &Theme) -> String {
     let mut highlighter = HighlightLines::new(syntax, theme);
-    let (mut output, bg) = start_coloured_html_snippet(theme);
+    let (mut output, bg) = start_highlighted_html_snippet(theme);
 
     for line in LinesWithEndings::from(s) {
         let regions = highlighter.highlight(line, ss);
-        append_coloured_html_for_styles(&regions[..], IncludeBackground::IfDifferent(bg), &mut output);
+        append_highlighted_html_for_styled_line(&regions[..], IncludeBackground::IfDifferent(bg), &mut output);
     }
     output.push_str("</pre>\n");
     output
 }
 
-/// Convenience method that combines `start_coloured_html_snippet`, `styles_to_coloured_html`
+/// Convenience method that combines `start_highlighted_html_snippet`, `styled_line_to_highlighted_html`
 /// and `HighlightFile` from `syntect::easy` to create a full highlighted HTML snippet for
 /// a file.
 ///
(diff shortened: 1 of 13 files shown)


src/html.rs
```diff
@@ -214,34 +206,34 @@ pub fn styles_to_coloured_html(v: &[(Style, &str)], bg: IncludeBackground) -> St
...
+/// `highlighted_html_for_string`.
 ///
 /// If you don't care about the background color you can just prefix the lines from
 /// `styles_to_coloured_html` with a `<pre>`. This is meant to be used with `IncludeBackground::IfDifferent`.
+/// As of `v3.0` this method also returns the background color to be passed to `IfDifferent`.
 ///
 /// You're responsible for creating the string `</pre>` to close this, I'm not gonna provide a
 /// helper for that :-)
-pub fn start_coloured_html_snippet(t: &Theme) -> String {
+pub fn start_coloured_html_snippet(t: &Theme) -> (String, Color) {
```
robinst: Noticed that the function name uses British English spelling `coloured`. I'd change it to `color` to make it consistent with the type. Also it's more common for code to use US English spelling anyway.

Same for `append_coloured_html_for_styles` and other methods.

src/html.rs
```diff
@@ -35,24 +36,15 @@ fn scope_to_classes(s: &mut String, scope: Scope, style: ClassStyle) {
...
 /// a string (which can contain many lines).
 ///
-/// Note that the `syntax` passed in must be from a `SyntaxSet` compiled for no newline characters.
-/// This is easy to get with `SyntaxSet::load_defaults_nonewlines()`. If you think this is the wrong
-/// choice of `SyntaxSet` to accept, I'm not sure of it either, email me.
-pub fn highlighted_snippet_for_string(s: &str, ss: &SyntaxSet, syntax: &SyntaxReference, theme: &Theme) -> String {
-    let mut output = String::new();
+/// Note that the `syntax` passed in must be from a `SyntaxSet` compiled for newline characters.
+/// This is easy to get with `SyntaxSet::load_defaults_newlines()`. (Note: this was different before v3.0)
+pub fn highlighted_html_for_string(s: &str, ss: &SyntaxSet, syntax: &SyntaxReference, theme: &Theme) -> String {
```
robinst: (I actually like the new function name better anyway 👍.)

testdata/test3.html
```diff
@@ -1,32 +1,32 @@
...
-<span style="color:#c0c5ce;">    border: </span><span style="color:#d08770;">5px </span><span style="color:#96b5b4;">#ffffff</span><span style="color:#c0c5ce;">;</span>
-<span style="color:#c0c5ce;">    background: </span><span style="color:#96b5b4;">url</span><span style="color:#c0c5ce;">(&quot;</span><span style="color:#a3be8c;">wow</span><span style="color:#c0c5ce;">&quot;);</span>
-<span style="color:#c0c5ce;">  }</span>
-<span style="color:#c0c5ce;">&lt;/</span><span style="color:#bf616a;">style</span><span style="color:#c0c5ce;">&gt;</span>
-</pre>
+<span style="color:#c0c5ce;">&lt;</span><span style="color:#bf616a;">script </span><span style="color:#d08770;">type</span><span style="color:#c0c5ce;">=&quot;</span><span style="color:#a3be8c;">text/javascript</span><span style="color:#c0c5ce;">&quot;&gt;
+</span><span style="color:#c0c5ce;">  </span><span style="color:#b48ead;">var </span><span style="color:#bf616a;">lol </span><span style="color:#c0c5ce;">= &quot;</span><span style="color:#a3be8c;">JS nesting</span><span style="color:#c0c5ce;">&quot;;
```
robinst: Would it be easy to add a newline here after the closing tag (in addition to the one from the line itself)? Seems a bit weird to have an end tag for the previous line and then start the open tag for the next line on the same line.

trishume: I think the answer is no since this is in a `<pre>` tag so newlines are actually significant, note the lack of `<br>` tags.

trishume:
@robinst Done. I find/replaced "colour", I had already switched to US spelling at some point but there were instances left over or from not paying attention.

I then realized that "colored" wasn't the best name for the HTML helpers because they did both text colour, background colour and font style. So I made them more consistent with the file and string methods by using "highlighted" and then made it clear from the names that they are to be used on single lines.

testdata/test3.html
robinst: Ooh yes, good point :)

PR #209, opened 2018-09-23
Reengineer theme application to match Sublime

### LINKED ISSUE #133: Color Scheme rules applied differently to Sublime

trishume:
Fixes #133 and additional problems discovered in #203 while not noticeably regressing end-to-end performance. 

This supersedes #203 and I'll close it and point to this PR. This PR does use the main piece of code from #203 to reimplement one of the highlighter methods as well as the ideas and cases from that PR so thanks @keith-hall!

It took a number of hours and a few false starts to figure out a way to be correct like #203 while not regressing performance but I came up with a way that is asymptotically the same as current syntect master and with a small difference in constant factors that is within the noise of the current benchmarks.

It does this by caching the matching styles and scores of single-scope selectors (the majority) because pushing additional things can't make them un-match. This doesn't apply to multi-scope selectors with exclusions which are applied at the end to finalize the style.

This PR removes some previously public methods that had incorrect behaviour. I don't think anyone used them but this PR should go in before the major release anyhow to be semver compliant.

After this PR is reviewed and merged that should _finally_ be everything necessary before the 3.0 release.

### DIFF
--- a/src/highlighting/highlighter.rs
+++ b/src/highlighting/highlighter.rs
@@ -7,7 +7,7 @@ use std::iter::Iterator;
 
 use parsing::{Scope, ScopeStack, BasicScopeStackOp, ScopeStackOp, MatchPower, ATOM_LEN_BITS};
 use super::selector::ScopeSelector;
-use super::theme::Theme;
+use super::theme::{Theme, ThemeItem};
 use super::style::{Color, FontStyle, Style, StyleModifier};
 
 /// Basically a wrapper around a `Theme` preparing it to be used for highlighting.
@@ -47,6 +47,7 @@ pub struct Highlighter<'a> {
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct HighlightState {
     styles: Vec<Style>,
+    single_caches: Vec<ScoredStyle>,
     pub path: ScopeStack,
 }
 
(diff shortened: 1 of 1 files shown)


src/highlighting/highlighter.rs
```diff
@@ -69,15 +70,17 @@ impl HighlightState {
...
-            initial_styles.push(style);
+            let prefix = initial_stack.bottom_n(i);
+            let new_cache = highlighter.update_single_cache_for_push(&single_caches[i], prefix);
+            styles.push(highlighter.finalize_style_with_multis(&new_cache, prefix));
+            single_caches.push(new_cache);
         }
 
         HighlightState {
-            styles: initial_styles,
+            styles, single_caches,
```
robinst: Mind putting this on its own line? I was a bit confused until I realized that this is a `,` not a `:`.

trishume:
@robinst yah a rustfmt PR would be fine with me, although I'm somewhat inclined to wait for the "coming soon" 1.0 release where they say the formatting will be stable.

robinst:
Let's wait for that then, we're not in a hurry :)

Issue #204, opened 2018-09-02
Release new version with updated deps?


ignatenkobrain:
We want to package syntect in Fedora, but some of deps are old in released version. Could you please release a new version with those fixes?

robinst:
A new major release is planned soon. On master, the dependencies should be up-to-date. Would you mind checking if master is good?

trishume:
Because of the way Cargo and semver works it should be fine to use syntect with any newer minor versions. Trouble should only arise with newer major versions, are there are new major versions master is missing that you need?

robinst:
Maybe what would really solve @ignatenkobrain's problem is if we started to include a `Cargo.lock` in the repository with the newest versions? AFAIK we could do that and it would not affect projects that pull in syntect as a dependency, as cargo skips the lock file in that case. It would only affect building syntect from a checkout.

But that's just a guess, I'm not familiar with how Fedora's packaging works.

trishume:
I'm not sure I'm understanding what's going wrong correctly but it sounds like your package importer may not behave ideally. It should accept any newer minor version of the dependencies and should not fail if the `Cargo.toml` lists an older minor version, only an older major version. The `Cargo.toml` version means the oldest point release of the major version that has all the features needed.

If the problem is actually us using an outdated major version even on `master`, let us know.

ignatenkobrain:
No, in master everything is good. Problem is with latest version on crates.io

trishume:
Now released as `v3.0.0`

PR #210, opened 2018-10-02
Update Dependencies

Caemor:
As requested in #204 I took a look at the dependencies and update them.
The biggest change is in walkdir, which sets it's minimum rust version 1.23.
Finally I tested the new versions with `cargo test`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ onig = { version = "4.2", optional = true }
 walkdir = "2.2"
 regex-syntax = { version = "0.6", optional = true }
 lazy_static = "1.1.0"
-lazycell = "1.0"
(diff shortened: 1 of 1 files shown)


trishume:
So unless I'm misunderstanding how Cargo and semver work, this only makes syntect more strict about dependencies for no reason, since there's no major version bumps. 

My understanding: Any new project that depends on syntect should already have the latest minor version selected by Cargo, the minor version in `Cargo.toml` specifies the minimum minor version needed for the features that syntect uses. Bumping those numbers without changing any code to use the new features just makes things unnecessarily strict.

This is kinda fancy and tripped me up when I started using Rust.

Caemor:
Yes it seems I misunderstood the updates.

Issue #211, opened 2018-10-02
Elm syntax does not compile on master anymore


Keats:
The syntaxes are located at https://github.com/elm-community/SublimeElmLanguageSupport/tree/master/Syntaxes . I added the repo to the packages folder of master for trying it quickly.

Error message while running `make packs` in syntect master

```bash
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ParseSyntax(RegexCompileError("$+(?!^ {4})", Error(-114, target of repeat operator is invalid)), Some("testdata/Packages/SublimeElmLanguageSupport/Syntaxes/Elm Compile Messages.sublime-syntax"))', libcore/result.rs:1009:5
```

Offending line: https://github.com/elm-community/SublimeElmLanguageSupport/blob/master/Syntaxes/Elm%20Compile%20Messages.sublime-syntax#L67

keith-hall:
I wonder if this is related to the regex rewriting for no newlines mode, as it seems to be converting the `\n` to `$`, and as it is followed by a `+`, it's seen as invalid by Oniguruma.

robinst:
Yeah looks like nonewlines mode. We could tweak the `\n` to `$` rewriting some more, but fixing the grammar would also be good.

Looks like the line to tweak for the rewriting is here: https://github.com/trishume/syntect/blob/master/src/parsing/yaml_load.rs#L561

Keats:
So the line in the grammar should be `match: '(?!^ {4})'` ?

trishume:
Can someone confirm the syntax compiles now that the fix is merged?

Keats:
It does work

PR #212, opened 2018-10-04
Tweak nonewlines regex rewriting (#211)

robinst:
Using `\n+` in a regex doesn't make any sense for Sublime Text syntaxes, but let's not produce an invalid regex in that case.

### DIFF
--- a/src/parsing/yaml_load.rs
+++ b/src/parsing/yaml_load.rs
@@ -556,9 +556,10 @@ impl<'a> RegexRewriter<'a> {
                     self.next();
                     if let Some(c2) = self.peek() {
                         self.next();
-                        // Replacing `\n?` with `$?` would make parsing later fail with
-                        // "target of repeat operator is invalid"
-                        if c2 == b'n' && self.peek() != Some(b'?') {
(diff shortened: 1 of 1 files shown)


src/parsing/yaml_load.rs
```diff
@@ -556,9 +556,10 @@ impl<'a> RegexRewriter<'a> {
                     self.next();
                     if let Some(c2) = self.peek() {
                         self.next();
-                        // Replacing `\n?` with `$?` would make parsing later fail with
-                        // "target of repeat operator is invalid"
-                        if c2 == b'n' && self.peek() != Some(b'?') {
+                        // Replacing `\n` with `$` in `\n?` or `\n+` would make parsing later fail
+                        // with "target of repeat operator is invalid"
+                        let c3 = self.peek();
+                        if c2 == b'n' && c3 != Some(b'?') && c3 != Some(b'+') {
```
keith-hall: I think it would make sense to also check for `*` here, as `\n*` -> `$*` would also be invalid

robinst: Makes sense, done!

Issue #205, opened 2018-09-03
Incorrect highlighting with .fish files


orf:
Given the following small snippet of fish code:

```abc.fish
test -n "$_CONDA_EXE"; or set _CONDA_EXE "$PWD/shell/bin/conda"

test -n "$CONDA_SHLVL"; or set -gx CONDA_SHLVL "0"

function __conda_add_prompt
  if set -q CONDA_DEFAULT_ENV
      set_color normal
      echo -n '('
      set_color -o green
      echo -n $CONDA_DEFAULT_ENV
      set_color normal
      echo -n ') '
  end
end
```

Syntect seems to render the following, at least through `bat`:

![image](https://user-images.githubusercontent.com/1027207/44990717-d09b0200-af89-11e8-887b-92196c587a5a.png)

Not sure if this should be reported here or in `bat`, so sorry if this is the wrong place 👍

sharkdp:
This is how it looks like in Sublime Text:
![image](https://user-images.githubusercontent.com/4209276/44999824-62256680-afc0-11e8-8e98-2a213d738537.png)

It looks slightly different (I'm not sure if my Sublime Text uses the same version of sublimehq/Packages as `bat`) but the main issue is definitely that the Fish syntax is quite different from Bash syntax. So it really looks like a sublimehq/Packages issue. 

Syntect users (such as `bat`) could potentially solve the problem by adding a dedicated Fish syntax (and possibly also Zsh), but I'm not sure how `file_extension` conflicts are handled (which `.sublime-syntax` takes precedence?)

keith-hall:
I believe precedence works in Sublime Text by lexicographical order https://github.com/guillermooo/sublime-undocs/blob/a571f5ba987176607f12bb141965bf4596775eeb/source/extensibility/packages.rst#id62, so a package named `ShellScriptFish` would take precedence over `ShellScriptBash` (for example) were they to have conflicting file extension mappings. Users of Sublime Text can also set a preference to open particular extensions with a particular syntax, which always takes priority over what is in the syntax definition.

keith-hall:
[...]

PR #217, opened 2018-10-13
when searching for a syntax, search in reverse order

keith-hall:
this will prioritize syntaxes loaded last over those loaded first when searching for a suitable syntax by any means. As syntaxes are loaded deterministically in lexicographical order, this matches Sublime Text's behavior in case of conflicts, and should allow custom syntaxes to be used in favor of "built-in" syntaxes, which is useful for #205.

### DIFF
--- a/src/parsing/syntax_set.rs
+++ b/src/parsing/syntax_set.rs
@@ -127,15 +127,15 @@ impl SyntaxSet {
     /// This and all similar methods below do a linear search of syntaxes, this should be fast
     /// because there aren't many syntaxes, but don't think you can call it a bajillion times per second.
     pub fn find_syntax_by_scope(&self, scope: Scope) -> Option<&SyntaxReference> {
-        self.syntaxes.iter().find(|&s| s.scope == scope)
+        self.syntaxes.iter().rev().find(|&s| s.scope == scope)
     }
 
     pub fn find_syntax_by_name<'a>(&'a self, name: &str) -> Option<&'a SyntaxReference> {
-        self.syntaxes.iter().find(|&s| name == &s.name)
+        self.syntaxes.iter().rev().find(|&s| name == &s.name)
     }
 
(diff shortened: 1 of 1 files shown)


src/parsing/syntax_set.rs
```diff
@@ -572,6 +572,7 @@ impl FirstLineCache {
                 }
             }
         }
+        regexes.reverse();
```
robinst: Why not do `rev` in `find_syntax_by_first_line`, same as the others?

keith-hall: because, when I tried `in &cache.regexes.iter().rev()`, I got some compiler error I didn't know how to work around and took the simplest solution I could get working :stuck_out_tongue: 
I'm planning to add some tests when I get time :+1:

keith-hall: I have now added some tests. If you want us to use a more consistent way to do a reverse search of the first line match regexes, please feel free to push a new commit to this PR with the desired implementation - it's beyond my Rust skills at the moment, unfortunately.

robinst: This works (just removed the `&`):

```
for &(ref reg, i) in cache.regexes.iter().rev() {
```

(I can't push a commit for this PR as it's from your fork.)

src/parsing/syntax_set.rs
```diff
@@ -736,6 +739,59 @@ mod tests {
...
+            
+            builder.build()
+        };
+
+        let mut syntax = syntax_set.find_syntax_by_extension("a").unwrap();
+        assert_eq!(syntax.name, "A improved");
+        syntax = syntax_set.find_syntax_by_scope(Scope::new(&"source.a").unwrap()).unwrap();
+        assert_eq!(syntax.name, "A improved");
+        syntax = syntax_set.find_syntax_by_first_line(&"syntax a").unwrap();
+        assert_eq!(syntax.name, "C");
```
robinst: Nice tests :)

Issue #178, opened 2018-06-13
multiple with_prototypes at the same stack level are currently unsupported


keith-hall:
For example, this test, if put in `parser.rs`, would fail:

```rust
#[test]
    fn can_parse_two_with_prototypes_at_same_stack_level() {
        let syntax = r#"
%YAML 1.2
---
# See http://www.sublimetext.com/docs/3/syntax.html
scope: source.example-wp
contexts:
  main:
    - match: a
      scope: a
      push:
        - match: b
          scope: b
          set:
            - match: c
              scope: c
          with_prototype:
            - match: '2'
              scope: '2'
      with_prototype:
        - match: '1'
          scope: '1'
"#;

        let syntax = SyntaxDefinition::load_from_str(&syntax, true, None).unwrap();
        expect_scope_stacks_with_syntax("abc12", &["<1>", "<2>"], syntax);
    }
```

```
failures:

---- parsing::parser::tests::can_parse_two_with_prototypes_at_same_stack_level stdout ----
	abc12
^ +source.example-wp
abc12
^ +a
abc12
 ^ pop 1
abc12
 ^ +b
abc12
  ^ pop 1
abc12
  ^ +c
abc12
   ^ pop 1
abc12
    ^ +2
abc12
     ^ pop 1
ScopeStack { clear_stack: [], scopes: [<source.example-wp>] }
ScopeStack { clear_stack: [], scopes: [<source.example-wp>, <a>] }
ScopeStack { clear_stack: [], scopes: [<source.example-wp>] }
ScopeStack { clear_stack: [], scopes: [<source.example-wp>, <b>] }
ScopeStack { clear_stack: [], scopes: [<source.example-wp>] }
ScopeStack { clear_stack: [], scopes: [<source.example-wp>, <c>] }
ScopeStack { clear_stack: [], scopes: [<source.example-wp>] }
ScopeStack { clear_stack: [], scopes: [<source.example-wp>, <2>] }
ScopeStack { clear_stack: [], scopes: [<source.example-wp>] }
thread 'parsing::parser::tests::can_parse_two_with_prototypes_at_same_stack_level' panicked at 'expected scope stack '<1>' missing', src/parsing/parser.rs:1324:13
```

but using this syntax definition in Sublime Text, the `1` and the `2` both receive scopes.

Issue #160, opened 2018-05-08
Java syntax highlighting wrong (JavaDoc does not end)


sharkdp:
The following Java file

``` java
/**
 * test
 */
class ExampleClass {
}
```

renders correctly with `syncat` and sublimehq/Packages pinned to fa6b862 (see testdata/Packages):

![image](https://user-images.githubusercontent.com/4209276/39777798-db75b1ac-5304-11e8-8a64-b318d1acce85.png)

However, when updating sublimehq/Packages to master (currently at 0f852a3) ...

``` bash
(cd testdata/Packages; git fetch; git checkout master)
make packs
```

the docstring highlighting does not end:

![image](https://user-images.githubusercontent.com/4209276/39777746-aa904570-5304-11e8-81a6-82679809b9be.png)

I'm not sure if syntect always aims to support the current state of sublimehq/Packages, but I thought it might be worth to report this.

Keats:
Bisecting shows it come from https://github.com/sublimehq/Packages/pull/1351
It's a rewrite of JavaDoc so not easy to pinpoint the reason, unless @keith-hall can magically figure out the issue from looking at the regexes like he seems to be able to!

Since it keeps treating the rest of the class as JavaDoc, I'm guessing some context is not popped? Related to #104 maybe?

slimsag:
I wasted a bit of time due to not finding this issue earlier, but hopefully someone else can benefit from what I've discovered here. I've managed to boil down the JavaDoc syntax into this much smaller form which demonstrates the issue:

```yaml
%YAML 1.2
---
name: Javadoc
scope: text.html.javadoc
contexts:
  prototype:
    - match: ^(\*)
      captures:
        1: punctuation.definition.comment.javadoc

  main:
    - meta_include_prototype: false
    - match: /\*\*
      scope: comment.block.documentation.javadoc punctuation.definition.comment.begin.javadoc
      embed: contents
      embed_scope: comment.block.documentation.javadoc text.html.javadoc
      escape: \*/
      escape_captures:
        0: comment.block.documentation.javadoc punctuation.definition.comment.end.javadoc

  contents:
    - match: ''
      scope: comment.block.documentation.javadoc
```

I am using the newlines (NOT no newlines) mode with a patched version of synhtml, which produces the following output:

### Test case 1

```java
package org.joda.time.chrono;/**
**/class BasicChronology extends AssembledChronology {
}
```

[...]

PR #220, opened 2018-10-17
correctly apply multiple with_prototypes when a context is set

keith-hall:
fixes https://github.com/trishume/syntect/issues/178 and fixes https://github.com/trishume/syntect/issues/160 and also the ASP syntax test failures.

However, due to my limited Rust knowledge, in it's current state, this PR also breaks something - any help anyone can give me here would be highly appreciated! I temporarily removed `lvl.captures.as_ref()` from line 312 of `parser.rs` (replaced it with `None`) as I was getting a compiler error which I don't know how to fix properly:

```
error[E0597]: `lvl` does not live long enough
   --> src\parsing\parser.rs:312:126
    |
312 |             let with_prototypes = self.stack[proto_start..].iter().flat_map(|lvl| lvl.prototype.iter().map(|ctx| (true, ctx, lvl.captures.as_ref())));
    |                                                                                                            -----             ^^^                   - borrowed value only lives until here
    |                                                                                                            |                 |
    |                                                                                                            |                 borrowed value does not live long enough
    |                                                                                                            capture occurs here
...
369 |     }
    |     - borrowed value needs to live until here
```

Does anyone know what I need to do to satisfy the borrow checker please?

### DIFF
--- a/src/parsing/parser.rs
+++ b/src/parsing/parser.rs
@@ -1402,7 +1402,7 @@ contexts:
 
     #[test]
     fn can_parse_two_with_prototypes_at_same_stack_level() {
-        let syntax = r#"
+        let syntax_yamlstr = r#"
 %YAML 1.2
 ---
 # See http://www.sublimetext.com/docs/3/syntax.html
@@ -1425,8 +1425,39 @@ contexts:
           scope: '1'
 "#;
 
-        let syntax = SyntaxDefinition::load_from_str(&syntax, true, None).unwrap();
+        let syntax = SyntaxDefinition::load_from_str(&syntax_yamlstr, true, None).unwrap();
         expect_scope_stacks_with_syntax("abc12", &["<1>", "<2>"], syntax);
+    }
+
+    #[test]
+    fn can_parse_two_with_prototypes_at_same_stack_level_updated_captures() {
+        let syntax_yamlstr = r#"
+%YAML 1.2
+---
+# See http://www.sublimetext.com/docs/3/syntax.html
+scope: source.example-wp
+contexts:
+  main:
+    - match: (a)
+      scope: a
+      push:
+        - match: (b)
+          scope: b
+          set:
+            - match: c
+              scope: c
+          with_prototype:
+            - match: d
+              scope: d
+      with_prototype:
+        - match: \1
+          scope: '1'
+          pop: true
+"#;
+
+        let syntax = SyntaxDefinition::load_from_str(&syntax_yamlstr, true, None).unwrap();
+        expect_scope_stacks_with_syntax("aa", &["<a>", "<1>"], syntax.clone());
+        expect_scope_stacks_with_syntax("abcdb", &["<a>", "<b>", "<c>", "<d>", "<1>"], syntax);
     }
 
     fn expect_scope_stacks(line_without_newline: &str, expect: &[&str], syntax: &str) {
--- a/src/parsing/parser.rs
+++ b/src/parsing/parser.rs
@@ -1436,6 +1436,56 @@ contexts:
     }
 
     #[test]
+    fn can_parse_two_with_prototypes_at_same_stack_level_set_multiple() {
+        let syntax_yamlstr = r#"
+%YAML 1.2
+---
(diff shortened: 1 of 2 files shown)


robinst:
You can use a move closure here:

```
lvl.prototype.iter().map(move |ctx| (true, ctx, lvl.captures.as_ref()))
```

(See [this blog post](https://llogiq.github.io/2017/03/06/lifetime.html))

src/parsing/parser.rs
```diff
@@ -36,7 +36,7 @@ pub struct ParseState {
 #[derive(Debug, Clone, Eq, PartialEq)]
 struct StateLevel {
     context: ContextId,
-    prototype: Option<ContextId>,
+    prototype: Vec<ContextId>,
```
robinst: Should probably be called `prototypes` now?

src/parsing/parser.rs
```diff
@@ -617,29 +619,23 @@ impl ParseState {
             MatchOperation::None => return false,
         };
         for (i, r) in ctx_refs.iter().enumerate() {
-            let proto_id = if i == ctx_refs.len() - 1 {
-                // a `with_prototype` stays active when the context is `set`
-                // until the context layer in the stack (where the `with_prototype`
-                // was initially applied) is popped off.
+            let mut proto_ids = old_proto_ids.clone().unwrap_or(Vec::new());
```
robinst: Not sure if it's necessary in this case, but clippy might recommend doing this instead:

`unwrap_or_else(|| Vec::new())`

The reason is that with `unwrap_or`, the argument is always constructed, even when the option had a value. With `unwrap_or_else`, that only happens when it's actually needed.

testdata/known_syntest_failures.txt
```diff
@@ -1,5 +1,4 @@
 loading syntax definitions from testdata/Packages
-FAILED testdata/Packages/ASP/syntax_test_asp.asp: 162
```
robinst: 🎉 very nice!

Have you checked if it happens to fix any others of the syntests?

keith-hall: it doesn't fix any other tests at the current submodule version, but will allow us to [update the sublimehq/Packages submodule](https://github.com/trishume/syntect/pull/166) and not have failing Java(doc) tests etc.

src/parsing/parser.rs
keith-hall: agreed, renamed :+1:

src/parsing/parser.rs
keith-hall: that makes sense, it's what I originally wanted to achieve; updated - thanks.

keith-hall:
Thanks for that @robinst, it worked perfectly ❤️ 
I have made the improvements you suggested and also fixed a tiny merge conflict :)

robinst:
It's good to finally have these bugs fixed :)!

src/parsing/parser.rs
```diff
@@ -617,29 +619,23 @@ impl ParseState {
             MatchOperation::None => return false,
         };
         for (i, r) in ctx_refs.iter().enumerate() {
-            let proto_id = if i == ctx_refs.len() - 1 {
-                // a `with_prototype` stays active when the context is `set`
-                // until the context layer in the stack (where the `with_prototype`
-                // was initially applied) is popped off.
+            let mut proto_ids = old_proto_ids.clone().unwrap_or_else(|| Vec::new());
```
trishume: Suppose a `set` has multiple context refs, this appears to add the old prototypes to each new pushed frame, whereas the old behaviour was to only add the old prototypes to the last pushed frame (see the old `else { None }` block).

This seems unintended, but maybe it was intended.

keith-hall: nice spot, it should in fact keep the old prototypes on the *first* pushed frame, there is no need to duplicate it on the other frames as it will be applied from the first pushed frame anyway. I've fixed this and added a test which verifies that it matches ST's behavior.

robinst:
So after we merge this, we can:

* Update the packs
* Release a new version 🎉

robinst:
@keith-hall Do you wanna do the honors of raising a PR to update the packs?

PR #221, opened 2018-10-18
Don't set prototype for prototype contexts included via direct reference

### LINKED ISSUE #219: Hangs forever on simple YAML file when re-building SyntaxSet
#### How to reproduce?

From the `syntect` root folder, perform these steps:

1. Create a simple YAML file:
   ```bash
   echo "# test" > test.yml
   ```
2. Run `syncat` on this file. Everything works as expected:
   ```
   ▶ cargo run --example=syncat -- test.yml
       Finished dev [unoptimized + debuginfo] target(s) in 0.08s
        Running `target/debug/examples/syncat test.yml`
   # test
   ```
3. Create an empty folder
   ```
   mkdir empty
   ```
4. Run `syncat` again with `--extra-sntaxes` set to the empty folder
   ```
   ▶ cargo run --example=syncat -- test.yml --extra-syntaxes empty
       Finished dev [unoptimized + debuginfo] target(s) in 0.08s
        Running `target/debug/examples/syncat test.yml --extra-syntaxes empty`
   ```
   This seems to hang forever.

#### Additional remarks

* This bug only appears if the YAML file contains a line that begins with a comment
* I could not reproduce this behavior with another syntax.

robinst:
After a SyntaxSet is built once, all the references are replaced by
direct references (via index).

When it's converted back to a builder and then built again, the code
that marks prototypes needs to handle those references too, otherwise
some contexts include the prototype when they should not.

Fixes #219.

### DIFF
--- a/src/parsing/syntax_set.rs
+++ b/src/parsing/syntax_set.rs
@@ -427,9 +427,10 @@ impl SyntaxSetBuilder {
             }
 
             for context_id in syntax.contexts.values() {
-                let mut context = &mut all_contexts[context_id.index()];
+                let index = context_id.index();
+                let mut context = &mut all_contexts[index];
                 if let Some(prototype_id) = prototype {
-                    if context.meta_include_prototype && !no_prototype.contains(&context_id.index()) {
+                    if context.meta_include_prototype && !no_prototype.contains(&index) {
                         context.prototype = Some(*prototype_id);
                     }
                 }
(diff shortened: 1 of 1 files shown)


Keats:
@trishume any chance you can publish a 3.0.2 with the current master? Some nice bugfixes in there :)

trishume:
@Keats done, I published `v3.0.2`

Keats:
Thanks!

Issue #183, opened 2018-06-29
Metadata support tracking & discussion


cmyr:
### overview

My initial goal is going to be to support calculating per-line indentation changes (i.e. whether or not we should increase or decrease the indentation level after a given line). I want to do this properly, as discussed in #179: full scope selectors, matching against a line's trailing scope. Approximate TODOs:

- [ ] use full scope selectors for matching
- [ ] add metadata dumping/loading
- [ ] everything behind a feature flag
- [ ] unit tests for a good range of cases
- [ ] a simple API suitable for use from a text editor

Subsequent work:

- comment metadata, comment toggling
- real full-file reindent, with a CLI file reindenter?

maybe one day (not currently planned)
- Symbol lists
- other shell variables?
-  snippets

_I'm very receptive to input/feedback if people think various of these things are more or less important_

### Some Qs:

1. I'm going to need to add the 'Default' sublime package in order for this to work correctly. Are there any possible issues with this?

2. Support for user settings: I'm thinking specifically of `indent_to_bracket`. A text editor using the library would probably prefer if it could hand over this setting to us, and let us handle the implementation; this would be easier than, say, having a separate check for a trailing bracket and clobbering our indentation. Are there any other settings like this?

3. I think we'll need to support `indentParens` and `indentSquareBrackets`. Is there an argument against doing this? I guess one could argue that it should be a user setting as well. We could perhaps support it being both...

4. I've been thinking about testing a bit, and I'm getting hung up on one particular problem. The following two bits of python code are both valid indentations:

```python
# one
if feeling == "happiness":
    do_a_dance()
get_on_with_life()

# two
if feeling == "happiness":
    do_a_dance()
    get_on_with_life()
```

Is this handled in any way by the regex-based system, or is there some other mechanism? I took a quick look and none of the regexes seemed to be looking for whitespace; in any case to detect an indentation decrease we would need to be matching at least two lines, I think?

5. Do I need to manually adjust the regexes to include the _implicit_ `^` anchor?

That's all that I can remember, I'm sure more things will come up.

cc @trishume @keith-hall

keith-hall:
regarding 4, Python is an indentation sensitive language, so the `reindent` functionality should never be used for it. When working with auto indentation, the user should backspace a level of indentation as and when required. The TextMate indentation system doesn't cater for decreasing indentation after a particular line i.e. `pass`, one can only get it to unindent `pass` itself, which is no use.

[...]

PR #223, opened 2018-11-02
Add support for loading metadata files

cmyr:
This adds initial support for loading metadata from `.tmPreferences`
files. The main focus of this patch is loading the patterns used to
determine indentation. I expect to follow up with support for comment
strings, with support for other features possible in the future.

This is all gated behind a new `metadata` feature. If enabled, a
top-level `Metadata` object will be included in `SyntaxSet`; this object
can be queried for the metadata available for a given scope.

This patch does not attempt to implement any actual indentation logic;
that is left up to the consumer.

---------

So I've added a `DefaultPackage` package to `testdata`; it's just two files, but maybe we should just put it in the repo with the other packages?

I've generally tried to keep the scope of this as narrow as possible, no bells & whistles; happy to iterate from here.

I've implemented _simple_ autoindent (just using `increaseIndentPattern` and `decreaseIndentPattern`) in xi using this, and I'm happy with the results.

Some of the logic around loading is a bit gnarly right now; I was doing a bunch of logging to figure out what all metadata fields existed, etcetera. So let's call this a rough cut?


src/parsing/metadata.rs
```diff
@@ -0,0 +1,417 @@
...
+impl Metadata {
+    fn metadata_matching_scope(&self, scope_path: &[Scope]) -> Vec<(MatchPower, &MetadataSet)> {
+        let mut metadata_matches = self.scoped_metadata
+            .iter()
+            .filter_map(|meta_set| {
+                meta_set.selector.does_match(scope_path)
+                    .map(|score| (score, meta_set))
+            }).collect::<Vec<_>>();
+
+        metadata_matches.sort_by(|one, two| two.0.cmp(&one.0));
```
trishume: Hint: `sort_by_key` makes this cleaner

src/parsing/metadata.rs
```diff
@@ -0,0 +1,417 @@
...
+    pub fn add_raw(&mut self, raw: RawMetadataEntry) {
+        let RawMetadataEntry { scope, settings } = raw;
+        let scoped_settings = self.loaded.entry(scope.clone())
+            .or_insert_with(|| Dict::new());
+
+        for (key, value) in settings {
+            if !KEYS_WE_USE.contains(&key.as_str()) {
+                continue;
+            }
+            //NOTE: because we can't guarantee the order that files get loaded,
```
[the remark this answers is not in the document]
cmyr: Comment is out of date and I agree just doing alphabetical order on the files is much simpler, will make that change.

cmyr:
@trishume thanks for the quick look. I agree that the initial loading/conversion feels a bit over-complicated, but couldn't think of an easier approach that correctly handles adding new files.

I'll do some cleanup on this a bit later on today.

As an addendum: I'd like to add support for the comment-related members of `shellVariables`, which I could either add on here or do as a follow-up PR. In any case, I see three basic approaches to this:

- expose a `shellVariables` that includes all the keys in the metadata files, and just leave it entirely up to the consumer
- extract the comment related metadata items, and have, say,
    ```rust
    struct MetadataItems {
        // ...
        comment_one: Option<(String, Option<String>)>,
        comment_two: Option<(String, Option<String>)>,
        comment_three: Option<(String, Option<String>)>,
    }
    ```
- or just go ahead and use the interface we expect the client to want, which is more like,
    ```rust
    struct MetadataItems {
        // ...
        line_comment: Option<String>,
        block_comment: Option<(String, String)>,
    }
    ```

This latter is the cleanest and the easiest to use but requires us to make some assumptions and to throw away some metadata. It feels like the best approach for now, but if you have an opinion I'm happy to hear it.

keith-hall:
I personally vote for:

> expose a `shellVariables` that includes all the keys in the metadata files, and just leave it entirely up to the consumer

as it is the most flexible/useful :)

cmyr:
Okay I've updated this to simplify the merging logic and address some of the other feedback.

I've noticed a probable bug here, as well: we go to pretty major effort to keep around the 'raw' metadata, so that we can correctly merge if new files are added; however we don't include the raw metadata when we do `gendata`, so adding new metadata to a syntax set that was loaded from a dump won't work. I'll resolve this in an upcoming commit.

cmyr:
I've gone ahead and added a commit that makes the change described above. Let me know if you prefer a different approach.

src/parsing/metadata.rs
```diff
@@ -0,0 +1,447 @@
...
+    pub fn disable_indent_next_line(&self, line: &str) -> bool {
+        self.best_match(|ind| ind.disable_indent_next_line_pattern.as_ref().map(|p| p.is_match(line)))
+            .unwrap_or(false)
+    }
+
+    fn best_match<T, F>(&self, f: F) -> Option<T>
+        where F: FnMut(&MetadataItems) -> Option<T>
+    {
+        self.items.iter()
+            .map(|(_, meta_set)| &meta_set.items)
```
trishume: Do you know that it's the case that Sublime does this trying multiple patterns thing? Seems rarely applicable and I wouldn't be surprised if Sublime only tries the matchiest, which would let you simplify a bunch of this.

cmyr: I believe sublime does it this way; this is why we need the new default package, because it includes some base settings for the `source` and `comment` scopes. @keith-hall suggested this implementation in #183 or #179.

examples/gendata.rs
```diff
@@ -2,26 +2,40 @@
 //! syntect, not as a helpful example for beginners.
 //! Although it is a valid example for serializing syntaxes, you probably won't need
 //! to do this yourself unless you want to cache your own compiled grammars.
+//!
+//! The standard command to generate the syntax dumps (including metadata) is:
```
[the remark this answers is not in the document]
cmyr: Done.

src/highlighting/mod.rs
```diff
@@ -3,7 +3,7 @@
 //! settings like selection color, `ThemeSet` for loading themes,
 //! as well as things starting with `Highlight` for how to highlight text.
 mod selector;
-mod settings;
+pub(crate) mod settings;
```
trishume: This was previously public and even though I think literally nobody uses it I'd rather not break semver compatibility for little reason.

src/parsing/metadata.rs
```diff
@@ -0,0 +1,447 @@
...
+    /// Adds the provided `RawMetadataEntry`. When creating the final `Metadata`
+    /// object, all `RawMetadataEntry` items are sorted by path, and items that
+    /// share a scope selector are merged; last writer wins.
+    pub fn add_raw(&mut self, raw: RawMetadataEntry) {
+        self.loaded.push(raw);
+    }
+
+    /// Generates a `MetadataSet` from a single file
+    #[cfg(test)]
+    pub fn quick_load(path: &str) -> Result<MetadataSet, LoadingError> {
```
trishume:
```suggestion
    pub(crate) fn quick_load(path: &str) -> Result<MetadataSet, LoadingError> {
```

src/highlighting/mod.rs
cmyr: this was previously private, I exposed it because I needed `StackSelectors`.

src/parsing/metadata.rs
cmyr: Done.

src/highlighting/mod.rs
cmyr: er, not `StackSelectors` but `Settings` and `SettingsObject`? because I'm loading plists.

trishume: oops derp cool

cmyr: looked at this again, it's because I'm using `load_plist`. Could reexport manually if that makes more sense?

cmyr:
Okay, should be good to go?

cmyr:
Thanks for taking the time out of your Sunday to get it merged, much appreciated.

PR #230, opened 2019-01-21
Setter funcion to update private attribute `metadata`

kanishkarj:
This change is necessary to load custom syntaxes in xi-syntect-plugin.

### DIFF
--- a/src/parsing/syntax_set.rs
+++ b/src/parsing/syntax_set.rs
@@ -144,6 +144,10 @@ impl SyntaxSet {
         &self.syntaxes[..]
     }
 
+    pub fn set_metadata(&self, metadata: Metadata) {
+        self.metadata = metadata;
+    }
+    
     /// The loaded metadata for this set.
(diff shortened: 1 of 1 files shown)


src/parsing/syntax_set.rs
```diff
@@ -144,6 +144,10 @@ impl SyntaxSet {
         &self.syntaxes[..]
     }
 
+    pub fn set_metadata(&mut self, metadata: Metadata) {
```
cmyr: this needs to be guarded by a cfg feature, like below.

kanishkarj: Shall do it.

kanishkarj: #[cfg(feature = "metadata")]
will this do?

kanishkarj: Check this out!

trishume:
Cool looks good to me.

PR #235, opened 2019-02-17
Added classed HTML generator struct

terhechte:
This abstracts the necessary logic away that is required to
generated HTML tags with class names instead of colored
style tags.

It also fixes a bug where `tokens_to_classed_html` previously
did not add the remainder of a parsed line to the output.


terhechte:
Thanks! I like the deprecation of the old function in your refactoring. I'm currently using syntect on a fork anyway because I need more built-in languages in the assets (particularly Swift), so I don't mind the git dependency :)

trishume:
In case you didn't know and you want to, you can use your own built in language bundle without using a fork. The functions you need to generate and load your own bundle are public. Then you just use the cargo features to disable the default built-in bundle, then include your own in the binary the same way syntect does (check out the source).

PR #236, opened 2019-02-23
chore: Update pretty_assertions to 0.6

ignatenkobrain:


trishume:
I'm guessing if you want to package a version with updated dependencies you might want me to do a release at some point? If so just let me know and I can push one out.

ignatenkobrain:
It would be much appreciated :)

trishume:
I just published v3.1.0 with both the version bumps

ignatenkobrain:
Thank you!

PR #243, opened 2019-03-15
Make `ScopeRepository::new` public

brupelo:
I'm in the middle of porting all unit tests from syntect to python... this would help, while I can pin the syntect dependency locally I'd prefer to have it pinned to the main upstream

### DIFF
--- a/src/parsing/scope.rs
+++ b/src/parsing/scope.rs
@@ -125,7 +125,7 @@ fn pack_as_u16s(atoms: &[usize]) -> Result<Scope, ParseScopeError> {
 }
 
 impl ScopeRepository {
-    fn new() -> ScopeRepository {
+    pub fn new() -> ScopeRepository {
         ScopeRepository {
             atoms: Vec::new(),
             atom_index_map: HashMap::new(),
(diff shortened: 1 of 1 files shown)


trishume:
Sorry not going to do this. There is no legitimate reason to create a new `ScopeRepository`. Replicating unit tests of internal interfaces is not a good idea, just unit test the things that the binding needs to expose.

brupelo:
Sure, no worries, I'll use my own fork to suit my needs from now on. 

About not being a good idea, maybe you're right but at this point all the unit tests living in syntect are really helpful on my rust learning path though :). Right now I'm in the middle of porting all public elements to python and trying to keep the original interface... that of course won't be ideal as the API won't end up being pythonic, so having all unit tests in place will help me to eventually make a nicer public python interface. That or maybe I'll just give up on syntect and maybe end up using [https://github.com/tree-sitter/tree-sitter](https://github.com/tree-sitter/tree-sitter) antlr or some other solution, I'm not sure yet :/ , I'm just messing around.

In the past 2-3 few weeks I've been exploring/researching quite a lot of github projects to explore what are my options in order to have a ST-like text editor to replace my existing Scintilla widgets in some of my custom projects and so far I haven't found anything that satisfies me...  You can read more about it [here](https://github.com/edbee/edbee-lib/issues/82#issuecomment-472350452).

I'm not sure yet the direction to take and I'm still messing around but probably I'll end up rolling up my own text editor widget from scratch eventually as I've learned quite a lot thanks to this research stage, in the meantime this binding project is just a "hello world" project that is helping me to learn more about rust and python wrapping using pyo3.

Issue #222, opened 2018-10-30
Embed rewriting not perfect


keith-hall:
After the latest fixes, I tried updating the submodules, and can see we still have some syntax test failures, this time relating to `embed`...`escape`:

```
FAILED testdata/Packages\Java\syntax_test_java.java: 817
FAILED testdata/Packages\Markdown\syntax_test_markdown.md: 61
```

We get the same failures in Sublime Text, when rewriting the `embed`...`escape` rules to `with_prototype` like `syntect`'s `yaml_load` is doing, so the good news is that `syntect`s parser is working correctly:

`JavaDoc.sublime-syntax`:
```diff
--- Shipped Packages/Java/JavaDoc.sublime-syntax    2018-10-11 19:11:54
+++ Packages/Java/JavaDoc.sublime-syntax    2018-10-30 11:28:20
@@ -20,11 +20,12 @@
     - meta_include_prototype: false
     - match: /\*\*
       scope: comment.block.documentation.javadoc punctuation.definition.comment.begin.javadoc
-      embed: contents
-      embed_scope: comment.block.documentation.javadoc text.html.javadoc
-      escape: \*/
-      escape_captures:
-        0: comment.block.documentation.javadoc punctuation.definition.comment.end.javadoc
+      push:
+        - [{ meta_include_prototype: false }, { meta_content_scope: 'comment.block.documentation.javadoc text.html.javadoc' }, { match: '\*/', scope: comment.block.documentation.javadoc punctuation.definition.comment.end.javadoc, pop: true }]
+        - contents
+      with_prototype:
+        - match: (?=\*/)
+          pop: true
```

`Markdown.sublime-syntax`:
```diff
--- Shipped Packages/Markdown/Markdown.sublime-syntax   2018-10-11 19:11:54
+++ Packages/Markdown/Markdown.sublime-syntax   2018-10-30 09:44:46
@@ -1106,12 +1106,12 @@ contexts:
         0: meta.code-fence.definition.begin.html-php.markdown-gfm
         2: punctuation.definition.raw.code-fence.begin.markdown
         5: constant.other.language-name.markdown
-      embed: scope:embedding.php
-      embed_scope: markup.raw.code-fence.html-php.markdown-gfm
-      escape: '{{code_fence_escape}}'
-      escape_captures:
-        0: meta.code-fence.definition.end.html-php.markdown-gfm
-        1: punctuation.definition.raw.code-fence.end.markdown
+      push:
[...]

PR #246, opened 2019-03-24
Update packages

robinst:
Needs a newer yaml-rust to get this fix: https://github.com/chyh1990/yaml-rust/pull/122

There are currently test failures and more syntests are failing than before. Needs to be investigated.


robinst:
syntest for Java fails to parse Javadoc comments. It looks like it's related to embed rewriting, so we'll probably have to fix #222.

Enselic:
I'm doing a bit of a PR inbox cleanup, and the `yaml-rust` bump of this PR has already been made elsewhere, and the other bumps are very likely to be outdated by now, so let's close this PR.

Of course anyone is free to re-open this PR if you disagree.

PR #254, opened 2019-07-24
Extract some methods from parse_match_pattern

robinst:
Makes it a bit less long and enables reuse later.

### DIFF
--- a/src/parsing/yaml_load.rs
+++ b/src/parsing/yaml_load.rs
@@ -292,69 +292,22 @@ impl SyntaxDefinition {
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/parsing/yaml_load.rs
```diff
@@ -292,69 +292,22 @@ impl SyntaxDefinition {
         }
     }
 
-    fn resolve_variables(raw_regex: &str, state: &ParserState) -> String {
```
robinst: These two were just moved further down to make it a bit more consistent (the reading order of methods in the file is top down).

PR #255, opened 2019-08-08
Update dependencies and fix deprecated

botika:


Keats:
Note that upgrading onig to 5.0 means building syntect now requires clang.
I'll need to stick with `=3.2.0` for Zola as Azure Pipelines doesn't have LLVM built-in

trishume:
Damn that's unfortunate, maybe file an issue with onig and see what they say about it? Maybe it's an easy fix.

robinst:
Might get fixed with https://github.com/rust-onig/rust-onig/pull/113

PR #259, opened 2019-09-03
fix panic when highlighter cache is empty

keith-hall:
this fixes a panic which occurs when highlighting a JSON file using a recent version of `JSON.sublime-syntax` from the sublimehq/Packages repo. This bug was originally discovered at https://github.com/sharkdp/bat/pull/644 - the steps for reproducing the panic are there.

Note, I haven't added a unit test for this yet, and am not sure when I will get chance to work on one... If someone else wants to take it on, please do.


sharkdp:
Awesome, thank you for taking a look at this, @keith-hall!

sharkdp:
It would be fantastic to get a patch release with this bugfix, as it prevents us from updating `sublimhq/Packages` in `bat` (https://github.com/sharkdp/bat/pull/644).

(only if you find the time, I definitely don't want to bother you with downstream problems :smiley:)

trishume:
@sharkdp Done, released as v3.3.0

sharkdp:
@trishume Thank you so much!

PR #262, opened 2019-09-14
add initial latex output functionality

qtfkwk:
I'm relatively new to Syntect but it looks great! I'm working on a project using [Tectonic](https://tectonic-typesetting.github.io/) ([crates.io: tectonic](https://crates.io/crates/tectonic)) and my motivation is to use Syntect to highlight source code blocks in LaTeX. Unfortunately, the current set of output formats appears to be html and terminal. This PR adds initial LaTeX output functionality akin to the `as_24_bit_terminal_escaped` function.

### DIFF
--- a/examples/latex-demo.rs
+++ b/examples/latex-demo.rs
@@ -7,7 +7,7 @@ fn main() {
     // Load these once at the start of your program
     let ps = SyntaxSet::load_defaults_newlines();
     let ts = ThemeSet::load_defaults();
-    
+
     let syntax = ps.find_syntax_by_extension("rs").unwrap();
     let s = "pub struct Wow { hi: u64 }\nfn blah() -> u64 {}\n";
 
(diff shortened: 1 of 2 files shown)


trishume:
Also your doctest doesn't pass CI, I won't merge unless that's fixed.

qtfkwk:
Doctest is passing now... at least `cargo test --doc` is clean.

Regarding your advice to avoid a heap allocation... I'm also relatively new to Rust, but I believe you're referencing the `replacements` `Vec<(&str, &str)>` and want it as a global? That seems like the top candidate to me. AFAICT, the `content` string must be heap allocated to avoid changing the referenced str. I guess alternatively, I could also/rather perhaps attempt to modify the `prev_style` to an `Option<&Style>` and possibly avoid the `.clone()`? Any additional details/thoughts would be appreciated.

qtfkwk:
Works now following this update to Cargo.toml:

```.toml
#syntect = { git = "https://github.com/qtfkwk/syntect", branch = "latex" }
syntect = { git = "https://github.com/trishume/syntect" }
```

Much appreciated!

qtfkwk:
Works now following this update to Cargo.toml:

```.toml
#syntect = { git = "https://github.com/qtfkwk/syntect", branch = "latex" }
#syntect = { git = "https://github.com/trishume/syntect" }
syntect = "3.3.0"
```

Much appreciated!

Issue #261, opened 2019-09-10
HighlightState with a non-empty initial_stack crashing?


yonilevy:
Hi,

I'm trying to build a `HighlightState` with a cached `initial_stack`, as described in the docs:

> Alternatively you can save space by caching only the `path` field of this struct
> then re-create the `HighlightState` when needed by passing that stack as the `initial_stack`
> parameter to the `new` method. This takes less space but a small amount of time to re-create the style stack.

I'm probably missing something (very new to rust), but to me it appears this is guaranteed to crash:
The first call to `update_single_cache_for_push` in `HighlightState::new`, will pass `path` of length 0, which results in `path.len() - 1` panicking with `attempt to subtract with overflow`.

Is there an example somewhere of code using HighlightState with a non-empty initial_stack? What am I missing?

Thanks ahead, and many thanks for building this incredible library!

keith-hall:
Hi, thanks for reporting. It's possible that there are currently no unit tests covering this scenario, and that you are the first person to try using it... 
I will investigate when I'm able to make time, but maybe someone else will beat me to it :)

keith-hall:
Looks like this was fixed in #263

PR #263, opened 2019-09-14
bugfix for HighlightState with non-empty initial stack

yonilevy:
I believe this fixes issue #261, though I don't feel confident enough with the code base to assure this is the exact intended behaviour, so definitely needs a review from someone who knows their way around this repo :)

### DIFF
--- a/src/highlighting/highlighter.rs
+++ b/src/highlighting/highlighter.rs
@@ -86,7 +86,7 @@ impl HighlightState {
         let mut styles = vec![highlighter.get_default()];
         let mut single_caches = vec![ScoredStyle::from_style(styles[0])];
         for i in 0..initial_stack.len() {
-            let prefix = initial_stack.bottom_n(i);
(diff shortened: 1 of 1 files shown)


yonilevy:
Thanks for reviewing @keith-hall ❤️

My goal with this test was to create a scenario that would break prior to the change, and produce the expected result after. I totally get (the idea of) the test not actually replicating the most-useful way of using the feature, and not covering all the execution paths, but also not sure how I would test that(?)

At any rate as you said it's an improvement to crashing, so perhaps you could merge it and we would continue the cache testing discussion independently?

keith-hall:
I am amenable to that, but although I have the technical permissions to merge a PR, this is @trishume's repo and ultimately the decision should be his :)
I like the idea of separately adding an extra test in addition to the one you have already created, which could cover other execution paths :+1: This is something I should be able to make time for within the next week

Issue #268, opened 2019-11-19
Name options for CSS classes


uwearzt:
Ate the moment generated CSS classes always contain spaces, which are not handled nicely by CSS. Create a different ```ClassStyle``` for better CSS class handling.

Example old:
```
<span class="keyword control import include c++">
```
Example new:
```
<span class="keyword-control-import-include-c++">
```

PR #269, opened 2019-11-19
Add ClassStyle::Dashed for nice CSS class names

uwearzt:
Implements https://github.com/trishume/syntect/issues/268

Thanks for syntect, great piece of software :)

### DIFF
--- a/examples/synhtml-css-classes.rs
+++ b/examples/synhtml-css-classes.rs
@@ -1,0 +1,38 @@
+//! Prints highlighted HTML with CSS classes for a Rust and a C++ file to stdout.
+//! Run with ```cargo run --example synhtml-css-classes```
+use syntect::parsing::SyntaxSet;
+use syntect::html::ClassedHTMLGenerator;
+use syntect::html::ClassStyle;
+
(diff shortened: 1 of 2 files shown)


trishume:
I'm not sure I want to merge this absent a good pitch on why it's useful. It currently looks like not something you actually want. For example with dashed classes you can only ever write CSS selectors that work for one programming language at a time.

With spaced classes every element has many classes and matching keywords is as easy as `.keyword` and PHP keywords is `.keyword.php`. 

I acknowledge that you lose the ability to select on the order of tokens within a Sublime Text scope, but I don't think the solution to that is dashes because you just lose your ability to select on everything except with maximum specificity. If you have a use case for a meaningful CSS selector you want to write that doesn't work with the spaced style, I'm interested, and maybe we can come up with a better class generation method that works for that use case without requiring maximum specificity.

uwearzt:
Hmm, you are right, after playing around with it a little bit more, maximum specificity doesn't look like the correct solution.

Are you interested in my example program for generating html with class styles? I would send you a separate pull request.

PR #270, opened 2019-11-25
Move all regex usage to separate module to add support for fancy-regex

robinst:
This has the same goal as #34 but with a different approach.

Note that the fancy-regex implementation doesn't compile yet, but I thought it would be useful to get this reviewed earlier rather than later.

I haven't ported over the regex rewriting changes yet, I'm hoping that we can generate regexes that work on both onig and fancy-regex.

* [x] Add std::error::Error impl for fancy-regex, see https://github.com/fancy-regex/fancy-regex/pull/35
* [x] Release https://github.com/fancy-regex/fancy-regex/pull/33
* [x] Port regex rewriting changes


Cargo.toml
```diff
@@ -55,15 +56,19 @@ dump-create = ["flate2/default", "bincode"]
...
 metadata = ["parsing"]
 # The `assets` feature enables inclusion of the default theme and syntax packages.
 # For `assets` to do anything, it requires one of `dump-load-rs` or `dump-load` to be set.
 assets = []
 html = ["parsing", "assets"]
 yaml-load = ["yaml-rust", "parsing"]
-default = ["parsing", "assets", "html", "yaml-load", "dump-load", "dump-create"]
+default-onig = ["parsing", "assets", "html", "yaml-load", "dump-load", "dump-create", "regex-onig"]
+default-fancy = ["parsing", "assets", "html", "yaml-load", "dump-load", "dump-create", "regex-fancy"]
+default = ["default-onig"]
```
robinst: Not sure what the best way to structure the features is, any thoughts?

src/parsing/parser.rs
```diff
@@ -327,7 +323,7 @@ impl ParseState {
                 let match_pat = pat_context.match_at(pat_index);
 
                 if let Some(match_region) = self.search(
-                    line, start, match_pat, captures, search_cache, regions
+                    line, start, match_pat, captures, search_cache
```
robinst: Note that the previous code reused the regions. I should benchmark what the impact of this is, but I decided for the straightforward API for now because I didn't want to complicate things more.

src/parsing/regex.rs
```diff
@@ -0,0 +1,254 @@
...
+/// * Lazily compiles regexes on first use to improve initialization time.
+#[derive(Debug)]
+pub struct Regex {
+    regex_str: String,
+    regex: AtomicLazyCell<regex_impl::Regex>,
+}
+
+/// A region contains text positions for capture groups in a match result.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct Region {
```
robinst: So these types and their methods become public API. Not sure about the naming, it's currently similar to what onig uses, but different from the regex crate/fancy-regex.

src/parsing/syntax_definition.rs
```diff
@@ -86,17 +85,14 @@ pub struct MatchIter<'a> {
     index_stack: Vec<usize>,
 }
 
-#[derive(Debug, Serialize, Deserialize)]
+#[derive(Clone, Debug, Eq, PartialEq, Serialize, Deserialize)]
 pub struct MatchPattern {
     pub has_captures: bool,
-    pub regex_str: String,
+    pub regex: Regex,
```
robinst: Note due to how the serialization/deserialization of `Regex` just delegates to the `String` inside it, this happens to work without changing the binary format for packs (i.e. no need to regenerate packs).

src/parsing/yaml_load.rs
```diff
@@ -424,25 +424,35 @@ impl SyntaxDefinition {
     }
 
     fn resolve_variables(raw_regex: &str, state: &ParserState<'_>) -> String {
-        state.variable_regex.replace_all(raw_regex, |caps: &Captures<'_>| {
```
robinst: This was the only use of `replace_all`. Putting that into the regex API would have been a bit complicated, so I rewrote this part to use `search` instead.

src/parsing/parser.rs
trishume: I think this is plausibly an actually important optimization, perhaps especially with onig but allocating an extra Vec on every search with fancy-regex isn't great either. But I forget since it was a long time ago that I optimized this.

I would be happy if you ran the benchmarks with onig in this PR vs the ones in master so we can see if this makes a difference that matters. If so we can just refactor the `regex` module interfaces to get a region passed in, then the fancy-regex implementation can `clear` the `Vec` before adding to it again.

Cargo.toml
trishume: The key thing about features is that they're additive so that if multiple crates specify different features, the union of them works for both crates. The other nice thing to do is to preserve compatibility with existing crates that depend on us without default features, although I'm okay with breaking that if there's no good way otherwise.

I can't see a clean way of making the features backwards-compatible, so maybe just change the `cfg` statements in the `regex` module so that if both `regex-fancy` and `regex-onig` are set then `regex-fancy` takes precedence, although I could see the precedence going the other way too, as long as it works with both set.

trishume:
Also I must say I love how this feature removes nearly as many lines as it adds, largely due to the regex abstraction removing a lot of duplication. Good work on that.

robinst:
Ok, so there's a problem with the Java syntax with fancy-regex, which I've narrowed down to this bug that will need to be fixed: https://github.com/fancy-regex/fancy-regex/issues/37

gilescope:
What can we do to help? Really keen on being able to land this.

trishume:
One avenue that might be an easier path towards fixing common Windows woes is helping with https://github.com/rust-onig/rust-onig/pull/126 which should make onig no longer require Clang on Windows, although it may still need to build C I'm not sure. You could maybe make it so that the `onig` crate uses a binary Windows build of the `oniguruma` library.

Alternatively you can try checking out this branch and seeing if you can rebase it and get it passing CI, which may require updating fancy-regex and maybe some more fixes. I think @robinst may have mentioned at some point that he was still occasionally tinkering with this but not sure if that's still true.

gilescope:
(Am trying to bring this PR up to date with PR 7

gilescope:
Should the feature be called regex-rs rather than regex-fancy to match with dump-create-rs and dump-load-rs?

gilescope:
Confirmed this branch builds fine on windows 10 and OSX using ```cargo build --features regex-fancy --no-default-features``` with no onix installed. Maybe we could add a line to the readme to explicitly say if you want to build a pure rust version this is how to do it. Aside from that I'm *very* happy with this PR.

gilescope:
The sooner we land it, the sooner people can say `cargo install cargo-expand` and it will `just work` (tm). This will be really great for a lot of rust devs!

Adarma:
This works fine for me on windows with the following `Cargo.toml`:
```
syntect = { git = "https://github.com/trishume/syntect", branch = "move-regex-use-to-module", default-features = false, features = ["default-fancy"] }
```

gilescope:
Wow, that's some test case.

robinst:
Yay, thanks for helping push this over the finish line :)!

The benchmarks match the shallow benchmarking that I've done. I've done some matching of single regexes (ones that don't need fancy features) and I've observed that the first N matches are slow, and only after some warmup do they get fast. It would be good to investigate this more to see if we can get more performance out of it.

For the test failure, I actually already looked into that and fixed it in the regex rewriting. All that was needed was updating the packs (which I put off until the end because it's hard to rebase)! Pushed that now and build should go green.

If you don't mind adding the bits to the readme @trishume, that would be awesome.

BurntSushi:
@robinst If you're able to widdle down a simple benchmark for me, I'd be happy to take a closer look.

trishume:
This is now released as [v4.0.0](https://github.com/trishume/syntect/releases/tag/v4.0.0)! Thanks again everyone for your work, especially @robinst!

Issue #272, opened 2019-12-02
Create example for ClassedHTMLGenerator


uwearzt:
So everyone can see how it is used.

uwearzt:
I will create a push request in the next days.

uwearzt:
branch example-classedhtmlgenerator

PR #274, opened 2019-12-02
Add example for ClassedHTMLGenerator

uwearzt:
fixes https://github.com/trishume/syntect/issues/272
fixes https://github.com/trishume/syntect/issues/275


jrappen:
@uwearzt use `fixes #222`

PR #276, opened 2019-12-26
Empty css spans

### LINKED ISSUE #273: ClassedHTMLGenerator generates Empty <span></span> tags.
See output of ClassedHTMLGenerator example.

uwearzt:
fixes #273

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,6 @@ Cargo.lock
 
 # VS Code
 .vscode/
+
+# macOS
+.DS_Store
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,13 +20,13 @@ exclude = [
 [dependencies]
 yaml-rust = { version = "0.4", optional = true }
 onig = { version = "5.0", optional = true }
-walkdir = "2.0"
+walkdir = "2.2.9"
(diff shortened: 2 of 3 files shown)


trishume:
I'm sorry about this, but I don't think I'm willing to accept this PR without it using a different solution. This PR adds a dependency on `regex` when there are already probably too many dependencies and a different regex engine (`onig`). It's also a pretty inefficient way to do it, not that this code is likely to be too performance sensitive. It's important to me that syntect be as lightweight and fast as I can make it, and in this case adding an extra `regex` dependency also burdens everyone, even those who don't use this feature.

A better way to do it might be to track the current length of `s` in `tokens_to_classed_spans` in a local variable before opening a span, and another flag variable for if any text was added. Then at the code that closes a span, check if any text was added to the span and if not then instead of closing, truncate the string to the stored length before the previous span was opened. If you want to get really fancy you could even have it detect multiple nested spans with no text in them and remove all the levels, which might require a stack, I don't think this is worth it though.

uwearzt:
@trishume You're right, regex might be a huge overkill for that purpose... I have reimplemented the feature according to your suggestion. Please have a look.

Thanks in advance
Uwe

PR #277, opened 2020-01-16
Upgrade to criterion 0.3 and plist 0.5

cuviper:
No code changes needed.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ regex-syntax = { version = "0.6", optional = true }
 lazy_static = "1.0"
 lazycell = "1.0"
 bitflags = "1.0"
-plist = "0.4"
+plist = "0.5"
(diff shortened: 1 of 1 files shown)


cuviper:
Hmm, I just saw that these were once updated, then reverted in commit d133329f47a6. It's not clear to me if you objected to those upgrades though, as your commit message only seems to be about compatibility with a semantic version. It's your call, of course.

trishume:
Thanks these are both probably good upgrades to do. 

The reason I reverted the modifications in the other commit is that it also bumped the minor versions of a bunch of dependencies that didn't need upgrading. Which makes dependency resolution more constrained for no reason.

These upgrades on the other hand are major version bumps in semver, so it's good to be on the latest version, and the tests passed so it doesn't seem like they broke anything hopefully. I don't think CI checks the benchmarks but I tested them locally and they still work.

Issue #271, opened 2019-11-27
sublime-syntax branch_point support


keith-hall:
Sublime Text build 4050 introduced a new feature to `sublime-syntax` grammars called branch points. The docs haven't been published yet, but the packages which ship with Sublime Text have started to use this new functionality. i.e. https://github.com/sublimehq/Packages/commit/c08f85346a04d2e12990c3008e7666306b59a05b

This essentially uses backtracking to effectively allow looking ahead multiple lines (up to 128) etc. and enables non-deterministic parsing as described at https://github.com/SublimeTextIssues/Core/issues/2241

My understanding is that some official documentation should be coming soon, but I wanted to log  this here now in case we want to get a head start on it i.e. by experimenting with it in Sublime to see how it works and thinking about how we can go about implementing it for `syntect`. This also serves as a note that updating the sublimehq/Packages submodule will likely cause those syntax definitions relying on this new feature to not work as expected until `syntect` adds support for it.

jrappen:
For other possible breaking changes, comparing the `st3` branch against `master` might also help:

* https://github.com/sublimehq/Packages/compare/st3...master

jrappen:
It might make sense to point the Packages gitmodule at the st3 branch until this is fixed.

jrappen:
Changed gitsubmodule target for `Packages` to `st3` branch for now in #279. This should be reverted to targetting `master` branch once all Sublime syntax features are supported by `syntect`.

FichteFoll:
Documentation on syntax definitions has been updated to include documentation for branching: https://www.sublimetext.com/docs/syntax.html#ver-dev

The documentation is in a not-so-public state yet, as Sublime Hq prepares to go public with their new dev cycle, but I figured I'd post this here already.

Keats:
Was some documentation added on that? Is it a big feature?

keith-hall:
It's a big feature, the current line by line api isn't suitable, because parsing the next line could affect the tokens for the previous line if there was a branch point that failed and the context stack/parse state rewound. There are some docs, but sublimehq have asked not to link to them publicly during the closed beta - you can find a link from the sublime text discord server.

jrappen:
New docs can be found at:

* <https://www.sublimetext.com/docs/syntax.html#ver-dev>
* <https://www.sublimetext.com/docs/syntax.html#ver-3.2>

michaelblyons:
[...]

PR #154, opened 2018-04-30
Update submodules

Keats:
This triggers a panic in the test `can_parse_includes`: Result::unwrap()` on an `Err` value: Error(-114, target of repeat operator is invalid)

```
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
             at libstd/sys/unix/backtrace/tracing/gcc_s.rs:49
   1: std::sys_common::backtrace::print
             at libstd/sys_common/backtrace.rs:71
             at libstd/sys_common/backtrace.rs:59
   2: std::panicking::default_hook::{{closure}}
             at libstd/panicking.rs:205
   3: std::panicking::default_hook
             at libstd/panicking.rs:215
   4: std::panicking::rust_panic_with_hook
             at libstd/panicking.rs:457
   5: std::panicking::begin_panic_fmt
             at libstd/panicking.rs:344
   6: rust_begin_unwind
             at libstd/panicking.rs:322
   7: core::panicking::panic_fmt
             at libcore/panicking.rs:71
   8: core::result::unwrap_failed
             at /checkout/src/libcore/macros.rs:26
   9: <core::result::Result<T, E>>::unwrap
             at /checkout/src/libcore/result.rs:782
  10: syntect::parsing::syntax_definition::MatchPattern::compile_regex
             at src/parsing/syntax_definition.rs:249
  11: syntect::parsing::syntax_definition::MatchPattern::ensure_compiled_if_possible
             at src/parsing/syntax_definition.rs:261
  12: syntect::parsing::parser::ParseState::search
             at src/parsing/parser.rs:399
  13: syntect::parsing::parser::ParseState::find_best_match
             at src/parsing/parser.rs:332
  14: syntect::parsing::parser::ParseState::parse_next_token
             at src/parsing/parser.rs:242
  15: syntect::parsing::parser::ParseState::parse_line
             at src/parsing/parser.rs:213
  16: syntect::parsing::parser::tests::ops
             at src/parsing/parser.rs:1272
  17: syntect::parsing::parser::tests::can_parse_includes
             at src/parsing/parser.rs:697
  18: syntect::__test::TESTS::{{closure}}
             at src/parsing/parser.rs:690
  19: core::ops::function::FnOnce::call_once
             at /checkout/src/libcore/ops/function.rs:223
  20: <F as alloc::boxed::FnBox<A>>::call_box
             at libtest/lib.rs:1451
             at /checkout/src/libcore/ops/function.rs:223
[...]

PR #279, opened 2020-01-28
Packages submodule, git files, and docs improvements

jrappen:
Update changelog:

- added changelog info from releases to changelog file

Changes to Packages submodule target:

- change submodule target `Packages` temporarily to `st3` branch
  - this should be reverted to targetting master once #271 is implemented
- address #246 
- address #166 (which superseded #154)


trishume:
Oh also I think in order to get the Packages actually updated you have to do some check out step in the submodule. I forget exactly how submodules work but I'm pretty sure this PR doesn't actually update the packages, just makes future updates target the correct branch. Also when really updating the packages you need to rebuild the checked in binary dumps.

Not updating the packages in this PR is fine though. I'm pretty sure based on the other issues that updating the packages might not work.

jrappen:
Can't test the build on this machine. Might try later today.

Pushed the requested changes regarding css names for now.

PR #280, opened 2020-02-05
Fix error with bitflags < 1.0.4

robinst:
Only bitflags 1.0.4 supports imports like `use bitflags::bitflags;`, see https://github.com/bitflags/bitflags/releases. My lock file still had 1.0.3 and so I was getting a compile error.

Bump the minimum version in Cargo.toml.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ walkdir = "2.0"
 regex-syntax = { version = "0.6", optional = true }
 lazy_static = "1.0"
 lazycell = "1.0"
-bitflags = "1.0"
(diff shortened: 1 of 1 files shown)


robinst:
Related to this, what do people think about checking in the `Cargo.lock` file?

trishume:
Thanks! I'm mildly against checking in the Cargo.lock file for exactly the reason that it helps catch errors like this, otherwise they'll only show up for people using syntect as a library in other things and not contributors working on it directly. I think it's also convention for libraries not to check in their Cargo.lock files.

PR #281, opened 2020-02-05
Fix .gitattributes

robinst:
Got warnings from Git like these because of the comments at end of lines:

    # is not a valid attribute name: .gitattributes:4

CC @jrappen


jrappen:
@robinst sorry about that, was under the impression trailing comments work

Issue #264, opened 2019-09-28
Build failed for onig


elgunhuseynli:
Examples aren't compiling on Windows 10.

```
Compiling onig_sys v69.2.0
error: failed to run custom build command for `onig_sys v69.2.0`
...
--- stdout
cargo:warning=couldn't execute `llvm-config --prefix` (error: The system cannot find the file specified. (os error 2))
cargo:warning=set the LLVM_CONFIG_PATH environment variable to a valid `llvm-config` executable
```

elgunhuseynli:
Oniguruma is regex library written in C. Why not just use Rust's own regex library? So we don't get build errors like that.

keith-hall:
https://github.com/rust-onig/rust-onig/issues/109

> Why not just use Rust's own regex library?

see https://github.com/trishume/syntect/issues/12

trishume:
See the new pure-Rust `fancy-regex` mode introduced in [v4.0.0](https://github.com/trishume/syntect/pull/270) to avoid this issue!

Still hoping that onig's build will eventually improve but for now `fancy-regex` is a good option.

Enselic:
Duplicate of #12

Enselic:
This issue can be closed with motivation:

Effectively a duplicate of #12

PR #283, opened 2020-03-19
Move regex use to module: Rebase

gilescope:
Attempt to bring pr 6 up to date with master.


gilescope:
Maybe to solve the no defaults we always include fancy and rely on the compiler to throw away the code if people choose onig? Either that or force people to pick a feature when they have no default features selected.

gilescope:
@trishume thoughts?

Adarma:
With this, I am still unable to build on Windows even without the onig feature enabled.
I created pull request #284 to make fancy the default on windows and not depend on onig at all.
onig remains default for other OSs with fancy as an optional feature. That chage builds on windows by default.

gilescope:
Totally happy with #284. Sorry on a mac here so didn't test it on windows. Curious what the build failure was, but really any which way we get a pure rust impl works for me. @trishume does #284 work for you? (I'm keen on cargo-expand being pure rust as lots of people try and use it)

Cargo.toml
```diff
@@ -20,7 +20,7 @@ exclude = [
 [dependencies]
 yaml-rust = { version = "0.4", optional = true }
 onig = { version = "5.0", optional = true }
-fancy-regex = { version = "0.3.0", optional = true }
+fancy-regex = { version = "0.3.0" }
```
trishume: Maybe I'm missing a catch here, but if we're going this way would it be possible to rejigger things so that `fancy-regex` is included by the `parsing` feature? This might require adding some cfgs to some modules to require the parsing feature to be built. It used to be the case that when no default features were included no regex engine was included, which was at least nice for Xi, which only uses the theme functionality.

gilescope:
Didn't realise it could work with no regex. Have introduced a NullObject implementation for when no regex feature is selected. Again happy with either PR :-) whatever gets us moving!

Adarma:
The build failure I get is the same as #264

robinst:
So this is merging into my branch? I actually rebased myself, pushed it now. Sorry about that. What's left on my branch is running the benchmarks again and add the documentation to the readme.

Do you want to rebase this branch again? All my commits should just fall away and we'll have a clean diff of your changes. If not, just cherry-pick your changes.

gilescope:
@robinst's solution is neater and passes all the tests so I'm going to close this PR to reduce confusion :-)

PR #284, opened 2020-03-19
No onig on windows

Adarma:
Even with onig as an unused feature, it still causes the build to fail on windows.

Conditionally excludes onig dependency on windows.
Use fancy-regex by default on windows.

Use onig-regex by default on not-windows.
fancy-regex is a feature for not-windows.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ exclude = [
 [dependencies]
 yaml-rust = { version = "0.4", optional = true }
 onig = { version = "5.0", optional = true }
-fancy-regex = { version = "0.2.0", optional = true }
+fancy-regex = { version = "0.3.0", optional = true }
 walkdir = "2.0"
 regex-syntax = { version = "0.6", optional = true }
 lazy_static = "1.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,8 +19,6 @@ exclude = [
 
 [dependencies]
 yaml-rust = { version = "0.4", optional = true }
-onig = { version = "5.0", optional = true }
-fancy-regex = { version = "0.3.2", optional = true }
 walkdir = "2.0"
 regex-syntax = { version = "0.6", optional = true }
 lazy_static = "1.0"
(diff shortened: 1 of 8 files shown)


Cargo.toml
```diff
@@ -34,6 +32,13 @@ serde = "1.0"
 serde_derive = "1.0"
 serde_json = "1.0"
 
+[target.'cfg(windows)'.dependencies]
+fancy-regex = "0.3.0"
```
trishume: So I don't quite understand why this works (if it does) given that the feature gates still look for `feature="regex-fancy"`.

Would it be possible to instead take an approach like

```
[target.'cfg(windows)'.features]
default = ["default-fancy"]
[target.'cfg(not(windows))'.dependencies]
default = ["default-onig"]
```
This would preserve the option of Windows users using onig, which as far as I understand (I might be wrong), does actually work on Windows if you install clang, and can build binary releases that don't require users install anything.

Also if fancy-regex works I might be comfortable just making it the default.

Adarma:
I tried `[target.'cfg(windows)'.features]`, but it seems it is not valid. Mentioned as an idea here:
https://github.com/rust-lang/cargo/issues/1197#issuecomment-268203727

The problem I have, is that with onig as a dependancy, even an unused optional dependancy, the build fails. Even `cargo doc` fails.

The only solution I found so far is to completely exclude onig on Windows.

I've not tried with clang installed. I expect many people will not have it.

The feature gates work because of the `any` and `all` I think. Not actually tested on Linux or Mac though.

Adarma:
Rebased onto @robinst `move-regex-use-to-module` branch

gilescope:
I think the docs should be updated to make it clear that to build a pure rust version you would have to both select the regex-fancy and also remove the default feature of onig:
```
cargo build --features regex-fancy --no-default-features
```
Using the above I've compiled @robinst 's PR branch on windows fine. I'm all for pure rust implementations but I suggest we switch the default on all OSes when people are happy there's no bugs / performance disadvantage. Could you re-check his branch works for you if you disable the default features?

Adarma:
Yes, this works fine for me:
```
syntect = { git = "https://github.com/trishume/syntect", branch = "move-regex-use-to-module", default-features = false, features = ["default-fancy"] }
```
Had not thought to disable default features.
This can be closed.

PR #292, opened 2020-05-10
Update plist to 1.0

ebarnard:
API is unchanged from 0.5 so this should be easy.


trishume:
Thanks for this, I accepted a different PR that did this along with bumping another dependency though.

PR #294, opened 2020-05-16
Format code using 'cargo fmt'

Atul9:


trishume:
I've kind of become lax at maintaining cargo fmt because it's easier to accept pull requests if I don't re-run it every time, and things seem fine without it.

It also appears that you use submitting low-effort 'cargo fmt' PRs as a way to game Github's contribution green boxes and charts, so if someone would appreciate the codebase being formatted with a newer rustfmt again I'll let some frequent contributor do it.

PR #293, opened 2020-05-16
Update dependencies

Keats:
Including the new version of onig that allows disabling bindgen.

PR #295, opened 2020-05-17
chore: Update plist to 1

ignatenkobrain:


sharkdp:
This is the third PR that proposes this, I think this can be closed: #292, #293

ignatenkobrain:
Uh oh, right.

PR #296, opened 2020-05-21
html: allow a configurable prefix on CSS class names.

quasicomputational:
CSS class names are a global namespace. Without some kind of namespace
partitioning, it's entirely possible to have class name collisions and
all sorts of chaos as different sources of CSS stomp on each other.

Hence, add a new `ClassStyle`, `SpacedPrefixed`, which tells syntact
to add a configurable prefix to all of the class names that it uses.

This patch breaks the API. Some of the breakage is not strictly
necessary (e.g., `ClassedHTMLGenerator::new` could stay as it is, and
a `new_with_style` method added). However, some is enforced:
`ClassStyle` can no longer be `Copy` because it can now contain a
`String`, and so `tokens_to_classed_spans` must now accept a
reference. Because this is a breaking API change anyway, I took the
opportunity to add `ClassStyle` parameters where relevant, to make the
parametricity of `ClassedHTMLGenerator` and `css_for_theme` clear.


quasicomputational:
Having slept on this and used it in anger, I've altered the API changes a bit - there's still a technical, enforced breakage, but much smaller than before. The commit message explains a little bit more.

src/html.rs
```diff
@@ -132,27 +143,73 @@ pub fn css_for_theme(theme: &Theme) -> String {
     css
 }
 
-/// Only one style for now, I may add more class styles later.
-/// Just here so I don't have to change the API
 #[derive(Debug, PartialEq, Eq, Clone, Copy)]
-pub enum ClassStyle {
+pub enum ClassStyle<'a> {
```
trishume, line 147: While you're at it could you add the new `#[non_exhaustive]` attribute here so adding new things isn't technically a breaking change (although unlikely to break anything in practice) anymore?

src/html.rs
```diff
@@ -132,27 +143,73 @@ pub fn css_for_theme(theme: &Theme) -> String {
...
-    assert!(style == ClassStyle::Spaced); // TODO more styles
-    let repo = SCOPE_REPO.lock().unwrap();
-    for i in 0..(scope.len()) {
-        let atom = scope.atom_at(i as usize);
-        let atom_s = repo.atom_str(atom);
-        if i != 0 {
-            s.push_str(" ")
-        }
-        s.push_str(atom_s);
+    match style {
```
trishume, line 165: Instead of putting the match statement on the outside and duplicating the loop code in each branch, could you put it around the `s.push_str(&prefix)` and just have the `ClassStyle::Spaced` branch be `()`? It should be a very predictable branch so shouldn't have a noticeably performance cost.

src/html.rs
```diff
@@ -132,27 +143,73 @@ pub fn css_for_theme(theme: &Theme) -> String {
...
+                }
+                s.push_str(&prefix);
+                s.push_str(atom_s);
+            }
+        },
+    }
+}
+
+fn scope_to_selector(s: &mut String, scope: Scope, style: ClassStyle) {
+    match style {
```
trishume, line 193: Same thing about pushing the match statement down here.

quasicomputational:
Yeah, makes sense. I've gone ahead and rolled back all the breaking changes (except for the new constructor, which is the most technical of technical breaks), instead deprecating the non-parameterised versions. `&'static str` for the prefix has a safety advantage, too, in that you can't accidentally XSS yourself, though that's also a fairly unlikely scenario.

trishume:
CI keeps being broken even when I retry it, at a stage before it seems to run any of my code. Travis has broken in this way before but fixed itself, but it's been much longer this time so maybe it won't. I ran the tests for this PR locally instead and they passed.

trishume:
I released `v4.3.0` with this change. Thanks for contributing!

PR #305, opened 2020-08-21
Rustfmt *; add doc links

sharnoff:
There's a lot here.

This mostly just started because I was interested in forking `syntect` for a more specialized application, and I decided that my first task was to update all of the docs so that they were easy to navigate - primarily adding links for all referenced types and functions, but also including certain opinionated stylization of doc comments.

The `rustfmt` happened because my editor typically runs it whenever a file is saved. Because it was changing many of the files I worked on, I just ran `cargo fmt` to make sure everything was formatted equally.

---

Feel free to not include this - there's a bunch of changes here and they aren't strictly speaking necessary. But I've found the docs I updated to be useful, just through a first look with
```
cargo doc --open
```

---

Update: I've done a little rewriting of history so now there's only two clear commits; the first is for running `cargo fmt` - that's all that is. The second is that visible changes I've discussed above.

### DIFF
--- a/benches/highlighting.rs
+++ b/benches/highlighting.rs
@@ -1,11 +1,11 @@
-use criterion::{Bencher, Criterion, criterion_group, criterion_main};
-use syntect::parsing::{SyntaxSet, SyntaxReference, ScopeStack};
-use syntect::highlighting::{ThemeSet, Theme};
-use syntect::easy::HighlightLines;
-use syntect::html::highlighted_html_for_string;
-use std::str::FromStr;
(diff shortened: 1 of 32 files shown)


trishume:
I used some git revert and squash trickery to merge this with only the doc changes, thanks for doing this and thanks for making it into two separate commits!

See https://github.com/trishume/syntect/commit/9c9f7f6631d527d24856ed1b27da90f84bf507ac

sharnoff:
Oh cool, great! :) Thanks!

Issue #303, opened 2020-08-13
Incorrect Python Comments Color Highlighting


mitnk:
Python Code like this,
```python
def foo():
    pass

# comments in middle will make normal code gray too
def bar():
    pass
```

will be rendered like this:
![image](https://user-images.githubusercontent.com/177049/90124302-bcc1c280-dd92-11ea-8822-d16e1fcd685c.png)

Using Python syntax:
```rust
...
let syntax = syntax_set.find_syntax_by_name("Python").unwrap();
...
```

keith-hall:
This is using `load_default_newlines`? Can you share a minimum reproducible example implementation code please? Afaik, the examples provided like `syncat` and `synhtml` are working correctly...

mitnk:
@keith-hall Here is a minimum code to reproduce it:
```rust
use syntect::html::ClassedHTMLGenerator;
use syntect::parsing::SyntaxSet;

fn main() {
    let code = "// one line comment\nvoid main() { }\n";

    let syntax_set = SyntaxSet::load_defaults_newlines();
    let syntax = syntax_set.find_syntax_by_name("C").unwrap();

    let mut html_generator = ClassedHTMLGenerator::new(&syntax, &syntax_set);
    for line in code.lines() {
        html_generator.parse_html_for_line(&line);
    }
    let html = html_generator.finalize();

    println!("html: {:?}", html);
}
```

In `html`, you can see, the code line (i.e. `void main() { }`) are compiled into the `comment <span>`. This issue happens in `C`, `Python`, but not `Rust`.

Since `ClassedHTMLGenerator` just deprecated in latest version of syntect, I will try using the new way to see if the issue still exists (may not related).

trishume:
@mitnk yah this is a common problem the docs warn about but it's easy to miss it and there's not much to indicate what's going wrong other than weird highlighting. All you need to do is either switch to `SyntaxSet::load_defaults_nonewlines()` or better yet use the iteration method used in the examples rather than `.lines()`, because `.lines()` doesn't include the newlines.

I might try and think of some way to warn or error if you provide the wrong type of lines, but it's made somewhat tricky by it being potentially okay for the last line to not have a newline after it even in the `_newlines` mode.

mitnk:
[...]

PR #307, opened 2020-08-23
Fix handling of newlines in classed HTML

trishume:
The examples used the `_newlines` syntax without providing newlines,
leading to occasional broken highlighting (#303). The function for producing
classed HTML added newlines for you so I deprecated that one and
added a new one which doesn't, so that I could change the examples to
use the `_newlines` variant in the proper way.

cc @mitnk @uwearzt @quasicomputational

### DIFF
--- a/src/html.rs
+++ b/src/html.rs
@@ -17,7 +17,10 @@ use std::path::Path;
 /// There is a [`finalize()`] method that must be called in the end in order
 /// to close all open `<span>` tags.
 ///
-/// The lines returned don't include a newline at the end.
+/// Note that because CSS classes have slightly different matching semantics
+/// than Textmate themes, this may produce somewhat less accurate
+/// highlighting than the other highlighting functions which directly use
+/// inline colors as opposed to classes and a stylesheet.
 ///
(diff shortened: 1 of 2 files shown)


src/html.rs
```diff
@@ -26,18 +29,19 @@ use std::path::Path;
...
 ///
 /// let current_code = r#"
 /// x <- 5
 /// y <- 6
 /// x + y
-/// "#.to_string();
+/// "#;
 ///
-/// let syntax_set = SyntaxSet::load_defaults_newlines();
+/// let syntax_set = SyntaxSet::load_defaults_nonewlines();
```
quasicomputational, line 40: Is this right? IIUC, either this needs to change, or the code feeding the HTML generator needs to change, but not both?

quasicomputational:
That nit aside, LVGTM - it fixed a mis-highlighting that I'd been meaning to get around to investigating for a while now, which is handy indeed!

src/html.rs
trishume: Took me a while to get around to this but you're very right thanks for catching this.

PR #316, opened 2020-11-15
Update syntaxes

Keats:
I tried to update syntaxes for Zola and ran into some errors, namely

 `ParseSyntax(MissingMandatoryKey("match"), Some("testdata/Packages/ASP/HTML (ASP).sublime-syntax"))`

I fixed some other tests to handle some of the updated syntaxes but I don't know enough about ST files to know if it's a new feature syntect is not handling or an issue in the syntax.


Keats:
cc @keith-hall

keith-hall:
Syntect doesn't support syntax definition inheritance yet, or other "version 2" features like popping multiple contexts in one go - that could be the likely cause. Realistically, syntect only supports ST3 features atm, so anything added in the ST4050 build or newer is not going to work. I recommend only updating to the ST3 branch from the Packages repo for now - it would be a lot of work to keep a fork with only compatible commits.

I guess the main blocker for now is https://github.com/trishume/syntect/issues/271 - it would be useful to discuss what the new public API should look like.

Keats:
ah I missed that issue sorry!

jrappen:
As Keith already said, this doesn't work and merging this PR should be deferred until ST4 support is added.

---

Your build breaks due to #279 having changed the submodule `Packages` to target the `st3` branch (temporarily). ST3 doesn't support branching (ST4050+) or even `v2` (ST4075+) and related changes.

Also compare the (new) docs at:

* <https://www.sublimetext.com/docs/syntax.html#ver-dev>
* <https://www.sublimetext.com/docs/syntax.html#ver-3.2>

Enselic:
Since supporting ST4 features probably is not close in time, I'm closing this PR for now to keep the PR inbox clean.

PR #322, opened 2020-12-29
Allow removing syntax from SyntaxSetBuilder

### LINKED ISSUE #321: Remove syntax from SyntaxSetBuilder
My usecase for this is loading the Sublime syntaxes and overriding some of them with newer definitions but without a way to remove from the builder some syntaxes might be in double. The alternative is to list exactly which syntax to load instead of using `add_from_folder` on the whole submodule. Would a PR for a `remove(name: &str)` (parallel to add) be accepted?

Keats:
Closes #321

### DIFF
--- a/src/parsing/syntax_set.rs
+++ b/src/parsing/syntax_set.rs
@@ -372,6 +372,12 @@ impl SyntaxSetBuilder {
         self.syntaxes.push(syntax);
     }
 
+    /// Remove a syntax from the set by name
+    /// Does not error if no matching syntax is found
+    pub fn remove(&mut self, name: &str) {
+        self.syntaxes.retain(|s| s.name != name);
+    }
+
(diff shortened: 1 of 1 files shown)


trishume:
I re-read the code and there's an `into_builder` method on `SyntaxSet` which converts something back to a builder while keeping existing index-based links in a way which is preserved when appending to the list of syntax sets but not when removing from the middle.

Can you please add a note to the doc comment warning that using `remove` on a `SyntaxSetBuilder` created via `into_builder` can cause panics during parsing? And if you really want to be a go-getter (not required) add a boolean flag to `SyntaxSetBuilder` based on whether it was created by `into_builder` and panic in `remove` if it was.

Enselic:
Closing to keep PR inbox clean. Of course feel free to re-open if you resume work on this.

PR #325, opened 2021-02-06
Derive Hash for Color and Style

nicholasbishop:


trishume:
Oh jeez sorry I forgot about this in my inbox forever. I hope you've been able to make do with Cargo's ability to use forks via git or something like that.

nicholasbishop:
No worries! I had worked around it locally:

```rust
#[derive(Eq, PartialEq)]
struct StyleWithHash(Style);

#[allow(clippy::derive_hash_xor_eq)]
impl Hash for StyleWithHash {
    fn hash<H: Hasher>(&self, state: &mut H) {
        let mut hash_color = |c: &syntect::highlighting::Color| {
            c.r.hash(state);
            c.g.hash(state);
            c.b.hash(state);
            c.a.hash(state);
        };
        hash_color(&self.0.foreground);
        hash_color(&self.0.background);
        self.0.font_style.hash(state);
    }
}
```

Thanks for merging :)

PR #330, opened 2021-04-14
Export the HTML escape helper

camdencheek:
When building custom HTML structure, this helper is useful to have
exported. Most of the remaining functions for building custom
CSS-classed HTML are already exported.

Feel free to close if you don't want this as part of the external interface. 
It's not terribly difficult to copy or reimplement, but I figured it'd be nice
if I can use the same escaping logic that's used internally. 

Closes #330

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -34,7 +34,7 @@ pub mod dumps;
 #[cfg(feature = "parsing")]
 pub mod easy;
 #[cfg(feature = "html")]
-mod escape;
+pub mod escape;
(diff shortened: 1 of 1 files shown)


keith-hall (approved):

Enselic:
It is generally a good idea to keep the API surface minimal. IMHO, exporting this would add unnecessary API to syntect. Ideally syntect should remove this code and use some existing crate for it.

So I actually think this PR should be closed.

Thank you anyway for taking the time to open a PR.

PR #332, opened 2021-04-20
Add public method find_unlinked_contexts to SyntaxSet

keith-hall:
To help ensure at asset/syntax set building time that all contexts which are push/set targets are resolvable, and no run-time errors should occur while parsing due to unlinked context references. Relates to: https://github.com/sharkdp/bat/issues/915#issuecomment-626380174

I wasn't sure the best way to present the results, so I'm just doing a debug formatting on any ContextReference which can't be found for now. Also, it may be useful in future to add some details about which context (i.e. the name given to the context) the match pattern with an invalid context reference was found in.

### DIFF
--- a/src/parsing/syntax_set.rs
+++ b/src/parsing/syntax_set.rs
@@ -359,6 +359,41 @@ impl SyntaxSet {
             self.first_line_cache.borrow().unwrap()
         }
     }
+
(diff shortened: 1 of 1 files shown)


trishume (approved):

trishume:
Thanks, this feature makes sense and the implementation seems fine

PR #337, opened 2021-06-22
Pass a ScopeStack to tokens_to_classed_spans

Keats:
Also remove deprecated functions


src/html.rs
```diff
@@ -544,6 +515,19 @@ mod tests {
         assert_eq!(html, include_str!("../testdata/test5.html"));
     }
 
+    #[test]
+    fn test_classed_html_generator_doesnt_panic() {
```
Keats, line 519: Updating the syntaxes caused a bunch of unrelated test failures so this test is not actually failing even without the fix. Should I remove it?

keith-hall: I think it makes sense to keep it, personally, for when we can/do update the syntaxes. Maybe it could also make sense to use a custom syntax definition purely for this test?

keith-hall (approved):

src/html.rs
Keats: Added

trishume:
This looks good, but want to pitch me on why to go with the major version bump rather than just renaming and then writing a tiny wrapper that passes an empty scopestack under the old name and marking it deprecated? It isn't that hard and it allows postponing a major version bump.

Keats:
It would mean there is an accessible broken API but I can do that if you want.

keith-hall:
I guess we are waiting for this before we make a new release? :)

trishume:
I reverted the removal of the deprecated methods myself and merged this in 77514b2ec9a9e85b88e478b75ca9152b3c24a2b8

I prefer deprecation because I think it actually makes it more likely people will update to the less broken version. There's no notification for new major versions of your cargo dependencies, and you need to map breaking changes to required upgrades yourself, but if you or your users pull a new semver-compatible syntect then you get a deprecation message pointing directly at your code and telling you how to update it and why.

trishume:
This was released in v4.6.0

PR #338, opened 2021-06-30
Improve find_unlinked_contexts output to make it easier to find the "culprit"

keith-hall:
Improve the output of the `find_unlinked_contexts` method to include the syntax name and scope from where the unlinked context reference originated, and deduplicate the results. I've changed it to return a HashSet instead of a Vec - this method has never been in a release so it shouldn't matter. But if you think keeping the Vec return type makes sense then we can change it back again :)

### DIFF
--- a/src/parsing/syntax_set.rs
+++ b/src/parsing/syntax_set.rs
@@ -384,36 +384,35 @@ impl SyntaxSet {
                     builder_contexts.insert(context_name, context);
 
                     for pattern in context.patterns.iter() {
-                        match pattern {
+                        let maybe_refs_to_check = match pattern {
                             Pattern::Match(match_pat) => {
-                                let mut maybe_refs_to_check = None;
                                 match &match_pat.operation {
                                     MatchOperation::Push(context_refs) => {
-                                        maybe_refs_to_check = Some(context_refs);
+                                        Some(context_refs)
                                     },
                                     MatchOperation::Set(context_refs) => {
-                                        maybe_refs_to_check = Some(context_refs);
+                                        Some(context_refs)
                                     },
-                                    _ => {},
-                                }
-
-                                if let Some(refs_to_check) = maybe_refs_to_check {
-                                    for context_ref in refs_to_check {
-                                        match context_ref {
-                                            ContextReference::Direct(_) => {},
-                                            _ => {
-                                                unlinked_contexts.insert(
-                                                    format!(
-                                                        "Syntax '{}' with scope '{}' has unresolved context reference {:?}",
-                                                        name, scope, &context_ref
-                                                    )
-                                                );
-                                            },
-                                        }
-                                    }
+                                    _ => None,
                                 }
                             },
-                            _ => {},
+                            _ => None,
+                        };
+
+                        if let Some(refs_to_check) = maybe_refs_to_check {
+                            for context_ref in refs_to_check {
+                                match context_ref {
+                                    ContextReference::Direct(_) => {},
+                                    _ => {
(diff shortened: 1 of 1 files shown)


src/parsing/syntax_set.rs
```diff
@@ -360,35 +360,62 @@ impl SyntaxSet {
...
-                                        unlinked_contexts.push(format!("{:?}", &context_ref));
+            let mut builder_contexts = HashMap::with_capacity(contexts.len());
+            for (context_name, context_id) in contexts {
+                if let Some(context) = context_map.remove(&context_id.index()) {
+                    builder_contexts.insert(context_name, context);
+
+                    for pattern in context.patterns.iter() {
+                        match pattern {
+                            Pattern::Match(match_pat) => {
+                                let mut maybe_refs_to_check = None;
```
trishume, line 389: `match` is an expression so this can just be 

```rust
let maybe_refs_to_check = match ... { pat => Some(context_refs), _ => None };
```

src/parsing/syntax_set.rs
```diff
@@ -360,35 +360,62 @@ impl SyntaxSet {
...
+                                    },
+                                    MatchOperation::Set(context_refs) => {
+                                        maybe_refs_to_check = Some(context_refs);
                                     },
+                                    _ => {},
                                 }
-                            }
+
+                                if let Some(refs_to_check) = maybe_refs_to_check {
+                                    for context_ref in refs_to_check {
```
trishume, line 401: If you want to go for maximum style points this can be `maybe_refs_to_check.iter().flatten()` since options can be iterated

trishume (approved):
Looks fine, a couple minor style cleanups you can make if you want, but otherwise just ping me to merge and I will.

src/parsing/syntax_set.rs
keith-hall: good point, thanks - I took your advice and improved this :)

src/parsing/syntax_set.rs
keith-hall: my Rust skills are still not good enough to get this working it seems. If someone can spare a few minutes to (help me) implement this, it would be great because I really like the suggestion!

sharkdp: My guess would be that @trishume meant something like this:
``` rust
for context_ref in maybe_refs_to_check.into_iter().flatten() {
    match context_ref {
        ContextReference::Direct(_) => {},
        _ => {
            unlinked_contexts.insert(
                format!(
                    "Syntax '{}' with scope '{}' has unresolved context reference {:?}",
                    name, scope, &context_ref
                )
            );
        },
    }
}
```


`maybe_refs_to_check` is a `Option<Vec<ContextReference>>`. The outer level (`Option`) can be turned into an iterator (with either zero or one elements). Calling `flatten()` on this iterator returns a new iterator that loops over the flattened structure: either zero elements (if the `Option` is `None`) or N elements (if the `Option` is `Some(vec![1, 2, …, N])`).

keith-hall: Thanks for the help and patient explanation, @sharkdp - it worked perfectly! I believe this is now ready to merge @trishume, if you're okay with my HashSet -> BTreeSet change to preserve insertion order.

trishume (approved):

src/parsing/syntax_set.rs
trishume: BTW a hash set won't preserve insertion order, it just iterates in sorted rather than random order, which is still an improvement.

PR #339, opened 2021-07-02
SyntaxReference: Add some docs

Enselic:
Hi and thanks for a great project!

I'm in the process of exploring the inner workings of syntect. Here is a PR that adds some docs to `SyntaxReference` that would have sped up my learning process. Hopefully someone that comes after me will find it helpful.


keith-hall (approved):

Enselic:
(I just force-pushed a fix for me not having backticks in references like the other docs have. Thanks for the quick Approved btw Keith!)

PR #340, opened 2021-07-04
SyntaxSetBuilder: Allow to list added syntaxes via a syntaxes() method

Enselic:
This is analogous to the existing public `SyntaxSet::syntaxes()` method.

### Background information

I am doing some prototyping around ways to improve the startup time of bat (see https://github.com/sharkdp/bat/issues/951). And the way to improve startup time is to only load the syntaxes needed for the given input file(s). Instead of having one giant `SyntaxSet`, we need several smaller ones.

The added method enables clients such as bat to analyze dependencies between `SyntaxDefinition`s and build up many smaller, self-contained `SyntaxSet`s.

An alternative to this PR could be to move out `SyntaxSetBuilder::add_from_folder()` to some place where the logic can live and be used independently, i.e. without taking a detour through a `SyntaxSetBuilder`. That however seems to like a significantly higher hanging fruit at this point.

### DIFF
--- a/src/parsing/syntax_set.rs
+++ b/src/parsing/syntax_set.rs
@@ -407,6 +407,11 @@ impl SyntaxSetBuilder {
         self.syntaxes.push(syntax);
     }
 
+    /// The list of syntaxes added so far.
+    pub fn syntaxes(&self) -> &[SyntaxDefinition] {
+        &self.syntaxes[..]
+    }
+
     /// A rarely useful method that loads in a syntax with no highlighting rules for plain text
     ///
     /// Exists mainly for adding the plain text syntax to syntax set dumps, because for some reason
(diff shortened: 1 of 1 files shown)


trishume:
This seems fine with me. You may have already thought about this, but syntaxes have dependencies on each other due to embedding, and while sometimes syntaxes will only depend on one or two others, there's other cases like the Ruby or Markdown syntax where it can reference tons of different languages based on the embedded snippet name. If you don't share the serialized definition this could bloat the bat binary a lot. Sharing them will require linking them together after loading the different syntaxes (which might be pretty fast and not a problem).

As a more ambitious project I'd also accept a sufficiently good contribution of an implementation of lazy-loading syntaxes on the first usage of a rule with a cross-syntax reference. This would also allow faster startup for languages like Markdown which reference a lot of languages, although given the number of syntaxes that bat includes Markdown will be faster even with the basic version.

Enselic:
@trishume There is absolutely no rush or urgency, but it would be great if you could make a new release of syntect where this API is available, when you get time.

Maybe I can help you update `CHANGELOG.md` for a new release? Just let me know.

Also thank you for the pointers with regards to dependencies between syntaxes. In case you are curious, the bat code to analyze dependencies, and the output of it, can be found in this PR: https://github.com/sharkdp/bat/pull/1762 (which is also the PR that depends on this added API).

trishume:
No problem, thanks for reminding me. This was released in `v4.6.0`

PR #342, opened 2021-07-05
Start using GitHub Actions instead of Travis CI

Enselic:
Somewhat quick and dirty conversion, but seems to work fine. Certainly
better than nothing, since Travis CI does not run at all any longer.

The badge should start working once this lands in master.
Test run in my fork: https://github.com/Enselic/syntect/runs/2988392147?check_suite_focus=true


Enselic:
Since this PR did not trigger a workflow here, I suspect Actions is disabled for this repo. You might want to look over the settings: https://github.com/trishume/syntect/settings/actions

trishume:
Thanks for doing this, I knew Travis was dying but didn't notice it finally died. It looked like I already had actions enabled on the repo when I went to the page, so I merged this and hopefully that will cause actions to start working.

PR #344, opened 2021-07-25
Bump fancy-regex to 0.7

### LINKED ISSUE #343: fancy-regex 0.7
Hello,

Seems that current dependency on fancy-regex is ^0.3 which is extremely old and the latest version is 0.7.0 nowadays (that's been around for a while too).

Would it be possible to update the dependency?

That would help a lot in my packaging work and improve ecosystem in general.

Thank you!

Enselic:
Closes #343

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ exclude = [
 [dependencies]
 yaml-rust = { version = "0.4", optional = true }
 onig = { version = "6.0", optional = true, default-features = false }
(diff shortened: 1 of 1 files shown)


Enselic:
All regression tests pass, and I find nothing worrying in https://github.com/fancy-regex/fancy-regex/blob/main/CHANGELOG.md, so seems OK to bump to me.

keith-hall (approved):

trishume:
Thanks!

PR #345, opened 2021-08-10
Add 'plist-load' feature, analogous to existing 'yaml-load' feature

Enselic:
## Why do this?

The `plist` package that `syntect` currently unconditionally depends brings in several dependencies on its own, the heaviest being `xml-rs`. For use cases where loading of themes is not needed, it is pleasant to not have to compile unneeded dependencies.

This new `plist-load` feature is very similar to the existing `yaml-load` feature, both in purpose, and how it is implemented. Whereas `yaml-load` is for syntax definitions, `plist-load` is for themes and theme preferences.

## Public API is not affected

I have taken care not to change the public API (with all features enabled) with this change.

That the public API has not changed (at least not in terms of added/removed/moved items) can be verified with this little oneliner:
```
rm -f /tmp/all-*.html && for commit in plist-load-feature  master; do git ch $commit && cargo clean -p syntect && cargo doc --all-features && cp target/doc/syntect/all.html /tmp/all-$commit.html; done && diff -u /tmp/all-*.html | diff-so-fancy
```

## Example of usefulness in practice

To demonstrate that this feature is useful, I recommend to compare the number of dependencies you have to build when you do this:

```
cp examples/synhtml.rs src/main.rs # temporarily side-step dev-dependencies
cargo run --no-default-features --features html,assets,dump-load,regex-onig -- examples/synhtml.rs
```

Before this change, you need to build 68 dependencies. After this change, it is only necessary to build 52 dependencies.

## Guide to Code Review

I recommend to Code Review each commit in this PR individually.
The first two commits are simple refactorings. The third commit just cuts and pastes code around. The fourth commit introduces the `plist-load` feature, and also adds an automated CI test for this feature.


keith-hall (approved):

trishume:
Thanks, this makes sense!

PR #346, opened 2021-08-10
CI: Add job to ensure MSRV compatibility, currently 1.42

Enselic:
Add a job to CI to ensure compatibility with current min supported rust version. This project currently works with Rust 1.42, but fails with Rust 1.41 with `cargo check --all-features --all-targets`.

For just `cargo check --all-features`, Rust 1.40 is enough, but doesn't hurt to use Rust 1.42 I think. Rust 1.42 is well over a year old by now.

To keep the job quick, I'm just doing a `cargo check`. A `cargo test` takes significantly longer time, and we already run `cargo test` in various configurations in the other job. It should be safe to assume that if the project code compiles with MSRV, the tests will also pass (if they pass in the main job).

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -8,9 +8,22 @@ on:
     branches: [ master ]
 
 env:
+  MIN_SUPPORTED_RUST_VERSION: "1.42"
   CARGO_TERM_COLOR: always
 
 jobs:
+  min_version:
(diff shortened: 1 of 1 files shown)


Enselic:
Hmm, I can tell master fails right now because of this CI check

I'll sort it out with a new PR shortly

PR #347, opened 2021-08-11
Add ScopedRangeIterator

icsaszar:
Simple change to also expose the `Range` of the scope and not just the string slice in `ScopeRegionIterator`. 
This is super useful for building custom parsing logic on top of syntect without using the highlighter (which also has `HighlightIterator` and `RangedHighlightIterator`).

### DIFF
--- a/src/easy.rs
+++ b/src/easy.rs
@@ -152,6 +152,69 @@ impl<'a> HighlightFile<'a> {
     }
 }
 
+/// Iterator over the regions of a line (including the ranges) which a given the operation from the parser applies.
+///
+/// To use, just keep your own [`ScopeStack`] and then `ScopeStack.apply(op)` the operation that is
+/// yielded at the top of your `for` loop over this iterator. Now you have a substring of the line
+/// and the scope stack for that token.
+///
+/// See the `synstats.rs` example for an example of using this iterator.
+///
+/// **Note:** This will often return empty regions, just `continue` after applying the op if you
+/// don't want them.
+///
+/// [`ScopeStack`]: ../parsing/struct.ScopeStack.html
+#[derive(Debug)]
+pub struct ScopeRangeIterator<'a> {
+    ops: &'a [(usize, ScopeStackOp)],
+    line: &'a str,
+    index: usize,
+    last_str_index: usize,
+}
+
+impl<'a> ScopeRangeIterator<'a> {
+    pub fn new(ops: &'a [(usize, ScopeStackOp)], line: &'a str) -> ScopeRangeIterator<'a> {
+        ScopeRangeIterator {
+            ops,
+            line,
+            index: 0,
+            last_str_index: 0,
+        }
+    }
+}
+
+static NOOP_OP: ScopeStackOp = ScopeStackOp::Noop;
+
+impl<'a> Iterator for ScopeRangeIterator<'a> {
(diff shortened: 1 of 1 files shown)


trishume (changes requested):
Thanks for the PR! I think this is a pretty reasonable thing to want. You can get this functionality using the existing API by using a mutable variable which you add the `len` of the `str` pieces to as you go along, but it's less convenient.

Given that it's achievable now with some friction though, I'm hesitant to introduce this much code duplication. Instead I think the current `ScopeRegionIterator` could be renamed and made to return a range *instead of* a `str`, and then the `ScopeRegionIterator` could be easily implemented on top of that by just indexing the range into the line, without duplicating the body. When you have the range it's pretty easy to get the `str` if you want it so I don't think the range version needs to yield three things.

I know I'm a bit late reviewing this and you may not feel like implementing the deduplicated version, if so that's fine, I apologize for the inconvenience of the counter variable.

icsaszar:
> You can get this functionality using the existing API by using a mutable variable which you add the len of the str pieces to as you go along, but it's less convenient.

Yes, I did some googling before I opened the PR and found this [stack overflow](https://stackoverflow.com/questions/38268529/how-to-get-the-byte-offset-between-str) answer, but I didn't like the idea of using pointer arithmetic.

> Given that it's achievable now with some friction though, I'm hesitant to introduce this much code duplication. Instead I think the current ScopeRegionIterator could be renamed and made to return a range instead of a str, and then the ScopeRegionIterator could be easily implemented on top of that by just indexing the range into the line, without duplicating the body. When you have the range it's pretty easy to get the str if you want it so I don't think the range version needs to yield three things.

Here I chose to return a 3 tuple (slice, range, op) be consistent with the `highlight` module. I'm not sure I understand the part with code duplication, I refactored `ScopedRegionIterator`'s `next` to simply call `ScopeRangeIterator`'s `next` and extract the slice and stack operation.

Also to get it right, you'd like me to change `ScopeRangeIterator` to return `(&'a str, std::ops::Range<usize>, &'a ScopeStackOp)` and `ScopedRegionIterator` to use the returned `Range` and get the slice based on that?

> I know I'm a bit late reviewing this and you may not feel like implementing the deduplicated version, if so that's fine, I apologize for the inconvenience of the counter variable.

No problem, I'll happily do the refactoring.

trishume:
Huh I must have misread and thought that it was duplicated, I agree that it's actually not. Also huh you're right, I had forgotten the highlight module returned a 3-tuple, not sure why it works that way.

I think it's a little bit cleaner to have it just return a `(std::ops::Range<usize>, &'a ScopeStackOp)` and then do the slicing in `ScopedRegionIterator` like you say. But I now agree it's a small enough deal I'm fine to merge as-is if you don't feel like it.

And the solution I meant with the mutable variable is like
```rust
let mut prefix_len = 0;
for (s, op) in ScopeRegionIterator::new(&ops, line) {
  let range = prefix_len..(prefix_len+s.len());
  prefix_len += s.len();
  // ...
}
```

icsaszar:
Better late then never, I updated `ScopeRangeIterator` to return `(std::ops::Range<usize>, &'a ScopeStackOp)` and updated the documentation to include references between `ScopeRangeIterator` and `ScopeRegionIterator`.

trishume (approved):
Looks good, thank you!

PR #369, opened 2021-08-23
Make looking up a syntax by extension use case-insensitive comparison

keith-hall:
this matches Sublime Text behavior - even on platforms/environments which use a case-sensitive file system

first reported at https://github.com/sharkdp/bat/issues/1807


Enselic:
LGTM. We already use `eq_ignore_ascii_case()` for `syntax.name` in `SyntaxSet::find_syntax_by_token()`, so this change harmonize nicely with existing code.

Strictly speaking this is breaking backwards compatibility, since in theory clients could exist that rely on current behavior, but I think the benefits of this change is bigger than this theoretical downside :+1:

sharkdp (approved):
Thanks!

PR #388, opened 2021-10-31
Add Criterion benchmark test that includes loading assets

Enselic:
This makes it easy to measure the impact lazy-loading will have on a complete syntect "pipeline".

This commit is designed to be cherry-pickable to the v4.6.0 tag to get a Criterion `--baseline` for the lazy-loading changes.

An upcoming PR will make existing benches make use of the helper methods that this commit adds. I will wait with that until #387 is merged however, to reduce problems with conflicts.

### DIFF
--- a/benches/load_and_highlight.rs
+++ b/benches/load_and_highlight.rs
@@ -1,0 +1,40 @@
+mod highlight_utils;
+mod utils;
+
+/// Measures the time it takes to run the whole pipeline:
+///  1. Load assets
+///  2. Parse
+///  3. Highlight
+fn run(b: &mut criterion::Bencher, file: &str) {
+    let path = utils::get_test_file_path(file);
+
+    b.iter(|| {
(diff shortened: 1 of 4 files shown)


benches/load_and_highlight.rs
```diff
@@ -0,0 +1,40 @@
...
+    })
+}
+
+fn load_and_highlight_benchmark(c: &mut criterion::Criterion) {
+    let mut group = c.benchmark_group("load_and_highlight");
+    for input in &[
+        "highlight_test.erb",
+        "InspiredGitHub.tmTheme",
+        "Ruby.sublime-syntax",
+        "parser.rs"
```
Enselic, line 28: I did not bother including `"jquery.js"` here, because the time to load assets is very small compared to highlighting that complicated file. And I skipped ` "scope.rs"` since that files changes over time, since it is an actual syntect source file. So it is not not ideal for a benchmark across time/commits.

trishume (approved):
Good idea!

PR #389, opened 2021-11-02
Blend alpha value on converting colors to ANSI color sequences

rhysd:
Hello,

thank you for this great library. I'm building [hgrep](https://github.com/rhysd/hgrep) on it and it works greatly.

## Problem

Colors have alpha values for their transparency. But they are ignored in `utils::as_24_bit_terminal_escaped`.

## Solution

This PR blends foreground colors and background colors when the foreground color's alpha value is not max (=255) so that the alpha value is now reflected to the converted ANSI color sequence.

I was inspired by [the implementation in Neovim](https://github.com/neovim/neovim/blob/649b3160a10f0053747383f0703b0a5a94548570/src/nvim/highlight.c#L526-L626), which is a famous text editor.

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -3,11 +3,28 @@
 //! * Iterating lines with `\n`s
 //! * Modifying ranges of highlighted output
 
-use crate::highlighting::{Style, StyleModifier};
-use std::fmt::Write;
-use std::ops::Range;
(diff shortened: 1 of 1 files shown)


trishume (approved):
Looks good to me, thanks for adding this!

Issue #334, opened 2021-05-10
Panic in regex parser


epage:
The [CI for my project reported the following](https://github.com/cobalt-org/cobalt.rs/pull/854#issuecomment-835467115)
```
---- syntax_highlight::syntect::test::markdown_renders_rust stdout ----
thread 'syntax_highlight::syntect::test::markdown_renders_rust' panicked at 'called `Option::unwrap()` on a `None` value', /Users/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/syntect-4.5.0/src/parsing/regex.rs:74:33
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    syntax_highlight::syntect::test::markdown_renders_rust
```
That corresponds with the unwrap from
```
            let regex =
                regex_impl::Regex::new(&self.regex_str).expect("regex string should be pre-tested");
            self.regex.fill(regex).ok();
            self.regex.borrow().unwrap()
```

I can't reproduce locally; I'm assuming this is some kind of race, which suggests AtomicLazyCell is not actually threadsafe.  

Both locally and the PR in question are using syntect 4.5.0.  `lazycell` is set to 1.3.0 in my lock file which is the latest version.

trishume:
Ouch yah, sorry for the delayed reply. Things I'm thinking of:

- That `.ok()` is a weird way to ignore the error from if the cell is already full, but it should be fine to drop it.
- This looks like a bug in `lazycell`. I wondered if they aren't using the right ordering parameters on their atomic operations, and this only causes a problem when the compiler does some optimization based on that. I checked the code and I'm not sure whether it's right or not, the orderings on the lock atomics are right but I haven't researched the semantics of how they're ordered with other memory operations and how UnsafeCell works, it's plausible that the code is wrong.

rhysd:
I may figure out the reason. Let's say there are 2 threads trying to fill the `self.regex` lazy cell.

1. Thread1 starts `.fill()` call and acquires lock
2. Thread2 starts `.fill()` call but cannot acquires lock (since Thread1 already did it)
3. `.fill()` in Thread2 returns `Err(value)` (but Thread2 ignores it by `.ok()`)
4. Thread2 calls `.borrow()`. However, at this point, the `.fill()` call in Thread1 is not completed yet. So `.borrow()` in Thread2 returns `None`
5. The `.unwrap()` crashes
6. Thread1 finishes to set value at `.fill()`

So,

```rust
cell.fill(something).ok();
[...]

PR #390, opened 2021-11-02
Update lazycell to v1.3.0

rhysd:
Related to #334

This PR updates lazycell crate to the latest.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ fancy-regex = { version = "0.7", optional = true }
 walkdir = "2.0"
 regex-syntax = { version = "0.6", optional = true }
 lazy_static = "1.0"
-lazycell = "1.0"
+lazycell = "1.3"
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -24,7 +24,7 @@ fancy-regex = { version = "0.7", optional = true }
 walkdir = "2.0"
 regex-syntax = { version = "0.6", optional = true }
 lazy_static = "1.0"
-lazycell = "1.0"
+lazycell = "1.3"
```
Enselic, line 27: Syntect actually does not specify a certain version, only a minimal version. It is then clients of syntect that chooses the actual version. Chances are that if you look in Cargo.toml of syntect (which is not version controlled), you already used the lastest version. And I don't think we need anything to bump minimal version?

rhysd:
This PR will be closed by #391.

PR #397, opened 2021-11-22
Bump MSRV to 1.51 to fix build with latest plist release

Enselic:
A recent release of plist bumped MSRV to 1.51 (see
https://github.com/ebarnard/rust-plist/pull/72 and discussions therein). One easy way for us to deal with that is to also bump *our* MSRV to 1.51. I can't think of why this would be a big problem, so here is a PR that does that. It fixes this build error in MSRV CI that we currently get on the master branch (https://github.com/trishume/syntect/runs/4282657925?check_suite_focus=true):

```
error: failed to parse manifest at `/home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/time-0.3.5/Cargo.toml`

Caused by:
  feature `resolver` is required

  this Cargo does not support nightly features, but if you
  switch to nightly channel you can add
  `cargo-features = ["resolver"]` to enable this feature
```
Note that the build error is in time-rs, but it is plist that pulls in that dependency.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -8,7 +8,7 @@ on:
     branches: [ master ]
 
 env:
-  MIN_SUPPORTED_RUST_VERSION: "1.46"
+  MIN_SUPPORTED_RUST_VERSION: "1.51"
   CARGO_TERM_COLOR: always
 
 jobs:


Enselic:
I have confirmed that syntect still fails to build with Rust 1.50. Rust 1.51 is needed. I'll go ahead and merge this now since:
* This PR was opened more than one month ago (which is the [magic number](https://github.com/trishume/syntect/pull/382#issuecomment-950443083))
* This PR does not change the actual syntect library code
* Actually, this PR does not really change _anything_. Rust 1.50 fails to build with latest syntect (and a freshly auto-constructed Cargo.lock) regardless of if we merge this PR or not. So this PR just makes the repo reflect reality, so to speak.

Another benefit of merging this is that I can make my other open PRs pass all CI checks. (I plan on doing a hard rebase on top of master now for all my other open PRs now that this PR is merged.)

Issue #394, opened 2021-11-10
Abandon `nonewlines` for v5.0.0?


Enselic:
Practically since the [beginning](https://github.com/trishume/syntect/commit/8e408de2f7), syntect has had both `newlines` and `nonewlines`.

However, it seems to me as if `nonewlines` is more of a problem than a solution. It was called a 'hack' in https://github.com/trishume/syntect/issues/196#issue-347643743.

So, why not remove `nonewlines` in the upcoming v5.0.0 release? I would happily make a PR for that.

I deeply apologize in advance for any ignorance on my part. I'm sure I'm missing something crucial.

trishume:
Yah seems plausible that this would be net good. I mainly wanted to have it around for use by people who already had strings without a newline on the end, and couldn't easily change. It does cause a lot of mysterious highlighting issues though.

I wonder if there's a way to just make it more annoying or clear that if you choose to use it you better be real sure you're not better off just adding newlines, and that you're opening yourself up to mysterious highlighting issues even though it'll mostly work.

Enselic:
Thank you for providing additional background information.

What I still don't quite understand is why someone would want buggy highlighting at all. Making sure lines includes `\n` seems to me to be a very low price to pay for getting highlighting that works. Especially considering that there is stable Rust APIs that preserves `\n` (and that we recommend ourselves in the docs). It is of course very possible that there is a use case that I just don't consider. I would love to be enlightened.

I do however think you make a very strong point in that it would be great if it was easy to do the right thing (include `\n`) and hard to do the wrong thing (not include `\n`). That holds true even if we remove the `nonewlines` APIs. For example, we could introduce a zero-cost type `LineWithNewline` instead of using `&str`. With appropriate scaffolding, it would be very hard to mistakenly not include `\n` in lines, and very easy to *do* include it. There are also other venues that could be explored. Such as an opt-in cargo feature that removes a runtime check for the presence of `\n`.

I think removing `nonewlines` APIs in addition to making correct use of the API more type-safe has great long term benefits for both users of and contributors to syntect.

Users would:
* be confronted with a simpler API with less tricky tradeoff choices 
* not have to encounter `nonewlines` specific bugs

Contributors would:
* not have to spend time maintaining `nonewlines` code paths 
* not have to spend time triaging bugs caused by `nonewlines`
* not have to spend time  debugging `nonewlines`-specific problems

So, what would the next steps be here? I would like to propose the following:
[...]

PR #399, opened 2021-11-25
Don't print double newlines in load_defaults_newlines() sample code

Enselic:
Since the lines themselves end with a newline character, the sample code needs
to use `print!()` instead of `println!()`, otherwise an extra newline character
is inserted for each line in the output.

Before PR:
with-println

After PR
with-print
:


Enselic:
This PR just changes documentation, so I am comfortable merging this without reviews. I want to give more time for review though, so I will leave this open for some more time. I plan on merging this in the beginning of next year in case it gets no reviews.

This PR is actually a baby-step towards a solution for #394. I have plans on more baby-steps to take in the context of that issue/proposal, but they will have to wait.

keith-hall (approved):

PR #400, opened 2021-11-25
benches: Make use of new utils to reduce code duplication

Enselic:
As mentioned in #388, we can reduce code duplication by making use of the utils that that PR introduced.

This PR does that.

(Note that MSRV CI will fail since the PR that fixes that is not merged yet)


Enselic:
This PR just de-duplicates code using helper functions that were approved in #388, so I consider this PR implicitly approved and uncontroversial.

I intend to merge this PR after a few days, when it has been open for a month.

keith-hall (approved):

trishume (approved):
To be clear I'm happy with your reasoning and would've been pleased with you merging the PR on your own. I just happen to finally be less busy on holiday to take a look at some syntect emails.

PR #403, opened 2021-12-28
Fix feature semver violations in v4.7.0

### LINKED ISSUE #402: Semver violation in v4.7.0 and v4.7.1
The 4.7.0 release violates semver. See:
* sharkdp/bat#1991
* dtolnay/cargo-expand#122

Enselic:
Fixes #402 

I ended up removing the `plist-load` feature because I see no way of keeping it around in a semver compatible and sane way. The [refactoring](https://github.com/trishume/syntect/pull/345) that made the introduction of that feature possible can we kept however, which is nice. So it should be easy to re-introduce it e.g. v5.0.0 at some later point.

I did consider making `plist-load` opt-out rather than opt-in to remain semver compatible. We could call it `no-plist-load`, and if enabled, the unneeded code could be excluded from the build. I think that is too unergonomic and counter-intuitive however, so I think it's better to simply remove it for now.

I also noticed that `dump-create-rs` and `dump-create` are the same since 2 years, since 2 years ago the default flate2 backend was changed to Rust. See https://github.com/rust-lang/flate2-rs/commit/c479d064e2. So there is no need to make a distinction any longer. This allows us to remain semver compatible with regards to the lazy-loading changes since now `parsing` can depend on `dump-create` and `dump-load` directly. I did not bother adding an entry about that in CHANGELOG.md since there is no semantic change for clients.

I could reproduce the build error in bat locally, and have confirmed that these changes fixes the build error.

I think it would be a good idea to not rush this PR in. Let's let it sit here for at least a few days so we have time to think this through and perhaps test some more. Any help with testing would be greatly appreciated.

(Oh, and I also noticed that I forgot to add a `pub` to an API, so I fixed this too in this PR.)

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -76,4 +76,4 @@ jobs:
     - name: Run Xi tests
       run: |
         # Test the build configuration that Xi uses
-        cargo test --lib --no-default-features --features "assets dump-load-rs"
+        cargo test --lib --no-default-features --features "assets dump-load"
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -59,7 +59,7 @@ jobs:
       run: |
         # Run these tests in release mode since they're slow as heck otherwise
         cargo test --features default-fancy --no-default-features --release
-    - name: Ensure highlight works without 'plist-load' and 'yaml-load' features
+    - name: Ensure highlight works without 'yaml-load'
       run: |
         cargo run --example synhtml --no-default-features --features html,assets,dump-load,dump-create,regex-onig -- examples/synhtml.rs
     - name: make stuff
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## [Version 4.7.1](https://github.com/trishume/syntect/compare/v4.7.0...v4.7.1) (2022-01-xx)
+
+- Remove 'plist-load' feature again due to semver violation. [#403](https://github.com/trishume/syntect/pull/403)
+
 ## [Version 4.7.0](https://github.com/trishume/syntect/compare/v4.6.0...v4.7.0) (2021-12-25)
 
 - Lazy-load syntaxes to significantly improve startup time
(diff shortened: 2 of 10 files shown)


Enselic:
I can't think of any changes to this PR, and I have verified locally that this fixes the build error in both `cargo-expand` and `bat` (with `path = "...,"` in `Cargo.toml`) after also reproducing the build errors.

So I think we should go ahead and merge this and release v4.7.1. Worst case we have to yank v4.7.1 too and try again in v4.7.2.

A couple of words on the embarrassing mistake of forgetting the `pub` that this PR fixes. That it could happen at all is because
* that API is not used in bat (I regularly verify my changes against bat)
* that API is not used in any integration tests in syntect (there are tests for the method, but they are internal and thus have access to internal API)

The correct long-term fix to the `pub` mistake would be to move relevant syntect test cases to `./tests/...` so that the tests only have access to public API. Then mistakes like these can be detected at compile time. That will be an adventure for another time though...

trishume (approved):
Thanks this looks good!

trishume:
I released v4.7.1!

Enselic:
@trishume We need to yank v4.7.1 too :cry: 

While the code indeed compiles, it doesn't run, since the dump format has changed due to the lazy-loading changes:

```
% cargo expand --lib
memory allocation of 7881122326526296064 bytes failed
```

I can't believe I didn't actually detect this when I verified my change. I feel like the worst developer in history and like I'm ruining all the trust I have built up. I can only say I am terribly  sorry to waste everyone's time.

Right now I see no other option than having to bump major version. But let's not rush that release either to give some time to let the situation sink in and allow us think this through. Again.

Enselic:
To be clear: Projects using pure syntect and the assets that come with syntect should be fine. The problem is projects using bat, since bat ships its own versions of integrated syntaxes that now are binary-incompatible with the latest syntect release.

trishume:
Yup seems like a major version bump is necessary, I didn't think about the fact that changing the binary format requires a semver bump either.

I sent you an offer to be a crates.io maintainer, you can feel free to yank releases in the future, and to publish releases on your own if I'm non-responsive for let's say 4 months, although I don't expect that I would be completely outside of extreme scenarios.

PR #405, opened 2021-12-29
Use Rust Cache Action to speed up builds.

varungandhi-src:
With caching, the time for the build step goes down from 1m 12s to 7s, as dependencies don't need to be rebuilt.

Before: https://github.com/sourcegraph/syntect/runs/4657761926
After: https://github.com/sourcegraph/syntect/runs/4658377802

In the After build, you'll see that the CI takes 3m 37s total, so this is roughly a ~25% improvement. 😄

Mirrors https://github.com/sourcegraph/syntect/pull/6

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -49,6 +49,7 @@ jobs:
     - uses: actions/checkout@v2
       with:
         submodules: true
+    - uses: Swatinem/rust-cache@v1
     - name: Build
       run: |
         cargo build


.github/workflows/CI.yml
```diff
@@ -49,6 +49,7 @@ jobs:
     - uses: actions/checkout@v2
       with:
         submodules: true
+    - uses: Swatinem/rust-cache@v1
```
Enselic, line 52: Thanks a lot for taking the time to make a contribution!

My spontaneous reaction to this PR is with skepticism, because it makes CI builds harder to debug and simulate locally. I don't find the CI build times terrible. It's just a few minutes.

There is also the aspect of security. An attacker that gets control over the cache can inject malicious code.

So I vote for not merging this.

varungandhi-src: That's a fair point. I noticed that [many popular repos are using this Action](https://sourcegraph.com/search?q=context:global+uses:+Swatinem/rust-cache%40v1&patternType=literal), so I figured it should be okay to rely on it, both from a security and correctness POV.

What do you think @trishume?

Enselic: Interesting. Some high profile projects are using it. Seems like this action is something to keep an eye on. For now I think we don't need it though.

PR #406, opened 2022-01-03
For 5.0.0: Rename `HighlightLines::highlight()` to `highlight_line()`

### LINKED ISSUE #314: Rename `HighlightLines::highlight()` to `highlight_line()` for clarity
With cobalt when I have the following code block
```python
def f(x):
  pass

def g(x):
  pass
```
it gets highlighted as:
screen

See https://github.com/cobalt-org/cobalt.rs/issues/802

If I switch from `regex-fancy` to `regex-onig`, it works.  

Unsure if this is a dupe of #287

Enselic:
To make it clear that the function takes one line at a time.

This require a semver major bump, but we need to do that anyway due to the
syntax lazy-loading changes.

Fixes #314

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -80,7 +80,7 @@ let syntax = ps.find_syntax_by_extension("rs").unwrap();
 let mut h = HighlightLines::new(syntax, &ts.themes["base16-ocean.dark"]);
 let s = "pub struct Wow { hi: u64 }\nfn blah() -> u64 {}";
 for line in LinesWithEndings::from(s) {
-    let ranges: Vec<(Style, &str)> = h.highlight(line, &ps);
+    let ranges: Vec<(Style, &str)> = h.highlight_line(line, &ps);
     let escaped = as_24_bit_terminal_escaped(&ranges[..], true);
     print!("{}", escaped);
 }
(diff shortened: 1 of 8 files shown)


Enselic:
@epage Would you mind taking a look at this fix for #314 since you filed that issue? Big thanks in advance.

epage:
Looks good.  Thanks!

Enselic:
Big thanks for taking a look.

(I will move out the CHANGELOG.md changes into a separate PR to avoid conflicts with other upcoming PRs.)

trishume:
Can you change this to a deprecation of the old name? I know it's a major version bump so we can just remove it, but this is the biggest API change and deprecations are way friendlier then renames given that the compile error can directly tell the user how to rename their call. It'll make upgrading to the new major version much smoother.

Maybe the warning can even be like "highlight was renamed to highlight_line to make it clear it should be passed a single line at a time"

Enselic:
Sure, no problem at all. I have pushed a commit to fix that.

Here is how the deprecation message looks in the wild. (Generated by temporarily using the deprecated variant in our own example code.)

```
warning: use of deprecated associated function `syntect::easy::HighlightLines::<'a>::highlight`: Renamed to `highlight_line` to make it clear it should be passed a single line at a time
   --> examples/syncat.rs:106:83
    |
106 |                     let regions: Vec<(Style, &str)> = highlighter.highlight_lines.highlight(&line, &ss);
    |                                                                                   ^^^^^^^^^
    |
    = note: `#[warn(deprecated)]` on by default
```

trishume (approved):

PR #410, opened 2022-01-03
Fix all current lint with Rust 1.57

Enselic:
I prefer to keep lints at a minimum to make it easy to see when new ones are introduced.

I would be happy to upgrade our CI check to reject lint warnings and not only lint errors like we have today, if that would be OK.

### DIFF
--- a/examples/syncat.rs
+++ b/examples/syncat.rs
@@ -82,7 +82,7 @@ fn main() {
             .unwrap_or_else(|| "base16-ocean.dark".to_string());
 
         let theme = ts.themes.get(&theme_file)
-            .map(|t| Cow::Borrowed(t))
+            .map(Cow::Borrowed)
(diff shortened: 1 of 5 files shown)


sharkdp (approved):

trishume (approved):

trishume:
I'd rather not have CI reject lint warnings, as not all contributors have them and I like being able to approve most PRs without going through a round-trip or doing minor tweaks myself. Also wouldn't that make it so that if new lints are added to clippy our CI starts failing?

I'm happy to accept PRs that fix up lints later though.

Enselic:
> I'd rather not have CI reject lint warnings

That's perfectly fine, just checking 👍 

> Also wouldn't that make it so that if new lints are added to clippy our CI starts failing?

Yes, if we were to use the stable toolchain. The easiest way to solve that is to run clippy with the MSRV toolchain. That way, new lints are only added when MSRV is manually bumped.

PR #411, opened 2022-01-03
For 5.0.0: Remove all deprecated items

Enselic:


trishume (changes requested):
I like removing the theme attributes, I think nobody uses those and they affect the binary format and clutter the docs.

I'd rather keep the deprecated html functions in case it's been a while since someone worked on a syntect project, it's just much friendlier. I really wish there was something like a deprecation attribute except it produced a build error instead of a warning.

One other option to keeping them is to keep them deprecated, but create a new empty enum type named `Deprecated` or something, and change those functions to take `Deprecated` as an argument, delete their bodies, and maybe also add `#[doc(hidden)]. Then check that when they're called they produce both a type error and a deprecation warning.

Enselic:
If you want to keep them, we'll keep them :) I restored the original code in `src/html.rs` now.

trishume (approved):

PR #414, opened 2022-01-07
Bump `plist` to 1.3.0

### LINKED ISSUE #385: RUSTSEC-2020-0159
IIUC, projects using syntect are currently affected by [RUSTSEC-2020-0159](https://rustsec.org/advisories/RUSTSEC-2020-0159). This is because syntect depends on the plist crate, and plist depends on chrono, and chrono contains the security problem. There is an unreleased fix for this in plist: https://github.com/ebarnard/rust-plist/pull/72 so IIUC if that were released, and syntect were to update to it, then that would fix the issue. There is also current discussion in the chrono repo about fixing the situation there: https://github.com/chronotope/chrono/issues/499. cc @sharkdp 

(Or is there a feature-based workaround for now?)

Enselic:
The commit where `plist` stops depending on `chrono` is https://github.com/ebarnard/rust-plist/commit/9ed64da1cc52ba766ae9ba84eefa6edffc31c90a, which was released in 1.3.0. So set that as the minimum version.

As a reminder, here is how cargo interprets semver strings: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html

Closes #385

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ walkdir = "2.0"
 regex-syntax = { version = "0.6", optional = true }
 lazy_static = "1.0"
 bitflags = "1.0.4"
-plist = "1"
+plist = "1.3"
 bincode = { version = "1.0", optional = true }
 flate2 = { version = "1.0", optional = true }
 fnv = { version = "1.0", optional = true }


Enselic:
Confirmed to fix the RUSTSEC: https://github.com/trishume/syntect/issues/385#issuecomment-1007681500

PR #417, opened 2022-01-07
For 5.0.0: Split up feature `assets` into `default-syntaxes` and `default-themes`

### LINKED ISSUE #310: Separate feature for theme/syntax bundling
If someone writes a 16-color application, or for some other reason don't want to have bundled themes, it would be good for them to be able to remove the bundled themes without losing the bundled syntax definitions.

I might have missed something, but I don't think you currently can disable the embedding of themes.

Enselic:
Also specify hard dependencies directly in Cargo.toml instead of asking users to manually enable them. This allows us to simplify cfg directives.

Fixes #310

This will conflict with #407, but that should not affect code review, and I prefer getting this reviewed early over not having to sort out conflicts. The conflicts shouldn't be that bad.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -73,7 +73,3 @@ jobs:
     - name: Run tests without default features
       run: |
         cargo test --lib --no-default-features
-    - name: Run Xi tests
-      run: |
-        # Test the build configuration that Xi uses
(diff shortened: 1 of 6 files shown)


Enselic:
Reporter OK:ed this in https://github.com/trishume/syntect/issues/310#issuecomment-1007855783

trishume (approved):

Issue #323, opened 2021-01-05
Support changes to sublime-syntax from Sublime Text `Build 4075`


jrappen:
#### Changes from `Build 4075`

* [x] hidden file extensions
    * [x] `hidden_file_extensions`(addressed by #419)
* [x] sublime-syntax v2
    * [x] `version: 2`
        * [x] set + meta_content_scope (addressed by #610)
        * [x] multiple push + clear_scopes (addressed by #610)
        * [x] embed_scope replacement (addressed by #612)
        * [x] set + clear_scopes (addressed by #612)
        * [x] embed escape meta scopes (addressed by #612)
* [x] inheritance
    * [x] `extends` (string value) (addressed by #610)
    * [x] `extends` (string array value where all syntaxes themselves extend from the same base) (addressed by #612)
* [x] meta patterns:
    * [x] `meta_prepend` (addressed by #610)
    * [x] `meta_append` (addressed by #610)
* [x] match patterns:
    * [x] pop multiple contexts via `pop: 2` etc.
* [x] including an external prototype (addressed by #610)

#### Documentation

<https://www.sublimetext.com/docs/syntax.html#ver-dev>

stefanobaghino:
It's quite nice to see some progress happening with #536! Thanks @jalil-salame!

I'm considering helping. Is there other work going on concerning this ticket apart from the aforementioned draft PR?

jalil-salame:
> It's quite nice to see some progress happening with #536! Thanks @jalil-salame!
> 
> I'm considering helping. Is there other work going on concerning this ticket apart from the aforementioned draft PR?

I need to figure out why the tests are failing in #535 (this updates sublime packages to a newer version) before I can open #536 for review. If you check the comments on #535 you'll see where I'm stuck (I also don't have much time so helping hands will be useful).

Keats:
Exciting, you're killing it @jalil-salame !

stefanobaghino:
Since #536 seems to be stuck, I started having a look with #610.

stefanobaghino:
Work continuing in https://github.com/trishume/syntect/pull/612 -- @jrappen re: including an external prototype -- isn't it handled by #610 already?

stefanobaghino:
@jrappen I had a second look and I'm under the impression that "embed escape meta scopes" is also handled by #612, meaning that that would close this issue. Does that make sense?

jrappen:
[...]

PR #419, opened 2022-01-16
Add support for hidden_file_extensions key.

varungandhi-src:
This was introduced in Sublime Text 4.

Addresses one part of https://github.com/trishume/syntect/issues/323.

As I understand it, it doesn't really matter if a file extension is "hidden" or not from the POV of syntax highlighting.

### DIFF
--- a/src/parsing/yaml_load.rs
+++ b/src/parsing/yaml_load.rs
@@ -135,14 +135,17 @@ impl SyntaxDefinition {
             top_level_scope,
         );
 
+        let mut file_extensions = Vec::new();
+        for extension_key in &["file_extensions", "hidden_file_extensions"] {
+            if let Ok(v) = get_key(h, extension_key, |x| x.as_vec()) {
+                file_extensions.extend(v.iter().filter_map(|y| y.as_str().map(|s| s.to_owned())))
+            }
+        }
+
         let defn = SyntaxDefinition {
             name: get_key(h, "name", |x| x.as_str()).unwrap_or_else(|_| fallback_name.unwrap_or("Unnamed")).to_owned(),
             scope: top_level_scope,
-            file_extensions: {
-                get_key(h, "file_extensions", |x| x.as_vec())
-                    .map(|v| v.iter().filter_map(|y| y.as_str()).map(|x| x.to_owned()).collect())
-                    .unwrap_or_else(|_| Vec::new())
-            },
+            file_extensions,
             // TODO maybe cache a compiled version of this Regex
             first_line_match: get_key(h, "first_line_match", |x| x.as_str())
                 .ok()
@@ -855,6 +858,7 @@ mod tests {
         name: C
         scope: source.c
         file_extensions: [c, h]
+        hidden_file_extensions: [k, l]
         hidden: true
         variables:
           ident: '[QY]+'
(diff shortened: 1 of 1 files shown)


varungandhi-src:
Is anyone actively working on fixing 323/adding support for the other new features? I didn't see any active PRs related to it, and no one had commented on the issue, but I figured I'd ask first.

keith-hall (approved):

keith-hall:
> Is anyone actively working on fixing 323/adding support for the other new features?

Not that I know of, no. Iirc I wrote a comment asking for advice on how the API might look like, but had no responses. I did have a quick look at implementing support for popping multiple contexts at once but it wasn't trivial with the current design.

trishume (approved):
I don't think anyone else is. However note that the real trickiness will be the backtracking support, which I believe will require a complete API overhaul, since currently you can parse in a streaming way since no line can affect a previous line, whereas with backtracking that's no longer true.

I think it might be reasonable to support everything except backtracking though, since there'll be lots of syntaxes which might use some new features but not backtracking.

trishume:
Oh, see the minimum supported rust version CI error, you just need to borrow the array you're iterating.

varungandhi-src:
CI error should be fixed with the latest push.

As for backtracking, thanks for bringing that up. I'd only seen issue 323 and not issue 271, so I hadn't realized that was another potential pain point before bumping to ST4. It looks like a bunch of grammars are [already using the backtracking-related features](https://github.com/trishume/syntect/issues/271#issuecomment-1014137293):

```
Batch File/Batch File.sublime-syntax
C#/C#.sublime-syntax
Git Formats/Git Commit.sublime-syntax
Java/Java.sublime-syntax
JavaScript/JSX.sublime-syntax
JavaScript/JavaScript.sublime-syntax
JavaScript/TSX.sublime-syntax
JavaScript/TypeScript.sublime-syntax
Lua/Lua.sublime-syntax
Markdown/Markdown.sublime-syntax
PHP/PHP Source.sublime-syntax
Python/Python.sublime-syntax
SQL/SQL.sublime-syntax
```

Apart from backtracking, it sounds like adding support for the 4075-specific features, such as syntax inheritance, is not as complicated and shouldn't affect the public API -- is that a fair assessment?

jrappen:
> ... grammars which are using `branch:` ...

also add these pending re-writes from PRs:

- Haskell
- JSON

varungandhi-src:
Could someone please kick-off the PR testing again? The clippy issue should be fixed, and I'm seeing a "First-time contributors need a maintainer to approve running workflows" message.

trishume:
Huh I tried re-running it and the clippy issue happened again. Did you already look into it? Looks like the problem is that the time crate updated past our MSRV, which of course isn't a problem with this PR. Perhaps we need to either bump our MSRV or constrain an older version of the time crate.

Enselic:
I created a PR to bump MSRV along with a motivation of why I believe that is preferred over putting an upper limit on `time`: https://github.com/trishume/syntect/pull/422

varungandhi-src:
Thanks @Enselic, we can retest and land this PR after your PR lands.

Enselic:
@varungandhi-src Could you rebase this so we get a green CI please?

varungandhi-src:
@Enselic rebased and force-pushed.

Enselic:
Since this has been Approved by both Tristan and Keith I will merge this now that CI is green. I took a quick look and the code looked good to me too.

PR #420, opened 2022-01-17
Implement `Error` and `Display` for all error enums by using `thiserror`

### LINKED ISSUE #94: Implement Error trait for error types

Enselic:
Also do some minor tweaks to the way errors are represented, and add basic integration tests for errors, to give a feel for how the errors behave. In general the backwards compatibility for clients should be good. As the [docs](https://docs.rs/thiserror/latest/thiserror/) for `thiserror` states (emphasis mine):

> Thiserror deliberately does not appear in your public API. You get the same thing as if you had written an implementation of std::error::Error by hand, and **switching from handwritten impls to thiserror or vice versa is not a breaking change.**

I.e. the code I remove should be implemented by the new `Error` derive.

Also mark all error enums as `#[non_exhaustive]` so we can add new error variants without semver breakage in the future.

Fixes #94

FWIW, `bat` has been using `thiserror` for [5 months](https://github.com/sharkdp/bat/pull/1820) without any issues so far.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,6 +33,7 @@ serde = "1.0"
 serde_derive = "1.0"
 serde_json = "1.0"
 once_cell = "1.8"
+thiserror = "1.0"
 
 [dev-dependencies]
 criterion = { version = "0.3", features = [ "html_reports" ] }
--- a/src/highlighting/settings.rs
+++ b/src/highlighting/settings.rs
@@ -13,9 +13,10 @@ pub trait ParseSettings: Sized {
 }
 
 /// An error parsing a settings file
-#[derive(Debug)]
+#[derive(Debug, thiserror::Error)]
 pub enum SettingsError {
     /// Incorrect Plist syntax
+    #[error("Incorrect Plist syntax: {0}")]
     Plist(PlistError),
 }
 
--- a/src/highlighting/settings.rs
+++ b/src/highlighting/settings.rs
@@ -14,6 +14,7 @@ pub trait ParseSettings: Sized {
 
 /// An error parsing a settings file
 #[derive(Debug, thiserror::Error)]
+#[non_exhaustive]
 pub enum SettingsError {
     /// Incorrect Plist syntax
     #[error("Incorrect Plist syntax: {0}")]
--- a/src/highlighting/theme_load.rs
+++ b/src/highlighting/theme_load.rs
@@ -11,27 +11,34 @@ use crate::parsing::ParseScopeError;
 
 use self::ParseThemeError::*;
 
-#[derive(Debug)]
+#[derive(Debug, thiserror::Error)]
 pub enum ParseThemeError {
+    #[error("Incorrect underline option")]
     IncorrectUnderlineOption,
+    #[error("Incorrect font style: {0}")]
     IncorrectFontStyle(String),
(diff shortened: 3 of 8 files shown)


Enselic:
@trishume Hi, just a heads-up that this soon has been open for more than a month, and as  per https://github.com/trishume/syntect/pull/382#issuecomment-950443083 I plan to merge this early next week.

Let me know if you want more time for review, that is no problem at all.

I am comfortable to merge this without review though because I consider it "obviously an improvement", and the changes to the public API are small, and when there is a change, it is obvious for clients what to do.

In fact, I've been working on a [tool](https://github.com/Enselic/public_items) lately to diff changes to public APIs of Rust libraries. Here are the exact changes this PR does to the public API:

```
cargo install cargo-public-items

git checkout origin/master
cargo public-items > /tmp/before-PR

git checkout PR
cargo public-items > /tmp/after-PR

diff -U0 /tmp/before-PR /tmp/after-PR | grep -v @@
```

which gives this output (manually cleaned up a bit)

```
-pub enum variant syntect::LoadingError::ParseSyntax(ParseSyntaxError,Option<String>)
+pub enum variant syntect::LoadingError::ParseSyntax(ParseSyntaxError,String)

-pub fn syntect::LoadingError::cause(&self) -> Option<&Error>
+pub fn syntect::LoadingError::source(&self) -> std::option::Option<&std::error::Error + 'static>

+pub fn syntect::highlighting::ParseThemeError::source(&self) -> std::option::Option<&std::error::Error + 'static>
+pub fn syntect::highlighting::ParseThemeError::fmt(&self, __formatter: &mut std::fmt::Formatter<'_>) -> std::fmt::Result
+pub fn syntect::highlighting::ParseThemeError::to_string(&self) -> String

+pub fn syntect::highlighting::SettingsError::fmt(&self, __formatter: &mut std::fmt::Formatter<'_>) -> std::fmt::Result
+pub fn syntect::highlighting::SettingsError::to_string(&self) -> String

+pub fn syntect::parsing::ParseScopeError::fmt(&self, __formatter: &mut std::fmt::Formatter<'_>) -> std::fmt::Result
+pub fn syntect::parsing::ParseScopeError::to_string(&self) -> String

-pub fn syntect::parsing::ParseSyntaxError::cause(&self) -> Option<&Error>
+pub fn syntect::parsing::ParseSyntaxError::source(&self) -> std::option::Option<&std::error::Error + 'static>

+pub fn syntect::parsing::ParseSyntaxError::from(source: ScanError) -> Self
```

each line represents a public item in the syntect API. - means "removed" and + means "added".

So the deprecated `Error::cause` has been replaced with `Error::source`, an `Option<String>` has been changed into an `String`. Apart from that there are only additions to the API (implementations of `Display` and `Error`).

Disclaimer: The tool is still in progress so I would not trust my life with it, but it gives a good indication at least.

trishume (approved):
Great PR! Sorry for leaving it for so long, this would indeed have been totally fine to merge on your own after a month. I moved cities and started a new job this month so I've been busy.

PR #427, opened 2022-03-12
Fall back to 'Plain Text' if a referenced syntax is missing

### LINKED ISSUE #421: panic on a svelte file
Here's the file:
```svelte
<script>
	import Navigation from '$lib/pages/Navigation.svelte';

	import '../app.postcss';
</script>

<Navigation />

<slot />

<style global lang="postcss">
	body,
	#svelte {
		@apply w-full h-full;
		@apply bg-neutral-100;
	}
</style>
```
It breaks both bat and broot, both using syntect 4.6.0.

image

```
thread 'main' panicked at 'Can only call resolve on linked references: ByScope { scope: <source.postcss>, sub_context: None }', /home/dys/.cargo/registry/src/github.com-1ecc6299db9ec823/syntect-4.6.0/src/parsing/syntax_definition.rs:191:18
stack backtrace:
   0: rust_begin_unwind
             at /rustc/f1edd0429582dd29cccacaf50fd134b05593bd9c/library/std/src/panicking.rs:517:5
   1: std::panicking::begin_panic_fmt
             at /rustc/f1edd0429582dd29cccacaf50fd134b05593bd9c/library/std/src/panicking.rs:460:5
   2: syntect::parsing::parser::ParseState::push_meta_ops
   3: <bat::printer::InteractivePrinter as bat::printer::Printer>::print_line
   4: bat::controller::Controller::print_file
   5: bat::controller::Controller::run_with_error_handler
   6: bat::run_controller
   7: bat::run
   8: bat::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

Enselic:
The consequence of this change is that if `syntect` encounters a syntax
that references another syntax that is not present, `syntect` will
fallback to Plain Text syntax instead of erroring/panicking. Falling
back to Plain Text in cases like this seems to be what Sublime Text is
doing too.

For example, `bat` has a syntax for Vue, but Vue references a syntax
with scope `text.pug`, which is not included in bat (by default).

So if `bat` encounters the following `vue-with-pug.vue` file:
```
<template lang='pug'>
  #container.col
    p This shall be formated as Plain Text as long as a Pug syntax definition is missing
</template>
```
`bat` currently panics with:

```
% bat ~/Desktop/vue-with-pug.vue
───────┬────────────────────────────────────────────────────────────────
       │ File: /Users/martin/Desktop/vue-with-pug.vue
       │ Size: 140 B
───────┼────────────────────────────────────────────────────────────────
thread 'main' panicked at 'Can only call resolve on linked references: ByScope { scope: <text.pug>, sub_context: None }', /Users/martin/.cargo/registry/src/github.com-1ecc6299db9ec823/syntect-4.6.0/src/parsing/syntax_definition.rs:191:18
```

With this change, instead of panicking, `bat` will print the file with
the pug parts highlighted as Plain Text:

Skärmavbild 2022-03-12 kl  08 03 16

Closes #421

### DIFF
--- a/src/parsing/syntax_set.rs
+++ b/src/parsing/syntax_set.rs
@@ -1024,10 +1024,16 @@ mod tests {
         let syntax = syntax_set.find_syntax_by_extension("z").unwrap();
         let mut parse_state = ParseState::new(syntax);
         let ops = parse_state.parse_line("z go_x x leave_x z", &syntax_set);
-        let expected_text_plain = (6, ScopeStackOp::Push(Scope::new("text.plain").unwrap()));
-        assert_ops_contain(&ops, &expected_text_plain);
-        let expected_text_plain_pop = (9, ScopeStackOp::Pop(1));
-        assert_ops_contain(&ops, &expected_text_plain_pop);
+        let expected_ops = vec![
+            (0, ScopeStackOp::Push(Scope::new("source.z").unwrap())),
+            (0, ScopeStackOp::Push(Scope::new("z").unwrap())),
+            (1, ScopeStackOp::Pop(1)),
+            (6, ScopeStackOp::Push(Scope::new("text.plain").unwrap())),
+            (9, ScopeStackOp::Pop(1)),
+            (17, ScopeStackOp::Push(Scope::new("z").unwrap())),
+            (18, ScopeStackOp::Pop(1)),
+        ];
+        assert_eq!(ops, expected_ops);
     }
 
     #[test]
--- a/src/parsing/syntax_set.rs
+++ b/src/parsing/syntax_set.rs
@@ -749,15 +749,15 @@ impl SyntaxSetBuilder {
                     all_context_ids[syntax_index].get(s)
                 }
             }
-            ByScope { scope, ref sub_context } => {
-                Self::find_id(sub_context, all_context_ids, syntaxes, |index_and_syntax| {
+            ByScope { scope, ref sub_context, with_escape } => {
+                Self::with_plain_text_fallback(all_context_ids, syntaxes, with_escape, Self::find_id(sub_context, all_context_ids, syntaxes, |index_and_syntax| {
                     index_and_syntax.1.scope == scope
-                })
+                }))
             }
-            File { ref name, ref sub_context } => {
-                Self::find_id(sub_context, all_context_ids, syntaxes, |index_and_syntax| {
+            File { ref name, ref sub_context, with_escape } => {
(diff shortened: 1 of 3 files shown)


src/parsing/syntax_set.rs
```diff
@@ -767,6 +767,32 @@ impl SyntaxSetBuilder {
         }
     }
 
+    fn with_plain_text_fallback<'a>(
```
Enselic, line 770: I tried putting this logic inside of `Self::find_id()` directly, which would simplify the code. But rustc complained about recursive instantiation of generics. This is because the `predicate` parameter is generic, and calling `find_id()` inside of itself causes recursion without a base case. This problem is explained in detail [here](https://stackoverflow.com/a/54617718/287761).

This was the most straightforward way I could come up with to work around that.

src/parsing/syntax_set.rs
```diff
@@ -954,6 +980,56 @@ mod tests {
...
+        let syntax_set = builder.build();
+
+        let syntax = syntax_set.find_syntax_by_extension("z").unwrap();
+        let mut parse_state = ParseState::new(syntax);
+        let ops = parse_state.parse_line("z go_x x leave_x z", &syntax_set);
+        let expected_text_plain = (6, ScopeStackOp::Push(Scope::new("text.plain").unwrap()));
+        assert_ops_contain(&ops, &expected_text_plain);
+        let expected_text_plain_pop = (9, ScopeStackOp::Pop(1));
+        assert_ops_contain(&ops, &expected_text_plain_pop);
+    }
```
keith-hall, line 1031: Would it also make sense to check the scope of the final `z`, just to be super sure that when the plain text embed substitute is popped off the stack, that it continues parsing as expected?

keith-hall (approved):

src/parsing/syntax_set.rs
Enselic: Big thanks for code review! I agree it makes sense to be more strict when asserting. So I changed from "contains" to "is exactly", which I just pushed a commit for. It looks like it is correct to me, but my experience in interpreting ops is limited, so would be great if you could also make a judgement.

src/parsing/syntax_set.rs
keith-hall: I was wondering about that, thanks for the explanation :)

src/parsing/syntax_set.rs
keith-hall: looks good! thanks

trishume (approved):

Issue #428, opened 2022-03-12
4.7 is not uploaded to crates.io


rhysd:
Hi,

I'm very interested in v4.7 changes described in [the changelog](https://github.com/trishume/syntect/blob/master/CHANGELOG.md). However the latest version uploaded to crates.io is [v4.6.0](https://crates.io/crates/syntect/versions).

I think there were some problems on the latest version since they were once uploaded then unpublished. Is there a plan to publish a new version for v4.7?

Enselic:
The lazy-loading of syntaxes changed the binary format of assets, which constitutes a semver breaking change. So the next release will be 5.0.0. Since we had to do a major bump, we wanted to take the opportunity to adjust the API a bit. The only thing left do do is solving https://github.com/trishume/syntect/issues/98 which is done in multiple increments, and there is an open PR awaiting review, but it is not the final PR

The best way to follow progress is subscribing to https://github.com/trishume/syntect/pull/409 (which will be updated before 5.0.0)

rhysd:
Thank you so much for sharing the situation. I understood and will watch the progress.

Enselic:
Hi rhysd-san, just wanted to ping you about that 5.0.0 has been published on crates.io now since a few days back, in case you didn't notice: https://github.com/trishume/syntect/releases/tag/v5.0.0

rhysd:
@Enselic Thank you for the notification. And congrats on the new major version! I'll try it soon.

PR #429, opened 2022-03-13
Fix misreading v4.7 information in documents

rhysd:
I confirmed that v4.7.0 and v4.7.1 were unpublished due to the semver violation issue at #428.

However the readme file says v4.7 should be specified on installing syntect. This would be misreading since it causes an error that no such version exists on crates.io. And noting that the versions were unpublished in changelog would be better so that users can know that these versions were once released but not available on crates.io.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 - Remove 'plist-load' feature again due to semver violation. [#403](https://github.com/trishume/syntect/pull/403)
 
+This version was removed from crates.io due to the semver violation issue.
+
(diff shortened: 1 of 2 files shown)


Enselic (approved):
Thank you for the PR, and sorry for not doing this change before. When the 4.7.0 and 4.7.1 release were topical, it was not known it would take several months until we could make 5.0.0 happen, so I never got around to make this kind of change. Sorry for that.

I will give @trishume at least a few days to review before I merge this, but I don't think it makes sense to wait longer than that, since this clarification is somewhat urgent.

If you think I am crossing a line here @trishume, don't hesitate to let me know. In that case I will never think such forbidden thoughts again ;)

Issue #168, opened 2018-05-29
Curated list of languages


vors:
Extracting from #113 into its own issue.

Many people expressed a desire to add a support for new languages.
The **current story is unclear** becase the list of syntaxes is coming from https://github.com/sublimehq/Packages/ and its README says

> Pull requests for new packages won't be accepted at this stage, as new packages can cause issues for users who have a package with the same name installed via Package Control. There are some planned changes that will address this in the future.

Decouping list of supported languages from `sublimehq/Packages` would allow to move forward.
List of already discussed options:

- Take dependency on some other currated list of syntaxes, i.e https://github.com/Keats/gutenberg/tree/master/sublime_syntaxes or https://github.com/github/linguist/tree/master/vendor/grammars
- create a new project for a list of curated grammar files and consume it from `syntect`

In https://github.com/trishume/syntect/pull/113#issuecomment-389018043 @trishume said

> If someone's willing to do the work to automatically convert all those tmBundles to sublime-syntaxes and add a cargo feature to bundle the extra syntaxes, I'm probably willing to accept that. Might need to set up Git LFS for the packdumps for that so they don't bloat the Git repo too much, but that's fine.

> I'm totally okay with some curation though, especially since we can use sublime-synaxes I bet there are some higher quality syntaxes for certain languages than the ones Github uses.

I'm opening this issue in a hope to continue this conversation and start some work.

sharkdp:
I would definitely be interested in this as well. I currently also maintain a (small) list of syntaxes for [bat](https://github.com/sharkdp/bat): [`bat/assets/syntaxes`](https://github.com/sharkdp/bat/tree/master/assets/syntaxes).

I think a separate repository might be useful such that the `syntect` repository will not be polluted with issues/PRs for new syntaxes.

Concerning linguist, note that a lot of these are `atom` syntaxes. Still, this could be a useful resource.

Regarding logistics, I think Git submodules are a great way of bundling different repositories since we keep the link to the upstream source. Unfortunately, a lot of Sublime Syntax repositories do not contain a `.sublime-syntax` file. Ideally, I think, the `.tmBundle` => `.sublime-syntax` conversion should be done by some script when the syntax-bundle repository is "built".

sharkdp:
Also, I just found [this fork of sublimehq/Packages](https://github.com/slimsag/Packages) which is used by [`syntect_server`](https://github.com/sourcegraph/syntect_server#adding-languages). It has a lot more syntaxes than the default `Packages` repo.

[...]

PR #459, opened 2022-11-17
mlir: mlir is currently not syntax highlighted

artagnon:
It seems that sublimehq/packages doesn't include syntax highlighting for MLIR files. Worth noting that they're syntax-highlighted on the GitHub interface, as well as on VSCode, with the MLIR package.


Enselic:
Thank you for taking the time to open a PR. Unfortunately we have decided to not maintain syntaxes bundled with syntect. Clients of syntect needs to take of that themselves. See https://github.com/trishume/syntect/issues/168 for some discussion regarding this.

Closing.

PR #464, opened 2023-01-31
Add languages example.

klnusbaum:
I was trying to figure out all of the languages that are supported and their supported file extensions. I couldn't find any documentation so I wrote this simple program to print everything out. I'm on the fence about whether or not it would be useful to others. Thought I'd put up a PR and discuss.


Enselic:
Thank you for taking the time to make a PR. I think this already exists though? Have you tried `cargo run --example=syncat -- -l`?

klnusbaum:
Oh, cool! My bad. Glad this already exists :)

PR #476, opened 2023-05-02
Update `regex-syntax`

reknih:
This PR removes a duplicate `regex-syntax` dependency. `syntect` previously depended on the crate both directly (0.6) and indirectly through `fancy-regex` (0.7), resulting in a dependency on both versions. This PR updates `regex-syntax` to 0.7 everywhere.


Enselic (approved):

jackpot51:
Is there anything else needed before this is merged?

Enselic:
I don't think so. Let's merge.

jackpot51:
@Enselic does this mean that @trishume has been unresponsive for two months already?

PR #479, opened 2023-05-24
implement PartialEq for Theme

CosmicHorrorDev:
I have a project where I have unit tests comparing structs that hold a syntect `Theme` that currently use an ugly hack to work around `Theme` missing `PartialEq`

This just adds a derive for `Theme` and all the inner structs that also need it in turn. It would probably be good for all publicly exposed structs/enums to implement the traits listed in the [`C-COMMON-TRAITS` API guidelines](https://rust-lang.github.io/api-guidelines/interoperability.html#c-common-traits), but that seemed like a pretty sweeping change to try and make


Enselic (approved):
Makes sense to me 👍

PR #483, opened 2023-07-04
Automate checking 'last 3 Rust versions' on CI

rhysd:
This PR replaces unmaintained [actions/toolchain](https://github.com/actions-rs/toolchain) with [dtolnay/rust-toolchain](https://github.com/dtolnay/rust-toolchain). It is well-maintained and supports 'stable minus N releases' syntax to specify toolchain version so that we don't need to maintain the `MIN_SUPPORTED_RUST_VERSION` variable manually. In addition, I updated `actions/checkout` since v2 seems no longer maintained in favor of v3.

I confirmed all jobs passed by this PR: https://github.com/rhysd/syntect/actions/runs/5453682953

Current master branch failed since Rust version was too old for some dependencies.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -15,7 +15,7 @@ jobs:
     name: Minimum supported rust version
     runs-on: ubuntu-20.04
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
     - uses: dtolnay/rust-toolchain@master
       with:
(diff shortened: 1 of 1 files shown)


Enselic (approved):
Looks like an improvement to me 👍

Enselic:
This PR makes CI more green than [running from master](https://github.com/trishume/syntect/actions/runs/5620735627), so let's merge.

PR #487, opened 2023-07-30
Bump to 5.1.0

Enselic:
Draft of release notes:

### Improvements
* Fix how `syntect::util::split_at()` handles multi-byte characters
* Allow case insensitive search for .tmtheme paths
* impl `PartialEq` for `Theme`

### Dependencies
* Upgrade `fancy-regex` to 0.11
* Upgrade `regex-syntax` to 0.7
* Replace `lazy_static` with `once_cell`

### Other
* Change MSRV policy to "last three stable versions"
* Make `Debug` impl of `syntect::highlighting::Color` less verbose

(I have confirmed that this version of syntect passes all bat regression tests: https://github.com/sharkdp/bat/pull/2632)


Enselic:
Note: Strictly speaking this should be 6.0.0 because we remove public API, but we have previously decided to take the risk. See https://github.com/trishume/syntect/pull/438#pullrequestreview-1035167525 and the rest of the discussion for details.

trishume (approved):
Thanks for putting this together!

trishume:
How about you try doing the release yourself and let me know if you run into any permissions snags.

Enselic:
Sounds good! Will try.

Enselic:
[Went smoothly!](https://crates.io/crates/syntect/5.1.0) (knock on wood)

PR #493, opened 2023-10-04
CI: Run syntax regression tests from bat

Enselic:
If bat master branch syntax regression tests work with the syntect code under test, that is a good indication that syntect work well.

Add such a check in CI.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -41,6 +41,28 @@ jobs:
     - uses: actions/checkout@v3
     - run: RUSTDOCFLAGS='--deny warnings' cargo doc --no-deps --document-private-items --all-features
 
+  bat-tests:
+    name: Run bat syntax regression tests
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+      with:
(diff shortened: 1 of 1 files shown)


keith-hall (approved):

.github/workflows/CI.yml
```diff
@@ -41,6 +41,28 @@ jobs:
...
+    - uses: actions/checkout@v3
+      with:
+        repository: 'sharkdp/bat'
+        path: 'bat'
+        ref: master
+        submodules: true
+    - name: bat/tests/syntax-tests/regression_test.sh
+      run: |
+        cd bat
+        sed -i 's%\[dependencies.syntect\]%[dependencies.syntect]\npath = "../syntect"%' Cargo.toml
```
keith-hall, line 60: It could be more robust to use a tool designed to edit toml, for example https://github.com/gnprice/toml-cli

Enselic: Interesting! Maybe we can start using if it the current sed commands begins to fail?

keith-hall: Fine by me 🙂

PR #499, opened 2023-11-27
Update `regex-syntax` to 0.8.

CGMossa:
I have updated the `regex-syntax` dependencies to 0.8.
I don't have a way to test this further than `cargo test --no-fail-fast`, which showed the same tests failing on `master` as on this. So it is the same in that regard.

PR #490, opened 2023-08-29
update bitflags

extrawurst:
closes #489

PR #500, opened 2023-11-27
fix `clippy` comments

CGMossa:
While working on another PR #499 , I ran `cargo clippy` to get familiar with the code-base. 

Depends on

- [x] https://github.com/Enselic/cargo-public-api/pull/518
- [x] https://github.com/trishume/syntect/pull/503

### DIFF
--- a/src/parsing/syntax_set.rs
+++ b/src/parsing/syntax_set.rs
@@ -357,11 +357,11 @@ impl SyntaxSet {
         let syntax = &self
             .syntaxes
             .get(context_id.syntax_index)
-            .ok_or_else(|| ParsingError::MissingContext(*context_id))?;
+            .ok_or(ParsingError::MissingContext(*context_id))?;
         syntax
             .contexts()
             .get(context_id.context_index)
-            .ok_or_else(|| ParsingError::MissingContext(*context_id))
+            .ok_or(ParsingError::MissingContext(*context_id))
     }
 
     fn first_line_cache(&self) -> &FirstLineCache {
@@ -670,7 +670,7 @@ impl SyntaxSetBuilder {
     /// Anything recursively included by the prototype shouldn't include the prototype.
     /// This marks them as such.
     fn recursively_mark_no_prototype(
-        syntax: &SyntaxReference,
+        _syntax: &SyntaxReference,
         context_id: &ContextId,
         syntax_context_ids: &HashMap<String, ContextId>,
         all_contexts: &[Vec<Context>],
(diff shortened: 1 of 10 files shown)


Enselic:
Thanks! Can you rebase on latest master so CI passes please?

Enselic:
The CI failure should be fixed by https://github.com/Enselic/cargo-public-api/pull/516. The plan is:
1. The above fix is merged and a new version of the crate is released.
1. I upgrade syntect to the new version
1. We rebase this PR and then it should pass :)

Enselic:
Can you rebase again please? Now it should work :)

Enselic:
Sorry for the mess... my fix was buggy, here is a fix of the fix: https://github.com/Enselic/cargo-public-api/pull/518

This time I have also confirmed that the `public_api` test passes for this PR with the above fix.

Thanks for the tip with the merge queue, but
* it only works for organizations (or payed accounts)
* I actually tested it during the beta but it was too buggy for my taste

This repo is pretty low traffic, but if it becomes high traffic it could make sense to look into something that implements the It'ss Not Rocket Science Rule.

CGMossa:
No worries. I also learned something new here. I've added a "Depends on" in the description, so we'll know when the time is right.

src/parsing/syntax_set.rs
```diff
@@ -670,7 +670,7 @@ impl SyntaxSetBuilder {
     /// Anything recursively included by the prototype shouldn't include the prototype.
     /// This marks them as such.
     fn recursively_mark_no_prototype(
-        syntax: &SyntaxReference,
+        _syntax: &SyntaxReference,
```
Enselic, line 673: Can't we remove this parameter entirely if it is unused?

CGMossa: Oh yeah. I've pushed this now.

Enselic:
It feels awkward and embarrassing to ask, but can you rebase again please?

CGMossa:
No worries. Let's gooooooooo! 😄

CGMossa:
Could I ask you to make a release to crates-io of syntect at some point :D 
Selfishly, I pushed for #499 in order to upgrade typst :P packages

Enselic:
Thanks, let's merge!

Regarding making a release, right now what's on master requires us to release 6.0.0, which I'd like to avoid. So we should probably revert the [bump](https://github.com/trishume/syntect/pull/490) that changed the API in an backwards incompatible way. Better yet would probably be to remove external types from the public API so we have full control over when we change the API. But that could be a big and invasive changed, maybe not even doable.

If we revert bitflags and then go over all other deps that can be bumped, I think we can make a 5.x release to crates-io. Maybe you are interested in going over and bumping deps? Not minor bumps, just x.0.0 bumps and 0.x.0 bumps, since they prevent a single version of a dep to be used by dependents, if the dependents depend on other versions of dependencies.

CGMossa:
Hmm.. I'm not an expert honestly. I've tried bumping "plist", in #504 and I've got an error.
Plus I tried to revert 
```sh
git revert e9819fb                                       
error: commit e9819fbcaf0eb5ea73448079aac78d00b8b4d140 is a merge but no -m option was given.        
fatal: revert failed
```

which should be the merge commit for the PR you mentioned.

🤷

PR #506, opened 2023-12-10
Add parsing of gutterSettings

### LINKED ISSUE #469: Support for gutterSettings
syntect currently reads from `gutter` and `gutterForeground` when loading themes: https://github.com/trishume/syntect/blob/c61ce60c72d67ad4e3dd06d60ff3b13ef4d2698c/src/highlighting/theme_load.rs#L254-L257
But many themes instead store their gutter settings in a `gutterSettings` key. See for instance the base16 themes at https://github.com/chriskempson/base16-textmate/tree/master/Themes . `gutterSettings` seems to be more popular than `gutterForeground`. A GitHub search for themes with `gutterSettings` returns [3k hits](https://github.com/search?q=path%3A*.tmTheme%20gutterSettings&type=code) while a search for `gutterForeground` returns [1.5k.](https://github.com/search?q=path%3A*.tmTheme+gutterForeground&type=code) Adding this would probably solve #315

timhillgit:
Many TextMate2 themes use `gutterSettings` to store colors for the gutter area. This is not supported by Sublime Text. This PR enables using `gutterSettings` to set the foreground and background of the gutter but only if they are not already set in `settings`.

Closes #469

### DIFF
--- a/src/highlighting/theme_load.rs
+++ b/src/highlighting/theme_load.rs
@@ -300,7 +300,7 @@ impl ParseSettings for Theme {
             _ => return Err(IncorrectSyntax),
         };
         let mut iter = items.into_iter();
-        let settings = match iter.next() {
+        let mut settings = match iter.next() {
             Some(Settings::Object(mut obj)) => {
                 match obj.remove("settings") {
                     Some(settings) => ThemeSettings::parse_settings(settings)?,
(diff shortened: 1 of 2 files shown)


keith-hall:
Looks good to me. Probably it would also make sense to update the unit tests at https://github.com/trishume/syntect/blob/d3f6d22a709e142ab93875104106998b5b22a524/src/highlighting/theme_set.rs#L87 to prove it parses it as expected.

timhillgit:
Ah, that's a good idea, I was only looking at `theme_load` so I missed it. I'll add that in.

timhillgit:
@keith-hall Are there any more unit tests or anything else you'd like to see before this is merged?

keith-hall (approved):

PR #511, opened 2023-12-28
Eliminate syntect library's dependency on serde's "derive" feature

dtolnay:
Here are charts from `cargo build --timings`.

Notice how **Before**, syn -> serde_derive -> serde -> {bincode,serde_json,plist} -> syntect is a bottleneck in the build, whereas **After**, nothing related to serde is on the critical path. Instead, libc -> cc -> onig_sys -> onig -> syntect is the new bottleneck.

The build is 15% faster end-to-end (7.6 seconds vs 8.9 seconds). More importantly, notice how if whatever downstream crate that depends on syntect has other dependencies that depend on serde, those can begin building after only 2.6 seconds, instead of after 6.3 seconds, which can speed up builds by up to 3.7 seconds, including making IDEs more responsive by that amount.

<table>
<tr><th>Before</th><th>After</th></tr>
<tr><td></td><td></td></tr>
</table>

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -33,6 +33,9 @@ jobs:
     - name: Run cargo check
       run: |
         cargo check --all-features --all-targets
+        # Check that if some other crate in the downstream dependency tree
+        # enables serde's "derive" feature, syntect still builds.
+        cargo check --all-features --features serde/derive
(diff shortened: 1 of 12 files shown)


Enselic:
Nice, thanks!

PR #514, opened 2024-01-07
Automate the release process with a GitHub Actions workflow

Enselic:
To make it easy to make releases, automate the process via GitHub Actions.

Create a workflow that:
1. Runs cargo publish
2. Pushes a git tag, but only if the publish was successful.
3. Creates a GitHub Release with release notes from CHANGELOG.md, but only if creating the tag was successful.

# New release process

After this lands, this will be the new release process:

1. Update Cargo.toml version and CHANGELOG.md on master
2. Run workflow https://github.com/trishume/syntect/actions/workflows/Release.yml on master
3. Done!

# Help needed to setup

Hi @trishume ! Would be great if you could help set up an automated release process. I only need your help with this:

## Create a crates.io token

1. https://crates.io/settings/tokens/new
1. For expiration, I recommend 'No expiration' so we don't have to keep updating it.
1. For 'Scopes', select 'publish-update'. That way, if the token leaks, it can't be used to remove us as owners on crates.io.
1. For 'Crates', limit it to `syntect`.
1. Click 'Generate token'

If you prefer me to use a token belonging to my account, that's perfectly fine with me, just let me know. I'll then share it with you through gpg.

## Create a GitHub Actions secret

1. https://github.com/trishume/syntect/settings/environments/new and call it `crates.io`
1. Add secret and call it `CARGO_REGISTRY_TOKEN` and paste the above token
1. In 'Deployment branches and tags', make it available only to the branch 'master' by selecting "Protected branches only" on the dropbox to the right, to minimize risk of it leaking

Done! I'll take it from there.

### DIFF
--- a/.github/workflows/Release.yml
+++ b/.github/workflows/Release.yml
@@ -1,0 +1,64 @@
+name: Release
+
+# To make a release:
+#
+# 1. Update Cargo.toml version and CHANGELOG.md on master
+# 2. Run workflow https://github.com/trishume/syntect/actions/workflows/Release.yml on master
+# 3. Done!
+
+on:
+  workflow_dispatch: # This workflow can only be triggered manually.
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  # Make sure regular CI passes before we make a release.
+  ci:
+    uses: ./.github/workflows/CI.yml
+
+  # After regular CI passes we publish to crates.io and push a git tag.
+  publish-and-tag:
+    needs: ci
+    runs-on: ubuntu-latest
+    environment:
+      name: crates.io
+      url: https://crates.io/crates/syntect
+    permissions:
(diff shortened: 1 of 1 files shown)


Enselic:
@trishume Friendly one-month ping. Collaborators can't access repo secrets, so you are the only one that can. Doing the steps above shouldn't take many minutes.

If you are uncomfortable with storing the secret on GitHub, I could make it a one-time masked input secret that is inputted each run, if you prefer. That way no secrets are stored anywhere.

But the risk of the secret leaking by GitHub is low (I have never heard of it happening), and if it happens it is not a super big deal, since it can't be used to remove us as owners from crates.io.

Whatever method you prefer, I will make sure to also add a document to the repo to document what steps to take to make a release. (I also wrote those steps in the PR description.)

PR #536, opened 2024-04-29
Implement extends

jalil-salame:
This adds support for the `extends` keyword which lets the tests pass on Build 4075 as requested in #323. The implementation is very hacky right now:

- We now require an optional `base_syntax` when parsing a syntax, if the `base_syntax` is not present and we find an `extends` keyword at the top level, we error out reporting which syntax file we expected.
- `SyntaxSetBuilder` handles this error and resolves the syntaxes in a loop until no more syntaxes can be resolved.
- We don't handle multiple-inheritance (`extends` might be a list).
- We don't re-evaluate the `contexts` as defined in the spec (you can override variables and have that affect the context).
- We only handle `extends` for syntaxes added through `SyntaxSetBuilder::load_from_folder`.

I am opening this PR on top of #535 to get the fixes to the tests.

I don't know if this is a correct implementation of the `extends` keyword, specifically, I don't know when contexts are evaluated; the contexts are properly merged, but the variables can be overwritten and thus the context needs to be re-evaluated, I don't know if I need to do something about it, or id the context evaluation is done after the `SyntaxSet` is built.

I am not happy with the implementation either, but I don't know how to improve it further. I will happily improve it if you have any ideas.

Ideally, `SyntaxSetBuilder::build` would return an error that indicates the syntax was unable to extend another syntax.


keith-hall:
I forgot to say this at the time, but thanks very much for your hard work on this! It is very much appreciated! :rocket: 

> I don't know when contexts are evaluated; the contexts are properly merged, but the variables can be overwritten and thus the context needs to be re-evaluated, I don't know if I need to do something about it, or if the context evaluation is done after the `SyntaxSet` is built.

The match patterns are currently evaluated when the syntax definition is parsed. In `yaml_load`, `parse_match_pattern` calls `parse_regex` which in turn calls `resolve_variables`. 
I raised a PR at https://github.com/jalil-salame/syntect/pull/2 with a WIP idea of how we can achieve updating the variable references. Maybe it will help us push this support for `extends` forward.

keith-hall:
I notice there is an idea to delay expansion of variables - https://github.com/trishume/syntect/issues/541#issue-2274546000 - perhaps we should also consider that as a nicer solution to the problem of allowing extended syntaxes to use the overridden variables?

stefanobaghino:
IIUC this has been superseded by #610 and can now be closed. 🙂

Issue #537, opened 2024-05-01
Replace `yaml-rust`  with `yaml-rust2` in order to not be affected by `RUSTSEC-2024-0320`


d3rpp:
`yaml-rust` appears to be [unmaintained](https://github.com/rustsec/advisory-db/issues/1921), however, the community has published an actively maintained fork under the name of `yaml-rust2`, It is a drop in replacement.

Details on "Vulnerability" - [RUSTSEC-2024-0320](https://rustsec.org/advisories/RUSTSEC-2024-0320.html).

Assuming this isn't a duplicate (I could not find this being brought up before in this issues list), I will submit a PR dropping this in, it's a single line so assuming the developers of `yaml-rust2` didn't change the API, shouldn't be an issue.

d3rpp:
Update: it appears, `yaml-rust` isn't being used in the crate at all?

I'll submit a PR to run the update regardless, though it may be worth looking into removing it all together since the dependency doesn't seem to be referenced in the code at all.

d3rpp:
> > Update: it appears, `yaml-rust` isn't being used in the crate at all?
> 
> 
> 
> It's used throughout this file
> 
> 
> 
> https://github.com/trishume/syntect/blob/de715e5a8194ecaf0d0d78fbdd97add54277ffbd/src/parsing/yaml_load.rs#L8-L9

Ah, my apologies

d3rpp:
PR: #526

PR #538, opened 2024-05-01
chore: replace yaml-rust with actively maintained yaml-rust2

d3rpp:
It would be worth investigating if the dependency is required at all, this simply replaces it with the `crate` key in the `Cargo.toml` with `yaml-rust2`.

TODO: update changelog when everything is sorted out 👍

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ exclude = [
 ]
 
 [dependencies]
-yaml-rust = { version = "0.4.5", optional = true, crate = "yaml-rust2" }
+yaml-rust = { version = "0.4.5", optional = true, package = "yaml-rust2" }
 onig = { version = "6.0", optional = true, default-features = false }
 fancy-regex = { version = "0.11", optional = true }
 walkdir = "2.0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ exclude = [
 ]
 
 [dependencies]
-yaml-rust = { version = "0.4.5", optional = true }
+yaml-rust = { version = "0.4.5", optional = true, crate = "yaml-rust2" }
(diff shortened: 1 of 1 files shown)


d3rpp:
Update: I was wrong it is in fact used 👍

d3rpp:
Closes: #537

Cargo.toml
```diff
@@ -18,7 +18,7 @@ exclude = [
 ]
 
 [dependencies]
-yaml-rust = { version = "0.4.5", optional = true }
+yaml-rust = { version = "0.4.5", optional = true, crate = "yaml-rust2" }
```
Enselic, line 21: Can you point me to documentation on this property please? I can't find it at https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html

Enselic: Seems like it does not work. CI says: warning: unused manifest key: dependencies.yaml-rust.crate

Enselic: Maybe we can use one of the other PRs

d3rpp: https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#renaming-dependencies-in-cargotoml

I have made a mistake in the PR
the key this PR is attempting to use is called `package` not `crate`, my apologies.

Enselic: No worries, I think we should use https://github.com/trishume/syntect/pull/544 though, it looks similar to the first PR to change to yaml2

d3rpp: > No worries, I think we should use #544 though, it looks similar to the first PR to change to yaml2

#544 seems like a cleaner change than this PR, so that would make sense 👍

d3rpp:
Closing as #544 does the fix in a better fashion

PR #540, opened 2024-05-02
refactor: Move `variables` into compressed contexts

### LINKED ISSUE #482: Consider compressing `SyntaxReference`'s `variables`
From my understanding the separation for what gets compressed and what doesn't for `SyntaxReference`s are whether the values are expected to be used when trying to match on the syntax vs actually highlighting using the syntax. By that logic I believe that `variables` should also be compressed until used (lingering question: should this be included within `LazyContexts` or kept separate?)

The main motivation being that some syntax definitions have very large variables that wind up bloating the serialized size of `SyntaxSet`s. For instance `C.sublime-syntax` has a `window_types` variable that's [over a KiB](https://github.com/sublimehq/Packages/blob/4d121a215099087e2c592aee0b3fa107bf491d7d/C%2B%2B/C.sublime-syntax#L44) (and `bat`'s Julia sublime syntax has a variable with a value that's over 8 KiBs by itself)

CosmicHorrorDev:
Hacky implementation that resolves #482

This moves `variables` into `LazyContexts` that get compressed for a `SyntaxReference`

This PR is mostly just a proof-of-concept, so I'd be more than happy to make changes


CosmicHorrorDev:
Worth mentioning that this will be a breaking change of course

CosmicHorrorDev:
Reported the CI issue upstream https://github.com/serde-rs/serde/issues/2734

PR #542, opened 2024-05-02
chore: Bump nightly version to fix the test suite

CosmicHorrorDev:
`serde` made a release adding support for saturating types that fails to compile on the specific nightly that's currently set as `public_api::MINIMUM_NIGHTLY_RUST_VERSION` and it can't be fixed at this point without breaking backwards compat. This PR bumps the nightly version used for the public API check to the first one that starts compiling successfully again


Enselic (approved):
Looks great, thanks! Bonus points for the sanity check.

PR #548, opened 2024-08-02
fix(sublime-syntax): catch backslash pattern

### LINKED ISSUE #547: Panics when a regex (from a `.sublime-syntax` file) includes unescaped backslash
Hello, I discovered a panic in Typst: https://github.com/typst/typst/issues/4421. I was able to recreate the panic and the use case of the library in Typst:

```rs
use syntect::{
    easy::HighlightLines,
    highlighting::ThemeSet,
    parsing::{SyntaxDefinition, SyntaxSetBuilder},
};

/// Content of `a.sublime-syntax` file.
static A_SUBLIME_SYNTAX: &str = r#"
%YAML 1.2
---
name: lang
file_extensions:
  - a
scope: source
contexts:
  main:
    - match: '\'
"#;

fn main() {
    let text = "text";
    let file_extension = "a";
    let theme = &ThemeSet::load_defaults().themes["base16-ocean.dark"];
    let syntax_set = {
        let mut out = SyntaxSetBuilder::new();
        out.add(SyntaxDefinition::load_from_str(A_SUBLIME_SYNTAX, false, None).unwrap());
        out.build()
    };
    let syntax = syntax_set.find_syntax_by_token(file_extension).unwrap();
    let _ = HighlightLines::new(syntax, theme).highlight_line(text, &syntax_set);
}
```

```toml
[dependencies]
syntect = "5.2.0"

[patch.crates-io]
syntect = { git = "https://github.com/trishume/syntect" }
```

```
thread 'main' panicked at /home/user/.local/share/cargo/git/checkouts/syntect-e33b39f181e4f0f4/d023aaa/src/parsing/regex.rs:70:53:
regex string should be pre-tested: Error(OnigError(-104), end pattern at escape)
```

https://github.com/trishume/syntect/blob/d023aaa509d9e5058d55f9aa787c88f9a74bb180/src/parsing/regex.rs#L70

There should be some way to get `Result::Err` from some function call to be able to show in stderr what exactly went wrong. This error is generally "invalid regex" or more specifically "invalid regex: backslash doesn't escape anything".

I was able to remove panic by validating the regex before searching:

https://github.com/trishume/syntect/blob/d023aaa509d9e5058d55f9aa787c88f9a74bb180/src/parsing/regex.rs#L64-L65

```rs
    pub fn search(
        &self,
        text: &str,
        begin: usize,
        end: usize,
        region: Option<&mut Region>,
    ) -> bool {
        if Regex::try_compile(self.regex_str()).is_none() {
            return false;
        }
        self.regex()
            .search(text, begin, end, region.map(|r| &mut r.region))
    }
```

But this wouldn't give away any `Error`. I also don't see any way to use `Regex::try_compile()` directly in Typst/MRA code, therefore some other way of validating all regexes in a `.sublime-syntax` file should be added.

Andrew15-5:
Fixes #547.

### DIFF
--- a/src/parsing/syntax_definition.rs
+++ b/src/parsing/syntax_definition.rs
@@ -241,6 +241,9 @@ where
 
         last_was_escape = c == '\\' && !last_was_escape;
     }
+    if last_was_escape {
+        reg_str.push('\\');
+    }
     reg_str
 }
 
--- a/src/parsing/yaml_load.rs
+++ b/src/parsing/yaml_load.rs
@@ -1319,4 +1319,25 @@ mod tests {
         println!("{:?}", valid_indexes);
         assert_eq!(valid_indexes, [0, 1, 5, 6]);
     }
+
+    #[test]
+    fn error_loading_syntax_with_unescaped_backslash() {
+        let load_err = SyntaxDefinition::load_from_str(
+            r#"
+            name: Unescaped Backslash
+            scope: source.c
+            file_extensions: [test]
+            contexts:
+              main:
+                - match: '\'
+            "#,
+            false,
+            None,
+        )
+        .unwrap_err();
+        match load_err {
+            ParseSyntaxError::RegexCompileError(bad_regex, _) => assert_eq!(bad_regex, r"\"),
+            _ => panic!("Unexpected error: {load_err}"),
+        }
+    }
 }


CosmicHorrorDev:
I've added a regression test (https://github.com/trishume/syntect/commit/71d3802aba711169664f5351981bbfce9c24e919) for this which I've confirmed fails on current `master`, but passes with this PR. Feel free to cherry-pick it if ya want!

CosmicHorrorDev:
And I switched it to https://github.com/trishume/syntect/commit/1d9e35cd6745b40a61efbf4d40ba690227aed23d which is slightly shorter, and also in the right place now :sweat_smile:

Andrew15-5:
Wait, huh? Cherry-picking tests from master? Never heard of this. If I do, will it be removed? I'm guessing yes, if force-pushed commits are apparently not an issue on master.

Regardless, if it's necessary, you can just include that commit in this branch, no? You should have write access after all.

CosmicHorrorDev:
The commit is building on your branch, it's just pushed to my fork, so you would have to add my fork as a remote > `git fetch` > then you'd be able to `git cherry-pick` it

> Regardless, if it's necessary, you can just include that commit in this branch, no? You should have write access after all.

That'd be nice, but I'm just a contributor, so I don't have the ability to do that. I could open a new PR that builds off of this one though

Enselic:
If this PR gets a regression test that fails without the fix I'm happy to merge!

Andrew15-5:
> it's just pushed to my fork

Oh, so it's not this master. Now it all make sense.

Enselic:
@CosmicHorrorDev does this PR look good to you now?

CosmicHorrorDev:
LGTM! I don't care about the commit authorship as long as the test is there ❤️

Enselic (approved):
Thanks for checking, let's merge then!

PR #554, opened 2024-10-20
dumps/serialize_to_writer: insert BufWriter to speed up serialization

lschuermann:
By buffering bincode's output prior to compression, we can achieve a ~30x speedup of syntect in debug builds (such as used in `cargo test`s by default) on a single `SyntaxSetBuilder::build()` call. For Tock's `license-checker` [1] crate using `syntect`, this brings overall test times down from ~75s to 4s on an i7-1360P. This change does not affect release builds much, bringing an individual test down from 0.17 to 0.13 seconds.

[1]: https://github.com/tock/tock/tree/a109dd65fb269479ca4fa020af0a54da740eca28/tools/license-checker

### DIFF
--- a/src/dumps.rs
+++ b/src/dumps.rs
@@ -123,7 +123,7 @@ fn serialize_to_writer_impl<T: Serialize, W: Write>(
     use_compression: bool,
 ) -> Result<()> {
     if use_compression {
-        let mut encoder = ZlibEncoder::new(output, Compression::best());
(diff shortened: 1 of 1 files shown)


keith-hall (approved):

Enselic (approved):
Nice

lschuermann:
@keith-hall @Enselic Thanks for the swift reviews! Funnily enough, it seems like this change has sped up syntect's own CI pipeline by 4-5x, hah!

PR #544, opened 2024-05-30
Fix unmaintained audit warning for yaml-rust

SleeplessOne1917:
When running [`cargo audit`](https://crates.io/crates/cargo-audit) on this repo, I get a warning that `yaml-rust` is no longer maintained:

```
Crate:     yaml-rust
Version:   0.4.5
Warning:   unmaintained
Title:     yaml-rust is unmaintained.
Date:      2024-03-20
ID:        RUSTSEC-2024-0320
URL:       https://rustsec.org/advisories/RUSTSEC-2024-0320
Dependency tree:
yaml-rust 0.4.5
└── syntect 5.2.0
```

This PR replaces that dependency with [`yaml-rust2`](https://crates.io/crates/yaml-rust2), which does the same thing and has the same API, but is also actively maintained.

Some other warnings popped up as well, but they were related to dev dependencies, so likely not a big deal.

PR #555, opened 2024-11-04
chore: use yaml-rust2 instead of unmaintained yaml-rust

kayagokalp:
Supersedes #544.

Uses `yaml-rust2` instead of `yaml-rust` which is unmaintained.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ exclude = [
 ]
 
 [dependencies]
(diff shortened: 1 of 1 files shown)


kayagokalp:
If getting #544 to the finish line is preferable I can also take over that as well. We have a dependency to this crate and want to remove the audit warning without having to patch on our end. Let me know how can I help

PR #565, opened 2025-01-10
fix clippy 1.84 warnings

extrawurst:

PR #568, opened 2025-01-12
tests: Use `insta` instead of `expect-test` for public API test

Enselic:
The failure in https://github.com/trishume/syntect/actions/runs/12734263547/job/35491626740?pr=565 (for https://github.com/trishume/syntect/pull/565) was too verbose and not very friendly.

Let's switch to `insta` which has a better way to deal with snapshots.

Marking as Draft until I have had time to do some more experiments. And to give others a chance to give feedback on the change as such.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -161,6 +161,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "console"
+version = "0.15.10"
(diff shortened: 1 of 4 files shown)


CosmicHorrorDev:
Just went through using this change in #576 and it was a much friendlier experience than I remember (although I'm already plenty familiar with `insta` so there is bias)

PR #569, opened 2025-01-18
refactor: add convenience `HighlightLines::from_state` constructor

alexpasmantier:
This PR adds a convenience method to easily be able to reinstantiate a `HighlightLines` object from a previous state the user might have cached.

### DIFF
--- a/src/easy.rs
+++ b/src/easy.rs
@@ -84,6 +84,20 @@ impl<'a> HighlightLines<'a> {
             HighlightIterator::new(&mut self.highlight_state, &ops[..], line, &self.highlighter);
         Ok(iter.collect())
     }
+
+    /// This starts again from a previous state, useful for highlighting a file incrementally for
+    /// which you've cached the highlight and parse state.
+    pub fn from_state(
+        theme: &'a Theme,
+        highlight_state: HighlightState,
+        parse_state: ParseState,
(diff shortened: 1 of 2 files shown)


keith-hall (approved):

keith-hall (approved):

Enselic (commented):
Personally I would prefer a test that demonstrates how to use this new function, that also serves as a regression test.

But I'm fine you merging this as-is Keith if you are confident this is needed and the right approach.

keith-hall:
I'm confident that this is needed and the right approach. I also agree that a test would be nice :slightly_smiling_face:

alexpasmantier:
Thank you both for your input 🙏🏻 

I added a `syntect::easy::HighlightLines<'a>::state` method, without which the first one wouldn't be very useful and an associated test.

Didn't add the test as an example to the method's documentation but can do so if you think it's a good idea.

Cheers

keith-hall (approved):
Nice, thanks for the test. Looks good to me

Enselic (approved):
Looks good to me too, thanks.

I'd like to ask trishume to allow CI to run without manual action, but that's a later thing.

alexpasmantier:
Thanks for the quick feedback 🙏
So should we merge this?

PR #576, opened 2025-03-15
refactor: Drop once_cell in favor of OnceLock/LazyLock

### LINKED ISSUE #575: Public dependency on `once_cell`
Now that there are things like [`OnceLock` (since 1.70.0)](https://doc.rust-lang.org/std/cell/struct.OnceCell.html) and [`LazyLock`(since 1.80.0)](https://doc.rust-lang.org/std/sync/struct.LazyLock.html) in the standard library it'd be nice to migrate off of `once_cell`. The one current issue with this is that it's a breaking change for this library due to [`syntect::parsing::SCOPE_REPO`](https://docs.rs/syntect/latest/syntect/parsing/static.SCOPE_REPO.html) being a `once_cell::sync::Lazy`. It'd be nice to change this in the next breaking release if possible (and I'd be willing to open a PR to handle it)

CosmicHorrorDev:
Resolves #575

I stuck with `OnceLock` for the library itself to keep the MSRV bump to just 1.70 instead of 1.80. This did involve refactoring `SCOPE_REPO` to a function that internally handles initialization to approximate what `Lazy` does


src/parsing/scope.rs
```diff
@@ -73,12 +74,11 @@ pub enum ParseScopeError {
 /// The structure used to keep track of the mapping between scope atom numbers and their string
 /// names
 ///
-/// It is only exposed in case you want to lock [`SCOPE_REPO`] and then allocate a bunch of scopes
+/// It is only exposed in case you want to lock [`scope_repo()`] and then allocate a bunch of scopes
 /// at once without thrashing the lock. In general, you should just use [`Scope::new()`].
 ///
 /// Only [`Scope`]s created by the same repository have valid comparison results.
 ///
-/// [`SCOPE_REPO`]: struct.SCOPE_REPO.html
```
CosmicHorrorDev, line 81: The function should get auto-linked, so I dropped this explicit link, but lmk if that's wrong for some reason

tests/snapshots/public_api__public_api.snap
```diff
@@ -1219,7 +1219,7 @@ impl core::marker::Unpin for syntect::parsing::SyntaxSetBuilder
 impl core::panic::unwind_safe::RefUnwindSafe for syntect::parsing::SyntaxSetBuilder
 impl core::panic::unwind_safe::UnwindSafe for syntect::parsing::SyntaxSetBuilder
 pub const syntect::parsing::ATOM_LEN_BITS: u16 = 3u16
-pub static syntect::parsing::SCOPE_REPO: once_cell::sync::Lazy<std::sync::mutex::Mutex<syntect::parsing::ScopeRepository>>
+pub fn syntect::parsing::scope_repo() -> &'static std::sync::mutex::Mutex<syntect::parsing::ScopeRepository>
```
CosmicHorrorDev, line 1222: Already mentioned in #575, but this is a breaking change :boom:

keith-hall (approved):

tests/snapshots/public_api__public_api.snap
Enselic: If we need to break semver anyway, it is probably better to completely remove it from the public API? We can always add it back in a minor release if someone depends on it.

stefanobaghino:
If merged, this would be superseded by #625.

PR #577, opened 2025-03-16
Speed up unittests runtime

CosmicHorrorDev:
Originally motivated by this comment:

https://github.com/trishume/syntect/blob/7fe13c0fd53cdfa0f9fea1aa14c5ba37f81d8b71/.github/workflows/CI.yml#L90-L91

Two changes

## d0c2431a159a86d4ae1001083b7d4cd7fa1c0c9b | Bump some deps dev `opt-level`

bumps up the `dev` profile opt-level for a few deps that were picked up when looking at a profile of some of the slower tests

## 96b0d19cf331d539434cfb71151f37610c49abb2 | Share testdata/Packages syntax set loading

shares the initialization of tests' various `"testdata/Packages"` syntax sets. This

### a) avoids doing common intensive work

but more importantly

### b) avoids some `SyntaxSet` loading pathological behavior

where loading a syntax set from a folder across N threads locks in way that's akin to sequentially loading the syntax sets from a single thread. Running the tests before this PR generally give that impression since they peg a bit above 1 CPU core for most of the duration even though each unittest runs in a separate thread. I'm assuming this is from different stages of loading syntax sets grabbing a process-wide lock

Here's a profile of said behavior demostrated by plopping the following code in `lib.rs` and running `cargo test --no-default-features --features=default-fancy slow`

https://share.firefox.dev/43NTvYw

```rust
#[cfg(test)]
mod slow {
    use crate::parsing::SyntaxSet;

    #[test]
    fn slow1() {
        SyntaxSet::load_from_folder("testdata/Packages").unwrap();
    }

    #[test]
    fn slow2() {
        SyntaxSet::load_from_folder("testdata/Packages").unwrap();
    }

    #[test]
    fn slow3() {
        SyntaxSet::load_from_folder("testdata/Packages").unwrap();
    }

    #[test]
    fn slow4() {
        SyntaxSet::load_from_folder("testdata/Packages").unwrap();
    }
}
```

You can see the staggered chunks of execution time. It's usually just compiling regexes which is no surprise

image

---

Anyhoo, here is the improvement when running unittests with the `default-fancy` feature on my laptop

| what | unittest run-time | % change from `master` |
:--: | --: | --: |
| `master` | 258.53s | -0.0% |
| just d0c2431a159a86d4ae1001083b7d4cd7fa1c0c9b | 57.91s | -77.6% |
| both commits | 22.69s | -91.2% |

`release` unittest runtimes drop from 35s to 12s for me as well :rocket:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -79,6 +79,12 @@ default = ["default-onig"]
 # [profile.release]
 # debug = true
 
+[profile.dev.package]
+aho-corasick.opt-level = 2
+fancy-regex.opt-level = 2
+regex-automata.opt-level = 2
+regex-syntax.opt-level = 2
+
 [lib]
 bench = false
 
--- a/src/highlighting/highlighter.rs
+++ b/src/highlighting/highlighter.rs
@@ -390,11 +390,12 @@ impl<'a> Highlighter<'a> {
 mod tests {
     use super::*;
     use crate::highlighting::{Color, FontStyle, Style, ThemeSet};
-    use crate::parsing::{ParseState, ScopeStack, SyntaxSet};
+    use crate::parsing::{ParseState, ScopeStack};
+    use crate::utils::testdata;
 
     #[test]
     fn can_parse() {
-        let ps = SyntaxSet::load_from_folder("testdata/Packages").unwrap();
+        let ps = &*testdata::PACKAGES_SYN_SET;
         let mut state = {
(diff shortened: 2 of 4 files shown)


keith-hall (approved):
Wow, nice, thanks! 🚀

Enselic (approved):
I haven't verified this myself but I see no reason to doubt that it will work, thanks!

PR #579, opened 2025-03-28
Add `code-to-pdf`

Tommypop2:
A small CLI that allows for generating a syntax-highlighted PDF of code (or any directory with text files).
It also works with images (embeds them into the document on their own page)

Repo is https://github.com/Tommypop2/code-to-pdf


keith-hall (approved):
Thanks!

PR #580, opened 2025-03-29
docs: Deprecate `SCOPE_REPO` for removal

CosmicHorrorDev:
Related https://github.com/trishume/syntect/issues/575#issuecomment-2731203196

Marks `syntect::parsing::SCOPE_REPO` as deprecated in anticipation of removal in the next semver-breaking release. The deprecation renders as follows

```text
warning: use of deprecated static `parsing::scope::SCOPE_REPO`: Deprecated in anticipation of removal in the next semver-breaking release under the justification that it's incredibly niche functionality to expose. If you rely on this functionality then please express your particular use-case in the github issue: https://github.com/trishume/syntect/issues/575
   --> src/parsing/scope.rs:237:24
    |
237 |         let mut repo = SCOPE_REPO.lock().unwrap();
    |                        ^^^^^^^^^^
```

### DIFF
--- a/src/html.rs
+++ b/src/html.rs
@@ -3,8 +3,8 @@ use crate::easy::{HighlightFile, HighlightLines};
 use crate::escape::Escape;
 use crate::highlighting::{Color, FontStyle, Style, Theme};
 use crate::parsing::{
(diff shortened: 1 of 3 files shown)


CosmicHorrorDev:
Rebased on top of `master` and added `since = "5.3.0"` to the deprecation under the assumption that this might make it out in the next 5.x release

PR #581, opened 2025-04-05
Readme: list Comrak.

kivikakk:
Hi! Thought I'd add [Comrak](https://github.com/kivikakk/comrak) to the list. :)

Not knowing how long other projects have been using syntect, I don't really know where to put it in the list (Comrak added syntect integration in 2021), so I'm adding it at the bottom.


keith-hall (approved):
Thanks! 🙂

PR #587, opened 2025-06-25
deps: switch `serde_json` to an optional dep

CosmicHorrorDev:
Moves the `serde_json` dep behind the features where it gets used

```bash
$ rg serde_json::
# behind the metadata feature
src/lib.rs
85:    ParseMetadata(#[from] serde_json::Error),

# behind the plist-load feature
src/highlighting/settings.rs
6:pub use serde_json::Value as Settings;

# behind the metadata feature
src/parsing/metadata.rs
20:type Dict = serde_json::Map<String, Settings>;
180:    match serde_json::from_value::<ShellVars>(vars) {
261:            serde_json::from_value(settings.into()).map_err(|e| format!("{}: {:?}", path, e))?;
438:        Ok(serde_json::from_value(contents)?)

# only used in tests
src/parsing/regex.rs
281:        let pattern: Regex = serde_json::from_str("\"just a string\"").unwrap();
283:        let back_to_str = serde_json::to_string(&pattern).unwrap();
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -30,7 +30,7 @@ flate2 = { version = "1.0", optional = true }
 fnv = { version = "1.0", optional = true }
 serde = "1.0"
 serde_derive = "1.0"
-serde_json = "1.0"
+serde_json = { version = "1.0", optional = true }
 once_cell = "1.8"
 thiserror = "1.0"
 
@@ -44,6 +44,7 @@ rustup-toolchain = "0.1.5"
 rustdoc-json = "0.8.8"
 public-api = "0.33.1"
 insta = "1.42.0"
+serde_json = "1.0"
 
 [features]
 
@@ -58,7 +59,7 @@ regex-onig = ["onig"]
 parsing = ["regex-syntax", "fnv", "dump-create", "dump-load"]
 
 # Support for .tmPreferenes metadata files (indentation, comment syntax, etc)
-metadata = ["parsing", "plist-load"]
+metadata = ["parsing", "plist-load", "dep:serde_json"]
 
 # Enables inclusion of the default syntax packages.
 default-syntaxes = ["parsing", "dump-load"]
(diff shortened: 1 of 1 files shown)


Enselic:
do we test with and without this feature in CI? if yes, where?

CosmicHorrorDev:
Not as far as I can see. I'd be happy to add more feature sets to CI if desired

Enselic:
that'd be great, thanks

CosmicHorrorDev:
Alright, so CI-wise the main bulk of different feature sets that are checked looks to be done here

https://github.com/trishume/syntect/blob/64644ffe064457265cbcee12a0c1baf9485ba6ee/.github/workflows/CI.yml#L82-L100

From all of those combinations we have the following relevant to `metadata` and `plist-load`

### Steps that enable neither 

- `cargo run --example synhtml --no-default-features --features html,default-syntaxes,default-themes,regex-onig -- examples/synhtml.rs`
- `cargo test --lib --example synstats --no-default-features --features default-syntaxes,yaml-load,regex-onig`
- `cargo test --lib --no-default-features`

### Steps that enable just `plist-load` (default)

- `cargo build`
- `cargo test --features default-fancy --no-default-features --release`

### Steps that enable both

- `cargo test --features metadata`

---

Overall the only category missing is a step that enables `metadata` without `plist-load`, but that doesn't really seem useful since `metadata` already enables `plist-load`

A more comprehensive option would be to additionally check feature-sets with something like [`cargo-hack`](https://crates.io/crates/cargo-hack). The only clunky part is that the `parsing` feature implicitly requires some regex backend to be enabled. The API difference between the two backends already seems covered decently well from the feature sets above, so forcing one regex backend helps avoid too much combinatorial explosion. Something like

```bash
cargo hack \
    --feature-powerset \
    --depth=2 \
    --features=regex-fancy \
    --exclude-features=regex-onig \
    check
```

covers all combinations of up-to two different features being enabled along with `regex-fancy` always being set. It checks 38 different feature sets and takes 49 seconds to finish from a clean build. There are a lot of other options as well like using `--at-least-one-of=regex-fancy,regex-onig` instead of `--features=..` and `--exlude-features=...`, tuning `--depth`, etc.

PR #589, opened 2025-07-24
docs: Enable `metadata` feature on docs.rs builds

CosmicHorrorDev:
It looks like `metadata` is the only feature that isn't on by default aside from features used to toggle the regex backend. This ensures that metadata API bits actually have documentation generated

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,10 @@ exclude = [
     "/codecov.yml"
 ]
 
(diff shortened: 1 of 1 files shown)


CosmicHorrorDev:
Alternatively `all-features = true` could be used although that would build `fancy-regex` too

keith-hall (approved):

Enselic (approved):

Issue #512, opened 2023-12-29
Remove dependencies from public API


Enselic:
The public API of syntect includes types defined in other crates. This makes it tricky to bump dependencies within a minor syntect release since bumping a dependency can change the public API in incompatible ways.

It would be nice if we for syntect 6.0.0 could go over the public API and remove external types from the public API.

# How to find inappropriate types in the current public API

Adding

    cargo-features = ["public-dependency"]

to Cargo.toml and building with Rust nightly currently gives these warnings:

<details>

<summary>output</summary>

```
warning: type `bincode::ErrorKind` from private dependency 'bincode' in public interface
  --> src/dumps.rs:45:1
   |
45 | pub fn dump_to_writer<T: Serialize, W: Write>(to_dump: &T, output: W) -> Result<()> {
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `#[warn(exported_private_dependencies)]` on by default

warning: type `bincode::ErrorKind` from private dependency 'bincode' in public interface
  --> src/dumps.rs:66:1
   |
66 | pub fn dump_to_file<T: Serialize, P: AsRef<Path>>(o: &T, path: P) -> Result<()> {
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

warning: type `bincode::ErrorKind` from private dependency 'bincode' in public interface
  --> src/dumps.rs:73:1
   |
73 | pub fn from_reader<T: DeserializeOwned, R: BufRead>(input: R) -> Result<T> {
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

warning: type `bincode::ErrorKind` from private dependency 'bincode' in public interface
  --> src/dumps.rs:87:1
   |
87 | pub fn from_dump_file<T: DeserializeOwned, P: AsRef<Path>>(path: P) -> Result<T> {
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

[...]

PR #590, opened 2025-07-26
refactor: replace `bitflags` with the expanded implementation

CosmicHorrorDev:
Ref #512

`bitflags` was only being used for `FontStyle`, so it seems reasonable to me to just replace it with its expanded code (courtesy of `cargo-expand` plus lots of manual cleanup). No more outdated dependency for anyone to complain about now :laughing::speak_no_evil:. This also gives us the liberty of customizing the docs to our specific use along with being able to freely `#[deprecate]` anything that we want to drop with the next major release (I'm looking at you `unsafe fn from_bits_unchecked(bits: u8) -> Self` :eyes:)

That being said the generated code was _very_ gross because it handled a bunch of edge-cases that we can ignore, so I also took a few passes at cleaning things up with -- what _should_ be -- equivalent implementations. Take the original expanded debug impl for instance (yes it was the worst of the bunch)

```rust
impl core::fmt::Debug for FontStyle {
    fn fmt(
        &self,
        f: &mut core::fmt::Formatter,
    ) -> core::fmt::Result {
        #[allow(non_snake_case)]
        trait __BitFlags {
            #[inline]
            fn BOLD(&self) -> bool {
                false
            }
            #[inline]
            fn UNDERLINE(&self) -> bool {
                false
            }
            #[inline]
            fn ITALIC(&self) -> bool {
                false
            }
        }
        #[allow(non_snake_case)]
        impl __BitFlags for FontStyle {
            #[allow(deprecated)]
            #[inline]
            fn BOLD(&self) -> bool {
                if Self::BOLD.bits == 0 && self.bits != 0 {
                    false
                } else {
                    self.bits & Self::BOLD.bits == Self::BOLD.bits
                }
            }
            #[allow(deprecated)]
            #[inline]
            fn UNDERLINE(&self) -> bool {
                if Self::UNDERLINE.bits == 0 && self.bits != 0 {
                    false
                } else {
                    self.bits & Self::UNDERLINE.bits == Self::UNDERLINE.bits
                }
            }
            #[allow(deprecated)]
            #[inline]
            fn ITALIC(&self) -> bool {
                if Self::ITALIC.bits == 0 && self.bits != 0 {
                    false
                } else {
                    self.bits & Self::ITALIC.bits == Self::ITALIC.bits
                }
            }
        }
        let mut first = true;
        if <Self as __BitFlags>::BOLD(self) {
            if !first {
                f.write_str(" | ")?;
            }
            first = false;
            f.write_str("BOLD")?;
        }
        if <Self as __BitFlags>::UNDERLINE(self) {
            if !first {
                f.write_str(" | ")?;
            }
            first = false;
            f.write_str("UNDERLINE")?;
        }
        if <Self as __BitFlags>::ITALIC(self) {
            if !first {
                f.write_str(" | ")?;
            }
            first = false;
            f.write_str("ITALIC")?;
        }
        let extra_bits = self.bits & !Self::all().bits();
        if extra_bits != 0 {
            if !first {
                f.write_str(" | ")?;
            }
            first = false;
            f.write_str("0x")?;
            core::fmt::LowerHex::fmt(&extra_bits, f)?;
        }
        if first {
            f.write_str("(empty)")?;
        }
        Ok(())
    }
}
```

Cursed looking? _Absolutely_

Here's the cleaned up version

```rust
impl fmt::Debug for FontStyle {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        let mut empty = true;

        let pairs = [
            (Self::BOLD, "BOLD"),
            (Self::UNDERLINE, "UNDERLINE"),
            (Self::ITALIC, "ITALIC"),
        ];
        for (flag, flag_str) in pairs {
            if self.contains(flag) {
                if !std::mem::take(&mut empty) {
                    f.write_str(" | ")?;
                }
                f.write_str(flag_str)?;
            }
        }

        let extra_bits = self.bits & !Self::all().bits();
        if extra_bits != 0 {
            if !std::mem::take(&mut empty) {
                f.write_str(" | ")?;
            }
            f.write_str("0x")?;
            fmt::LowerHex::fmt(&extra_bits, f)?;
        }

        if empty {
            f.write_str("(empty)")?;
        }

        Ok(())
    }
}
```

Unreachable branches like `if Self::BOLD.bits == 0 && self.bits != 0 {` were pruned (none of our flags `== 0`). The internal `__BitFlags` trait was removed entirely (including the default methods that were all defined). Etc. etc.

Note: I took the liberey of removing all of the `#[inline]`s because it's generally an optimization code smell to have on everything (and it was on virtually everything)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -817,7 +817,6 @@ name = "syntect"
 version = "5.2.0"
 dependencies = [
  "bincode",
- "bitflags 1.3.2",
  "criterion",
  "fancy-regex",
  "flate2",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,6 @@ onig = { version = "6.5.1", optional = true, default-features = false }
 fancy-regex = { version = "0.11", optional = true }
 walkdir = "2.0"
 regex-syntax = { version = "0.8", optional = true }
-bitflags = "1.0.4"
 plist = { version = "1.3", optional = true }
 bincode = { version = "1.0", optional = true }
 flate2 = { version = "1.0", optional = true }
--- a/src/highlighting/style.rs
+++ b/src/highlighting/style.rs
@@ -1,6 +1,5 @@
 // Code based on [https://github.com/defuz/sublimate/blob/master/src/core/syntax/scope.rs](https://github.com/defuz/sublimate/blob/master/src/core/syntax/scope.rs)
 // released under the MIT license by @defuz
-use bitflags::bitflags;
 use serde_derive::{Deserialize, Serialize};
 
 /// Foreground and background colors, with font style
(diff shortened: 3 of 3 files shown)


Enselic (approved):
Definitely a step in the right direction, thanks!

For the next breaking release we might want look over what the API should look like long term, but since the API remains unchanged to consumers I'd say let's merge this.

I'll merge it latest Monday, I want to give other maintainers a reasonable chance of giving input first though.

CosmicHorrorDev:
> For the next breaking release we might want look over what the API should look like long term, but since the API remains unchanged to consumers I'd say let's merge this.

Totally agree!

keith-hall (approved):
Nice, I appreciate the detailed description/ explanation as well btw, thank you! 🚀

Enselic:
If Keith is also on board let's merge :)

PR #591, opened 2025-08-04
deps: Update public-api test deps

CosmicHorrorDev:
This PR updates the dependencies used for the public API test, so that they can continue working on lockfiles that are `version = 4` (fixing the cause of the CI failure that a lot of recent PRs have been running into)

Nothing in the public API changed, so all of the snapshot changes are from the new nightly version:

- Having a new `Freeze` auto-trait, so now there's `Freeze` or `!Freeze` on everything
- Omitting a default generic when used e.g. omitting the `<Rhs = Self>` in `PartialEq<Rhs = Self> for Foo`
- `StructuralEq` no longer being a thing apparently
- `Mutex` is defined in the new `std::sync::poison` module
- `pub const syntect::parsing::ATOM_LEN_BITS: u16 = 3u16` being changed to show `pub const syntect::parsing::ATOM_LEN_BITS: u16`


CosmicHorrorDev:
CI failure already being tracked downstream:

- sharkdp/bat/issues/3363

Enselic:
thanks! code changes look good but can you rebase on master please? it's for https://github.com/trishume/syntect/pull/586 which will make the *.snap diff much easier to read

CosmicHorrorDev:
Hiya! I think this PR is already on top of the current `master` (#586 shows up in the [commits for the branch](https://github.com/CosmicHorrorDev/syntect/commits/update-public-api-test/)]

I don't see any syntax highlighting for it in the browser, but I do see invalid highlighting in the mobile app, so I guess they resolve when to do highlighting differently :shrug:

Enselic:
My bad, you're right. GitHub is buggy. Eventually I intend to replace insta with something that does not have broken diffs.

Enselic (approved):

PR #592, opened 2025-08-04
.gitattributes: Use `linguist-language=Text` not `...=txt`

Enselic:
`Text` is the [canonical name](https://github.com/github-linguist/linguist/blob/1b43482ac6ab97394a70946cb041f3fa5f3284b5/lib/linguist/languages.yml#L7624) and `txt` currently renders like non-text.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -30,4 +30,4 @@ testdata/*                              linguist-vendored
 *.png                         binary
 *.tex                                   diff=tex
 *.pdf                         binary    diff=astextplain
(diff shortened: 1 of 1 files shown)


Enselic:
aha I think the problem is that https://github.com/trishume/syntect/pull/591 is not rebased on https://github.com/trishume/syntect/pull/586

Enselic:
GitHub is buggy. It renders as Text but then when you expand the diff it renders as Jest Snapshot : (

PR #593, opened 2025-08-07
ci: check with many different feature sets

CosmicHorrorDev:
Uses the `cargo hack` command mentioned originally in https://github.com/trishume/syntect/pull/587#issuecomment-3034616172 to automatically ensure that `cargo check` works across many of the different feature sets possible with this crate

I opted to add it as a separate job to try and keep things relatively fast

Also worth noting that I used the `taiki-e/install-action` to install `cargo-hack`. taiki-e is the author of `cargo-hack` so trust for one means a good deal of trust in the other I suppose

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -106,3 +106,17 @@ jobs:
     - name: Docs
       run: |
         cargo doc
+
+  check-feature-powerset:
+    name: Check feature powerset
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


Enselic:
can we fix bat ci somehow? I'd like to know that we don't break bat, but with ci broken I can't know that

I would be happy to approve a bat PR that fixes CI.

keith-hall:
I fixed `bat`'s CI, and reran the regression test on this PR successfully

PR #595, opened 2025-08-11
Use `.txt` file for public API snapshot instead of `.snap` (from `insta`)

Enselic:
So that showing the diff in GitHub does not become [more or less unreadable](https://github.com/trishume/syntect/pull/591/files#diff-f3d28a949326c02c2b2c85c024a030561f3b2521457621bb2550792460c3a2ec) because of faulty syntax highlighting.

The `public-api` crate now has (optional) support for built-in snapshot testing that we use instead of `insta`. But note that the excellent diffing engine of `insta` (from the crate `similar`) is still used.

We get two new indirect `dev-dependencies`:
* `similar-asserts` - from the same author as `similar` and `insta`, so we can trust it.
* `snapshot-testing` - new but tiny and [easily auditable](https://github.com/Enselic/snapshot-testing?tab=readme-ov-file#audit-the-code). (And the author happens to be me.)

If the `public_api` test fails, instructions on how to update are printed:

    thread 'public_api' panicked at tests/public_api.rs:20:16:
    assertion failed: `(left == right)`:

    help: run with env var `UPDATE_SNAPSHOTS=yes` to update snapshots

    ...


Enselic:
I'll look into avoiding MSRV bump later when I get time, maybe today.

CosmicHorrorDev (approved):

CosmicHorrorDev:
Looks good from my pov. I like that it's still a snapshot test 🎉

Enselic:
Thanks for review! I'll merge after updating `snapshot-testing` to the latest version in Cargo.lock.

PR #596, opened 2025-08-12
deps: Update fancy-regex to v0.16.2

CosmicHorrorDev:
Updates `fancy-regex` to the latest version. It's not part of the public API, so no breaking change :partying_face:

This change looks like it makes the following `bat` syntaxes compile successfully with `fancy-regex` as the backend (verified by generating `two-face`'s dumps):

- `syntaxes/02_Extra/LiveScript.sublime-syntax`
- `syntaxes/02_Extra/SCSS_Sass/Syntaxes/Sass.sublime-syntax`
- `syntaxes/02_Extra/cmd-help/syntaxes/cmd-help.sublime-syntax`

while the following still fail:

- `syntaxes/02_Extra/Assembly (ARM).sublime-syntax`
- `syntaxes/02_Extra/Elixir/Regular Expressions (Elixir).sublime-syntax`
- `syntaxes/02_Extra/JavaScript (Babel).sublime-syntax`
- `syntaxes/02_Extra/PowerShell.sublime-syntax`
- `syntaxes/02_Extra/SLS/SLS.sublime-syntax`
- `syntaxes/02_Extra/VimHelp.sublime-syntax`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -60,18 +60,18 @@ dependencies = [
 
 [[package]]
 name = "bit-set"
-version = "0.5.3"
+version = "0.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0700ddab506f33b20a03b13996eccd309a48e5ff77d0d95926aa0210fb4e95f1"
+checksum = "08807e080ed7f9d5433fa9b275196cfc35414f66a0c79d864dc51a0d825231a3"
 dependencies = [
  "bit-vec",
 ]
 
 [[package]]
 name = "bit-vec"
-version = "0.6.3"
+version = "0.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "349f9b6a179ed607305526ca489b34ad0a41aed5f7980fa90eb03160b69598fb"
+checksum = "5e764a1d40d510daf35e07be9eb06e75770908c27d411ee6c92109c9840eaaf7"
 
 [[package]]
 name = "bitflags"
@@ -352,12 +352,13 @@ dependencies = [
 
 [[package]]
 name = "fancy-regex"
-version = "0.11.0"
+version = "0.16.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b95f7c0680e4142284cf8b22c14a476e87d61b004a3a0861872b32ef7ead40a2"
+checksum = "bf04c5ec15464ace8355a7b440a33aece288993475556d461154d7a62ad9947c"
 dependencies = [
  "bit-set",
- "regex",
+ "regex-automata",
+ "regex-syntax",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -352,9 +352,9 @@ dependencies = [
 
 [[package]]
 name = "fancy-regex"
-version = "0.16.1"
+version = "0.16.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "bf04c5ec15464ace8355a7b440a33aece288993475556d461154d7a62ad9947c"
+checksum = "998b056554fbe42e03ae0e152895cd1a7e1002aec800fdc6635d20270260c46f"
 dependencies = [
  "bit-set",
  "regex-automata",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ features = ["metadata"]
 [dependencies]
 yaml-rust = { version = "0.4.5", optional = true }
 onig = { version = "6.5.1", optional = true, default-features = false }
-fancy-regex = { version = "0.16.1", optional = true }
+fancy-regex = { version = "0.16.2", optional = true }
 walkdir = "2.0"
 regex-syntax = { version = "0.8", optional = true }
 plist = { version = "1.3", optional = true }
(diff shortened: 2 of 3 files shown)


keith-hall:
It seems like the syntest example is hanging 🤔 I will try to debug it later, at least it would be nice to see if it is a specific regex it gets stuck on etc

CosmicHorrorDev:
Ah yeah I'm seeing the hang too. It looks like it starts happening with `fancy-regex v0.13.0` if that helps. Help pinning down the problematic regex would be greatly appreciated!

CosmicHorrorDev:
Difference detected! It looks like the parser is getting stuck looping here forever without advancing anything

https://github.com/trishume/syntect/blob/ab2e6c9e1fceb46d94148bdcc3c091204ddd7625/src/parsing/parser.rs#L238-L246

I've got things minimized down to this difference at least

```toml
[package]
name = "fancy-bug"
version = "0.1.0"
edition = "2024"

[dependencies]
fancy-regex = "0.13.0"
fancy_regex_old = { package = "fancy-regex", version = "0.11.0" }
```

```rust
fn main() {
    let regex_str = r"\>";
    let haystack = "<T> void save(T obj);\n";

    let regex = fancy_regex::Regex::new(regex_str).unwrap();
    let captures = regex.captures(haystack);
    let region = Region::init_from_captures(&captures.unwrap().unwrap());

    let regex_old = fancy_regex_old::Regex::new(regex_str).unwrap();
    let captures_old = regex_old.captures(haystack);
    let region_old = Region::init_from_captures_old(&captures_old.unwrap().unwrap());
    println!("New: {region:?}\nOld: {region_old:?}");
}

#[derive(Clone, Debug, Eq, PartialEq)]
pub struct Region {
    positions: Vec<Option<(usize, usize)>>,
}

impl Region {
    fn init_from_captures(captures: &fancy_regex::Captures) -> Self {
        let mut positions = Vec::new();
        for i in 0..captures.len() {
            let pos = captures.get(i).map(|m| (m.start(), m.end()));
            positions.push(pos);
        }
        Self { positions }
    }

    fn init_from_captures_old(captures: &fancy_regex_old::Captures) -> Self {
        let mut positions = Vec::new();
        for i in 0..captures.len() {
            let pos = captures.get(i).map(|m| (m.start(), m.end()));
            positions.push(pos);
        }
        Self { positions }
    }
}
```

```console
$ cargo r
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.01s
     Running `target/debug/fancy-bug`
New: Region { positions: [Some((2, 2))] }
Old: Region { positions: [Some((2, 3))] }
```

keith-hall:
Nice find! I had also just come to the conclusion that the regex causing problems seemed to be `\>` but it wasn't a case of catastrophic backtracking or anything. It's an interesting case, `\>` should actually be just `>` in the syntax definition, but it is escaped instead of quoted as `>` is a special character in Yaml, and then that slash is being taken literally...

CosmicHorrorDev:
`>` isn't a special character for regexes, so I'm guessing it was erroneously escaped. I'll try fixing up the syntax to see if there are any deeper issues

CosmicHorrorDev:
Jinx 😺

CosmicHorrorDev:
@keith-hall since it looks like this is a bug in `fancy-regex`, would you be good with me moving things over to an issue in that repo? I can try to handle fixing things there if I get a bit of guidance (unless you'd rather take things on yourself of course)

CosmicHorrorDev:
Alright switching Ruby's syntax to no longer try to escape `>` and `<` and `make syntest-fancy` finishes ( :tada: ) with this diff

```diff
--- testdata/known_syntest_failures_fancy.txt   2025-05-16 15:47:51.312892736 -0600
+++ -   2025-08-12 22:06:47.954075324 -0600
@@ -1,5 +1,7 @@
 loading syntax definitions from testdata/Packages
 FAILED testdata/Packages/C#/tests/syntax_test_Strings.cs: 38
+FAILED testdata/Packages/Java/syntax_test_java.java: 75
 FAILED testdata/Packages/LaTeX/syntax_test_latex.tex: 1
+FAILED testdata/Packages/Lisp/syntax_test_lisp.lisp: 15
 FAILED testdata/Packages/Markdown/syntax_test_markdown.md: 11
```

~~So, it looks like there are potentially more failed tests for java and lisp now? Not sure what to make of that~~ _(upd: it looks like both of their syntax files also try to escape `>`, so it's probably because of that)_

keith-hall:
> @keith-hall since it looks like this is a bug in `fancy-regex`, would you be good with me moving things over to an issue in that repo? I can try to handle fixing things there if I get a bit of guidance (unless you'd rather take things on yourself of course)

Yes, we can consider it a bug in fancy-regex and open an issue there 👍
I do think it makes sense to prevent syntect from getting stuck, so I plan to look into that

keith-hall:
As for a hint of where to look in fancy-regex: https://github.com/fancy-regex/fancy-regex/pull/121/files#diff-5583e398a8deec11b274d9965eff8b5ade5226c7a020ca214a27d7d07dcc8a29R372-R375 (its a big diff which GitHub doesn't expand by default so in case that link doesn't help, it's parse.rs line 372)

Meanwhile `\<` and `\>` isn't mentioned at https://github.com/kkos/oniguruma/blob/master/doc/RE so probably it should be interpreted as a literal like before... Maybe we need a flag for having the parser behave contrary to Oniguruma in some cases but it shouldn't be the default at least

Just a flag/method for the RegexBuilder should suffice, as opposed to an inline flag controllable from the pattern itself.

CosmicHorrorDev:
Thanks for the `fancy-regex` release @keith-hall. I think things should be all good to go here now. It looks like the only CI failure is an existing issue :tada:

keith-hall (approved):
I haven't got round to making a patch for the jsonnet submodule in bat yet, so that CI would pass, sorry. Pretty sure we can merge this as is though 👍

PR #597, opened 2025-08-16
add stack depth check to the parser

keith-hall:
to prevent infinite looping when a context pushes itself recursively without consuming any input. Fixes a bug discovered in #596

### DIFF
--- a/examples/syntest.rs
+++ b/examples/syntest.rs
@@ -270,8 +270,8 @@ fn test_file(
             }
             if out_opts.debug && !line_only_has_assertion {
                 println!(
-                    "-- debugging line {} -- scope stack: {:?}",
-                    current_line_number, stack
+                    "-- debugging line {} -- scope stack: {:?}, -- parse state: {:?}",
+                    current_line_number, stack, state
                 );
             }
             let ops = state.parse_line(&line, ss).unwrap();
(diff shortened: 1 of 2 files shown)


Enselic (approved):
new regression test + all existing test passing = ✅

keith-hall:
I wonder if this broke something in certain versions of rustc:
https://github.com/sharkdp/bat/pull/3410
https://github.com/sharkdp/bat/actions/runs/18060101420/job/51395153259?pr=3410#step:4:311

> 
```
error[E0716]: temporary value dropped while borrowed
   --> /home/runner/.cargo/registry/src/index.crates.io-6f17d22bba15001f/syntect-5.3.0/src/parsing/parser.rs:466:19
    |
463 |           let (regex, can_cache) = match (match_pat.has_captures, captures) {
    |  __________________________________-
464 | |             (true, Some(captures)) => {
465 | |                 let (region, s) = captures;
466 | |                 (&match_pat.regex_with_refs(region, s), false)
    | |                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ creates a temporary value which is freed while still in use
467 | |             }
    | |             - temporary value is freed at the end of this statement
468 | |             _ => (match_pat.regex(), true),
469 | |         };
    | |_________- borrow later used here
    |
    = note: consider using a `let` binding to create a longer lived value
```

Enselic:
We have a min rustc version rule that says "we bump min supported rust version whenever there is the slightest MSRV-related problem", so the easiest way to resolve this is probably to bump MSRV.

PR #600, opened 2025-09-17
Replace unmaintained `yaml-rust` dependency with `yaml-rust2`

praseodym:
Replace the unmaintained `yaml-rust` dependency with `yaml-rust2`, as recommended by [RUSTSEC-2024-0320](https://rustsec.org/advisories/RUSTSEC-2024-0320).

Even though the [dependency is renamed in `Cargo.toml`](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#renaming-dependencies-in-cargotoml), the public API still changes as can be seen in the updated `tests/snapshots/public-api.txt`.

Supersedes #544 (has merge conflicts) and #555 (fails CI).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -27,6 +27,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "arraydeque"
+version = "0.5.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7d902e3d592a523def97af8f317b08ce16b7ab854c1985a0c671e6f15cebc236"
+
+[[package]]
 name = "atty"
 version = "0.2.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -377,6 +383,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3f9eec918d3f24069decb9af1554cad7c880e2da24a9afd88aca000531ab82c1"
 
 [[package]]
+name = "foldhash"
+version = "0.1.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d9c4f5dac5e15c24eb999c26181a6ca40b39fe946cbe4c263c7209467bc83af2"
+
+[[package]]
 name = "form_urlencoded"
 version = "1.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


praseodym:
It looks like the CI

praseodym:
The bat syntax regression test fails:

`
[bat error]: /home/runner/work/syntect/syntect/bat/assets/syntaxes/02_Extra/Jsonnet/jsonnet.sublime-syntax: Invalid YAML file syntax: while parsing a quoted scalar, found unknown escape character at byte 2838 line 93 column 14
`

It looks like the [`jsonnet.sublime-syntax`](https://github.com/gburiola/sublime-jsonnet-syntax/blob/e2c917ed7d6020e2ba6c18b06c7278c094db03f7/jsonnet.sublime-syntax#L93) file has actual invalid syntax (`\'`) on line 93 and 117. Online tools like https://www.yamllint.com/ and https://yamlchecker.com/ also report the same error.

This PR introduces a regression for bat, but only because bat's regression test includes a file with seemingly invalid syntax. How do we resolve this?

CosmicHorrorDev:
Oop, I think merging this was a breaking change. I was hoping to get out a `5.3.0` release 😅

keith-hall:
My bad, I probably rushed a bit because I felt that we should probably have upgraded this dependency a while ago... Would it make sense to push a commit reverting this PR, make the minor release, then revert the revert? 😅

CosmicHorrorDev:
I think that should work out totally fine 👍

keith-hall:
Just to be clear, I don't have cargo publish permissions for syntect so it will need one of the other maintainers to perform the release

Enselic:
@keith-hall I can add you as owner for https://crates.io/crates/syntect so you can make releases on your own. If you'd like?

CC other owners @trishume @robinst

keith-hall:
Sure, that could be convenient, thanks 👍 do we have any documentation on the release process/checklist for syntect? I had a quick look but didn't see anything

Enselic:
I invited you as owner now.

I don't think there is a checklist, but I think doing this is sufficient:
* Ensure CI passes for the commit you want to publish
* checkout that commmit locally
* cargo publish
* git tag and push the tag

My recommendation is to not have any long-term auth token around. Best would be if we enabled [trusted publishing](https://crates.io/docs/trusted-publishing). Here is a minimal example from one of my projects: https://github.com/Enselic/snapshot-testing/blob/main/.github/workflows/release.yml. But that is a nice-to-have and not blocking any releases.

keith-hall:
Thanks, release has been made. I noticed we have a [Release workflow in syntect](https://github.com/trishume/syntect/blob/master/.github/workflows/Release.yml), but when I tried it, it didn't work: 

> The workflow is not valid. .github/workflows/Release.yml (Line: 25, Col: 40): Invalid input, one_time_crates_io_token_secret is not defined in the referenced workflow.

So I published manually for now. And raised a new PR to re-apply yaml-rust2: https://github.com/trishume/syntect/pull/602

PR #601, opened 2025-09-26
Prepare for a 5.3.0 release

### LINKED ISSUE #599: new release
Hi thank you for creating and maintaining this package 😄, would it be possible to create a new release? This would enable packages that depend on this package to use the fixes that have been added since Feb 2024? The specific fix that I am interested in mind for my use case is #548 as it fixes  [typst issue 4421](https://github.com/typst/typst/issues/4421) .

CosmicHorrorDev:
Resolves #599

Updated the changelog with the new changes, tweaked the installation docs on the readme, and reverted the breaking change with `yaml-rust` -> `yaml-rust2` with it being part of the public API

I don't think that there were any changes that would require updating the embedded syntax/theme/metadata assets

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,80 @@
 # Changelog
 
+## [Version 5.3.0](https://github.com/trishume/syntect/compare/v5.2.0...v5.3.0) (UNRELEASED)
+
+### Improvements
+
+- Add `HighlightLines::from_state()` constructor [#569]
+
+### Fixes
+
+- Ignore UTF-8 BOM on syntax detection [#530]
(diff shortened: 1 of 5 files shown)


keith-hall (approved):
Thanks!

keith-hall:
Release 5.3.0 has been published. I noticed during publishing the following warning:

> warning: package `log v0.4.24` in Cargo.lock is yanked in registry `crates-io`, consider updating to a version that is not yanked

it seems like it isn't a concern, but we should probably upgrade it: https://github.com/rust-lang/log/issues/661#issuecomment-2585472287

PR #603, opened 2025-10-17
feat: Implement Pop N (#323)

mousetail:
One of the features introduced in sublime 4075 was to give a number to POP multiple frames from the context stack at once.

This should implement the feature.

I am not 100% sure what some of the optimization code regarding the captures and loop detection is but it seems like it should not matter.

### DIFF
--- a/src/parsing/parser.rs
+++ b/src/parsing/parser.rs
@@ -1086,6 +1086,39 @@ contexts:
     }
 
     #[test]
+    fn can_parse_prototype_that_pops_multiple_context() {
+        let syntax = r#"
+name: test
+scope: source.test
+contexts:
+  prototype:
+    - match: "!"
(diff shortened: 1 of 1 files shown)


keith-hall:
I guess the change affects the binary dump format. Running `make packs` and committing the results should get CI to pass.

mousetail:
> I guess the change affects the binary dump format. Running `make packs` and committing the results should get CI to pass.

I updated the packs and also public API snapshots. I guess that means this is a potentially breaking change

keith-hall:
The next release will be breaking anyway because we have changed the yaml parser crate, so now is the perfect time for it! 👍

It would be nice to have a simple test to prove that popping multiple contexts at a time yields the correct scopes.

PR #604, opened 2025-10-29
Add optional support sublime color scheme files via external crate

### LINKED ISSUE #244: Problem loading sublime-color-scheme files
I was trying to use the official [Monokai.sublime-color-scheme](https://raw.githubusercontent.com/twolfson/sublime-files/master/Packages/Color%20Scheme%20-%20Default/Monokai.sublime-color-scheme) with syntect but it seems this file-format is not supported or maybe it's just me doing something fishy.

I've tried to load it both as json or as xml (ie: used this converter [https://www.freeformatter.com/json-to-xml-converter.html](https://www.freeformatter.com/json-to-xml-converter.html)) but when feeding it with `ThemeSet::get_theme` it will crash.

On the other hand, if using a proper thTheme such as [monokai.tmTheme](https://github.com/richrd/suplemon/blob/master/suplemon/themes/monokai.tmTheme) there won't be any problem.

Lepidopteran:
It's been a while for any movement on issue #244, So I've come here to discuss adding a optional dependency on [`sublime-color-scheme`](https://crates.io/crates/sublime-color-scheme) crate to add the ability to parse `.sublime-color-scheme` files to syntect.

Any feedback is greatly appreciated!

Resolves #244

## Results

I've tested with a couple themes that use the `sublime-color-scheme` file format with great results taken from my [comment](https://github.com/trishume/syntect/issues/244#issuecomment-2696313343) on the issue:

### Kanagawa
Image

### Catppuccin
Image

### Nord
Image

### Monokai
Image

### Gruvbox Material
Image

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -70,7 +70,7 @@ default-syntaxes = ["parsing", "dump-load"]
 default-themes = ["dump-load"]
 
 # Support for parsing .sublime-color-scheme files
-sublime-color-scheme = ["sublime-color-scheme"]
+sublime-color-scheme = ["dep:sublime-color-scheme"]
 
 html = ["parsing"]
 # Support for parsing .tmTheme files and .tmPreferences files


Lepidopteran:
Im not fully sure, if I did something wrong with the Build/Test Check

Issue #541, opened 2024-05-02
Consider delaying `parse_regex`'s transformations


CosmicHorrorDev:
`syntect` performs a few transformations to syntax definition's regexes while loading everything. The most notable one in terms of the final syntax set's size being the expansion of the various `variables` that can be included with a syntax definition. This expansion can be delayed to when the regex is compiled instead which can significantly reduce the embedded syntax definitions (my notes from testing with generating a syntax set without applying the transformations to regexes say that it took the total size of all of the `LazyContext`s that are stored in `bat`'s syntax set from 720 to 673 KiB)

PR #610, opened 2026-03-04
Implement sublime-syntax Build 4075 features

stefanobaghino:
Makes progress towards #323.

This adds support for the remaining `.sublime-syntax` features introduced in Sublime Text Build 4075. `hidden_file_extensions` and `pop: N` were already handled; this PR covers the rest.

### `extends` (single syntax inheritance)

A child syntax can now declare `extends: Packages/Foo/Foo.sublime-syntax` to inherit all contexts and variables from a parent. The builder resolves extends chains (including transitive ones) in a fixed-point loop before the existing linking phase:

1. Merge variables -- parent first, child overrides win.
2. Merge contexts -- parent contexts absent from the child are inherited as-is; contexts the child redefines replace the parent version unless `meta_prepend` or `meta_append` is used.
3. Re-resolve every regex that stored a raw string, applying the merged variable map through the full pipeline (variable substitution, POSIX class replacement, newline handling).
4. If a child now has a `main` context but lacked `__start`/`__main`, the initial contexts are generated.

Missing or circular parents produce a warning on stderr and are skipped.

### `meta_prepend` / `meta_append`

When a child context uses `meta_prepend: true`, its patterns are placed before the parent's patterns. `meta_append: true` places them after. Parent meta scopes and clear_scopes are inherited when the child does not set them.

### `apply_prototype: true`

A new `Pattern::IncludeWithPrototype` variant. When iterating patterns, the external syntax's prototype is walked alongside the included context (respecting `meta_include_prototype`).

### `version: 2`

Parsed from YAML and stored on both `SyntaxDefinition` and `SyntaxReference`. The parser reads it at match time and applies two behavioral fixes in `push_meta_ops`:

- **set + meta_content_scope**: the old context's `meta_content_scope` is no longer applied to the matched text.
- **multiple push + clear_scopes**: only the topmost context's `clear_scopes` is used instead of summing all of them.

### Raw regex storage

Each `MatchPattern` now carries an optional `raw_regex_str` -- the regex as written in YAML before variable resolution. This is needed so that extends resolution can re-resolve parent regexes with overridden variables.

## Files touched

| File | What |
|---|---|
| `src/parsing/syntax_definition.rs` | New fields (`extends`, `version`, `raw_regex_str`, `merge_mode`), new enum `ContextMergeMode`, new pattern variant `IncludeWithPrototype` |
| `src/parsing/yaml_load.rs` | Parse `extends`, `version`, `meta_prepend`, `meta_append`, `apply_prototype`; store raw regex; re-resolution helpers |
| `src/parsing/syntax_set.rs` | `resolve_extends` phase in `build()`; `find_parent_index` helper; updated linking/backref/prototype code for new pattern variant |
| `src/parsing/parser.rs` | `current_syntax_version` helper; version-conditional branches in `push_meta_ops` |
| `assets/*.packdump` | Regenerated (`make packs`) |
| `tests/snapshots/public-api.txt` | Updated |

## Not included

- Multiple inheritance (`extends` as a list) -- rare in practice; deferred.
- Three of the five `version: 2` edge cases (embed_scope replacement, set+clear_scopes, embed escape meta scopes) -- they need more investigation into the exact Sublime behavior.

## Test plan

- 17 new unit tests covering parsing of each new key, extends resolution (inherit, override, prepend, append, variable override, missing parent), apply_prototype, version field propagation.
- All 146 existing + new tests pass (`cargo test`).
- Verified with `--features metadata` (129 pass) and `--no-default-features --features regex-fancy` (114 pass).

### DIFF
--- a/src/parsing/parser.rs
+++ b/src/parsing/parser.rs
@@ -563,7 +563,10 @@ impl ParseState {
     fn current_syntax_version(&self, syntax_set: &SyntaxSet) -> u32 {
         if let Some(level) = self.stack.last() {
             let syntax_index = level.context.syntax_index;
-            syntax_set.syntaxes().get(syntax_index).map_or(1, |s| s.version)
+            syntax_set
+                .syntaxes()
+                .get(syntax_index)
+                .map_or(1, |s| s.version)
         } else {
             1
         }
@@ -670,8 +673,8 @@ impl ParseState {
                                 // v2: set excludes parent meta_content_scope from matched text
                                 num_to_pop += cur_context.meta_scope.len();
                             } else {
-                                num_to_pop +=
-                                    cur_context.meta_content_scope.len() + cur_context.meta_scope.len();
+                                num_to_pop += cur_context.meta_content_scope.len()
+                                    + cur_context.meta_scope.len();
                             }
                         }
 
--- a/src/parsing/syntax_set.rs
+++ b/src/parsing/syntax_set.rs
@@ -860,9 +860,7 @@ impl SyntaxSetBuilder {
                 }
 
                 // If child now has a main context but didn't have __start/__main, add them
-                if child.contexts.contains_key("main")
-                    && !child.contexts.contains_key("__start")
-                {
+                if child.contexts.contains_key("main") && !child.contexts.contains_key("__start") {
                     // Re-add initial contexts now that main is available
                     let mut scope_repo = crate::parsing::scope::lock_global_scope_repo();
                     let top_level_scope = child.scope;
@@ -996,10 +994,19 @@ impl SyntaxSetBuilder {
                         }
                     }
                 }
-                Pattern::Include(ref reference)
-                | Pattern::IncludeWithPrototype(ref reference) => match reference {
-                    ContextReference::Named(ref s) => {
-                        if let Some(id) = syntax_context_ids.get(s) {
+                Pattern::Include(ref reference) | Pattern::IncludeWithPrototype(ref reference) => {
+                    match reference {
+                        ContextReference::Named(ref s) => {
+                            if let Some(id) = syntax_context_ids.get(s) {
+                                Self::recursively_mark_no_prototype(
+                                    id,
+                                    syntax_context_ids,
+                                    all_contexts,
+                                    no_prototype,
+                                );
+                            }
+                        }
+                        ContextReference::Direct(ref id) => {
                             Self::recursively_mark_no_prototype(
                                 id,
                                 syntax_context_ids,
@@ -1007,17 +1014,9 @@ impl SyntaxSetBuilder {
                                 no_prototype,
                             );
                         }
+                        _ => (),
                     }
-                    ContextReference::Direct(ref id) => {
-                        Self::recursively_mark_no_prototype(
-                            id,
-                            syntax_context_ids,
-                            all_contexts,
-                            no_prototype,
-                        );
-                    }
-                    _ => (),
-                },
+                }
             }
         }
     }
@@ -1811,7 +1810,10 @@ mod tests {
             .parse_line("override\n", &ss)
             .expect("parse failed");
         // Should match child's 'override' keyword, not base's ident
-        let expected = (0, ScopeStackOp::Push(Scope::new("keyword.override").unwrap()));
+        let expected = (
+            0,
+            ScopeStackOp::Push(Scope::new("keyword.override").unwrap()),
+        );
         assert_ops_contain(&ops, &expected);
     }
 
@@ -1880,9 +1882,7 @@ mod tests {
         let syntax = ss.find_syntax_by_name("Child").unwrap();
         let mut parse_state = ParseState::new(syntax);
         // 'abc' should still match base's ident pattern (it comes first since child is appended)
-        let ops = parse_state
-            .parse_line("abc\n", &ss)
-            .expect("parse failed");
+        let ops = parse_state.parse_line("abc\n", &ss).expect("parse failed");
         let expected = (0, ScopeStackOp::Push(Scope::new("variable.base").unwrap()));
         assert_ops_contain(&ops, &expected);
     }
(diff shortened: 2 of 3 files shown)


keith-hall (approved):
Wow, nice, thanks for taking this on - looks good to me!

keith-hall:
> ## Not included
> 
>     * Multiple inheritance (`extends` as a list) -- rare in practice; deferred.

We will for sure want to support this in a future PR, in order to support TSX etc. https://github.com/sublimehq/Packages/blob/master/JavaScript/TSX.sublime-syntax#L7-L9

CosmicHorrorDev:
the raw regex storage should put us closer to #541 which is nice

---

funny seeing that the failure is a formatting difference with the MSRV. formatting could probably be dropped from that job given that MSRV is more of a "it compiles" thingy

stefanobaghino:
> funny seeing that the failure is a formatting difference with the MSRV. formatting could probably be dropped from that job given that MSRV is more of a "it compiles" thingy

I still fixed it in https://github.com/trishume/syntect/pull/610/commits/eb4c55442017d5d4e9e8d57ea07a02ed5524f35a, seemed sensible. 🙂

keith-hall:
> * Verified with `--features metadata` (129 pass) and `--no-default-features --features regex-fancy` (114 pass).

Purely out of curiosity, I'd be interested to see the results when running with `fancy-regex`'s main branch - there's a bunch of Oniguruma compatibility fixes I have made but not released yet. (Waiting for https://github.com/fancy-regex/fancy-regex/pull/230 to be reviewed/merged mainly)

stefanobaghino:
I just wanted to let you know that I'm doing this because 2 years ago I couldn't get syntax highlighting for a preview feature of Java that has been removed since (but that was somehow already supported by the Sublime syntax definitions). 😄

Enselic:
I've skimmed the code through and have no objections as long as I am not seen as responsible for maintaining the code over time :)

If someone didn't already try, it would be good to test this implementation against a bunch of real-world syntaxes with bat that we couldn't use before because of the missing features. So we know it works for real, and not just with unit tests.

If I were to review this I would prefer to split this into smaller PRs instead of implementing all features in one giant PR. But I don't really want to do a detailed review :) so if others are fine with the code and functionality that split is not needed as far as I'm personally concerned.

Enselic:
One more thing: If we can implement this without changing the public API surface of syntect that would be even better, but I understand if that wouldn't work.

stefanobaghino:
> > * Verified with `--features metadata` (129 pass) and `--no-default-features --features regex-fancy` (114 pass).
> 
> Purely out of curiosity, I'd be interested to see the results when running with `fancy-regex`'s main branch - there's a bunch of Oniguruma compatibility fixes I have made but not released yet. (Waiting for [fancy-regex/fancy-regex#230](https://github.com/fancy-regex/fancy-regex/pull/230) to be reviewed/merged mainly)

I pointed the `fancy-regex` dependencies to both the latest state of the `main` and `subroutine_support` branches and the test results always matched the current state.

stefanobaghino:
@Enselic Thanks for your feedback. To be clear, my hope here is to be helpful and from your message I'm under the impression that this contribution might be counterproductive. In the interest of making this contribution sustainable, I'd like to make sure there's alignment. I apologize in advance if I'm asking any dumb question, but I hope those ensure that this contribution is welcome:
1. Who maintains this library? To be clear, I'm not expecting anyone specifically to be in such a position, I understand sometimes open source libraries are passion projects that became too successful.
2. Is there something in particular I can do to make this contribution easier to be acceptable for maintainers to take over in the future? Additional documentation? Splitting it into smaller chunks that are easier to review and digest? I'm happy to do either, or both, and/or take suggestions for further improvements.

I hope these questions don't come across the wrong way, all I hope is for this contribution to be genuinely helpful. 🙂 All I really want is for `delta` to highlight my Java code using the most recent syntax changes. 😄

jrappen:
the SublimeHQ devs and most of the more active 3rd-party plugin devs are on Discord:

<https://discord.gg/D43Pecu>

in case you had questions with regard to the currently-not-yet-implemented stuff.

jrappen:
Any plans to address #271?

stefanobaghino:
> Any plans to address #271?

Not as part of this PR, but to the best of my understanding that is unrelated, right?

jrappen:
This PR does not close #323, it only partially addresses it.

Multiple inheritance might be rare, but I know for certain, I've got an open PR to the default packages for JSON where I use a list value for `extends`. I haven't double checked, but there might be others in defaults. The thing about list values is (1) all entries need to extend the same base and (2) order matters, where changes by later entries override changes by prior entries to the common base.

michaelblyons:
Would it be nice to have multiple inheritance? Yes. Should it block this? I say no. And the same for `branch` support.

A huge percentage of extended syntaxes are just tweaks to a base language to fix interpolations when it's embedded. Holding that up for TSX and stale PRs for C and JSON seems like a bad trade-off. If the maintainers want the PR split or API stable, that's one thing. But dropping new functionality requirements is quite another. 

I wish I had the familiarity with syntect to review the code itself, but I don't.

stefanobaghino:
> This PR does not close #323, it only partially addresses it.

Fair point, I reworded the PR description.

I'd appreciate the maintainers letting me know whether they'd rather merge as is, apply any change they see fit, split into smaller PRs, or get #323 fully implemented. Happy to act on any guidance.

jrappen:
PR looks good from here.

I (or any maintainer) can mark whatever is partially resolved through this PR in the list at the top of #323 as "done".

keith-hall:
I think it is fine without splitting into smaller PRs. I suspect it should be possible to de-duplicate some of the code, but as syntect hasn't seen much love recently, I think it is worth merging as it is. We can always iterate on it later. (Gatekeeping at this point seems like it would detract from the goal and demotivate contributors.)

I was hoping to bump the Packages submodule to see how the syntax tests perform with the changes, but until we get branch point support, the failures would be too noisy. I unfortunately don't currently have time to test manually with syntaxes using v2 and inheritance etc.

In either case, this PR doesn't break any existing functionality and it has unit tests, so I am inclined to merge.

Edit: forgot to say, I'm pretty sure the public API changes are necessary and make sense 👍

stefanobaghino:
Continuing the work in https://github.com/trishume/syntect/pull/612.

jrappen:
image

---

CI on `master` fails as a result of this PR.

stefanobaghino:
> CI on `master` fails as a result of this PR.

I saw that when pushing my changes for the follow-up with #612. Not sure of whether the problem is caused strictly by what has been as part of this PR, since this check passed on CI to the best of my understanding. In any case, #612 should take care of it, specifically https://github.com/trishume/syntect/pull/612/changes/a2d182315b424fb3b67db9068ed8718269cfbb26.

stefanobaghino:
After having a second look, I just realize it was indeed in this PR that that check started failing. Sorry for not fixing that before, if I had noticed that I would have pushed a patch. As mentioned though, I ended up fixing the issue in #612 to make sure I had a green branch. 🙂

PR #612, opened 2026-03-10
Implement remaining sublime-syntax Build 4075 features

stefanobaghino:
Make progress on #323:
- v2 set + clear_scopes: allow set operations to apply clear_scopes
- v2 embed_scope replacement: embed_scope replaces embedded syntax scope
- Multiple inheritance: extends accepts a YAML list of syntax paths

### DIFF
--- a/src/parsing/syntax_set.rs
+++ b/src/parsing/syntax_set.rs
@@ -3018,9 +3018,10 @@ mod tests {
         };
 
         let syntax_definitions = vec![syntax];
-        let path_syntaxes = vec![
-            ("Packages/Test/syntaxes/Child.sublime-syntax".to_string(), 0usize),
-        ];
+        let path_syntaxes = vec![(
+            "Packages/Test/syntaxes/Child.sublime-syntax".to_string(),
+            0usize,
+        )];
         let mut name_to_index = HashMap::new();
         name_to_index.insert("Child".to_string(), 0);
 
@@ -3049,7 +3050,11 @@ mod tests {
             &syntax_definitions,
             &name_to_index,
         );
-        assert_eq!(result, Some(0), "bare filename should match via name lookup");
+        assert_eq!(
+            result,
+            Some(0),
+            "bare filename should match via name lookup"
+        );
 
         // A non-matching relative path should return None
         let result = SyntaxSetBuilder::find_parent_index(


jrappen:
Right now I don't have time to look into your issue with failed CI.

With regard to multiple inheritance think of it like this. Say you have a JSON syntax, and then a bunch of helper syntaxes where the first overrides JSON to add optional trailing commas in arrays and objects, the second adds support for single-quoted strings, the third adds support for null as an additional value, the fourth adds support for line and block comments. Now JSON has syntax flavors like GeoJSON, HJSON, JSON5, JSONC (with Comments), JSON.NET. So for JSONC as an example you would need to do the following: (1) add multiple inheritance where (2) the comments helper syntax goes into the list of syntaxes for the `extends` key before (3) the trailing commas helper syntax. `syntect` should produce an error (as does Sublime) if (1) not all entries in the list extend from the same syntax or (2) you add the base syntax to the list, which is the same as (1) but I just wanted to reiterate that explicitely in case it wasn't obvious.

Not sure if that's your issue with CI failing, otherwise just an FYI. Please make sure syntect produces verbose-enough errors appropriately.

stefanobaghino:
> Right now I don't have time to look into your issue with failed CI.

There's no need, I only added the comment to make sure it was clear why I made the change. The error is the same that you raised in https://github.com/trishume/syntect/pull/610#issuecomment-4038231504 and that I fixed in https://github.com/trishume/syntect/pull/612/changes/a2d182315b424fb3b67db9068ed8718269cfbb26.

> With regard to multiple inheritance think of it like this. Say you have a JSON syntax, and then a bunch of helper syntaxes where the first overrides JSON to add optional trailing commas in arrays and objects, the second adds support for single-quoted strings, the third adds support for null as an additional value, the fourth adds support for line and block comments. Now JSON has syntax flavors like GeoJSON, HJSON, JSON5, JSONC (with Comments), JSON.NET. So for JSONC as an example you would need to do the following: (1) add multiple inheritance where (2) the comments helper syntax goes into the list of syntaxes for the `extends` key before (3) the trailing commas helper syntax. `syntect` should produce an error (as does Sublime) if (1) not all entries in the list extend from the same syntax or (2) you add the base syntax to the list, which is the same as (1) but I just wanted to reiterate that explicitely in case it wasn't obvious.

Are you saying this pointing to some specific flaw in the PR?

jrappen:
Reading through the PR, I noticed:

- [x] some behavior with regard to addressing syntax v1 vs. v2 changed. I saw you addressing the stacking of `meta_scope`s. The complete list is here as a reference: <https://www.sublimetext.com/docs/syntax.html#compatibility>. I haven't checked if you addressed all, yet.
- [x] I saw you checking if each syntax in an `extends` list value exists. I didn't see you checking if whatever they extend from themselves:

  - [x] exists
  - [x] and is the same for all
    (syntax > inheritance > limitations)
    <https://www.sublimetext.com/docs/syntax.html#limitations>
  - [x] and has the same syntax version between the parents and their common base
    (syntax > inheritance > limitations)
  - [x] and has the same syntax version between the parents and the source
    (syntax > inheritance > limitations)

  compare <https://www.sublimetext.com/docs/syntax.html#multiple-inheritance>.

stefanobaghino:
Thanks for the detailed review!

> some behavior with regard to addressing syntax v1 vs. v2 changed. I saw you addressing the stacking of `meta_scope`s. The complete list is here as a reference. I haven't checked if you addressed all, yet.

I went through the compatibility list after your comment. The latest two commits add tests covering the remaining gaps I found and fix the ones that were failing: specifically, `set` in v2 no longer applies the parent context's `meta_content_scope` to the matched text, and the escape pattern in a v2 `embed` now correctly sees the `embed_scope` (which is the context's `meta_content_scope`). The v1/v2 differences that were already handled (e.g. `clear_scopes` stacking on multi-push, capture group scoping) continue to pass.

> I saw you checking if each syntax in an `extends` list value exists. I didn't see you checking if whatever they extend from themselves: exists, and is the same for all, and has the same syntax version between the parents and their common base, and has the same syntax version between the parents and the source

The "exists" case was already covered: unresolvable parents cause the child to remain in the unresolved set, which produces a warning at the end of the fixed-point loop. The other three were missing — the latest commits add them:

- Same root for all parents: if parents don't share a common ancestor syntax, the extends is skipped with a warning.
- Version consistency: the version check is enforced at each resolution step — parents must have the same version as the child extending them. Since this applies transitively throughout the chain, it covers both the parents-vs-common-base and source-vs-parents cases.

In all error cases syntect emits a warning (via `eprintln!`) and skips applying the `extends`, leaving the child syntax in its unextended state.

jrappen:
I'll try to look through this later today when I got more time on hand than right now.

---

Skimping through the PR I noticed the format of paths used when extending. I'm not sure if you are aware that one can use local (to the same package) relative paths as well. So as an example, if we're still sticking to the JSON one given above:

```text
- Packages / JSON
    - syntaxes
        - json-feature-comments.sublime-syntax
        - json-feature-trailing-commas.sublime-syntax
    - JSON.sublime-syntax
    - JSONC.sublime-syntax
```

things would look like this:

```yaml
# Packages/JSON/JSONC.sublime-syntax

name: JSONC (with Comments)
version: 2
extends:
  - "Packages/JSON/syntaxes/json-feature-comments.sublime-syntax"        # v2, extends JSON
  - "Packages/JSON/syntaxes/json-feature-trailing-commas.sublime-syntax" # v2, extends JSON
```

or like this:

```yaml
# Packages/JSON/JSONC.sublime-syntax

name: JSONC (with Comments)
version: 2
extends:
  - "syntaxes/json-feature-comments.sublime-syntax"        # v2, extends JSON
  - "syntaxes/json-feature-trailing-commas.sublime-syntax" # v2, extends JSON
```

Not sure if this is handled, might be for another issue and PR, though. Just wanted to document it. If you have time and energy to address this too, awesome.

stefanobaghino:
I hadn't thought about local relative paths explicitly. I looked into it, the current `find_parent_index` implementation already handles this: it normalizes the `extends` value and checks whether any loaded syntax path **ends with** it (after prepending `/`). So a relative path like `syntaxes/json-feature-comments.sublime-syntax` will match against the full loaded path `Packages/JSON/syntaxes/json-feature-comments.sublime-syntax`. I added a test in [`74e4476`](https://github.com/trishume/syntect/pull/612/commits/74e44763844ba23b0f200b6228439a77feec5d77) to make sure this keeps working.

stefanobaghino:
Is there some way in which I can facilitate the review process? 🙂

jrappen:
My comments above were addressed, marked them as "resolved". Looks good so far.

jrappen:
@keith-hall maybe you can take a look if you have a min?

keith-hall (approved):
LGTM

PR #614, opened 2026-03-21
Implement branch_point/branch/fail backtracking

### LINKED ISSUE #271: sublime-syntax branch_point support

stefanobaghino:
Closes #271.

## Summary

- Implements the `branch_point`, `branch`, and `fail` directives from the
  Sublime Text syntax definition spec, enabling speculative parsing with
  backtracking when a chosen branch turns out to be wrong.
- Handles cross-line backtracking: when `fail` fires on a later line than
  its `branch_point`, the parser re-parses intermediate lines under the
  corrected alternative and surfaces the corrected ops via a new
  `ParseLineOutput.replayed` field.
- **Breaking API change**: `ParseState::parse_line` now returns
  `Result<ParseLineOutput, ParsingError>` instead of a bare `Vec`.
  Migration: append `.ops` to the old call site. Callers that care about
  cross-line accuracy should also inspect `.replayed`.

## Design decisions

- Ops are always returned immediately (not withheld while a branch point
  is active). Only raw line strings are buffered for potential replay.
  This keeps the common-case caller contract unchanged.
- The inner parsing loop is extracted into `parse_line_inner` so that
  `handle_fail` can re-parse buffered lines without re-entrancy issues
  with the outer `parse_line` bookkeeping.
- Branch points older than 128 lines are expired, matching Sublime Text
  behavior and bounding memory usage.

## Known limitations

- Callers that maintain a `ScopeStack` incrementally will have applied
  the original (potentially wrong) ops for buffered lines. When
  `replayed` is non-empty, they must rewind and re-apply. syntect's own
  highlighter modules use `.ops` only and are unaffected.

## Test plan

- [x] `cargo test` — all 142 unit tests + 13 doc-tests pass
- [x] New `branch_cross_line_backtrack` test verifies cross-line replay
- [x] Public API snapshot updated and passing

### DIFF
--- a/examples/syntest.rs
+++ b/examples/syntest.rs
@@ -354,7 +354,11 @@ fn test_file(
             }
             let stack_before = stack.clone();
             let output = state.parse_line(&line, ss).unwrap();
-            let ParseLineOutput { ops, replayed, warnings } = output;
+            let ParseLineOutput {
+                ops,
+                replayed,
+                warnings,
+            } = output;
 
             for warning in &warnings {
                 eprintln!("Warning: {}", warning);
--- a/src/parsing/parser.rs
+++ b/src/parsing/parser.rs
@@ -311,7 +311,11 @@ impl ParseState {
 
         let warnings = std::mem::take(&mut self.warnings);
 
-        Ok(ParseLineOutput { ops, replayed, warnings })
+        Ok(ParseLineOutput {
+            ops,
+            replayed,
+            warnings,
+        })
     }
 
     /// Returns `true` when the parser is inside a `branch_point` and the
@@ -2890,7 +2894,9 @@ contexts:
             out_fail.replayed
         );
         assert!(
-            all_warnings.iter().any(|w| w.contains("expired") && w.contains("bp")),
+            all_warnings
+                .iter()
+                .any(|w| w.contains("expired") && w.contains("bp")),
             "expected a warning about branch point expiry, got: {:?}",
             all_warnings
         );


stefanobaghino:
I have a fix locally for the clippy issues.

The regression test failure seem benign, as Julia's syntax seem to have been fixed. 🙂 Having a look at that, happy to take input though, if you have any.

stefanobaghino:
I believe https://github.com/trishume/syntect/pull/614/commits/406273b9201f0330fb87587c35acd6da0743b93a and https://github.com/trishume/syntect/pull/614/commits/591c638755c698562cad86b32847c9937464f4bd should address the bat regression test failures ([diff](https://github.com/trishume/syntect/pull/614/changes/278a2b09367d190b841a8ecabc1448ee02e2933c..591c638755c698562cad86b32847c9937464f4bd)).

stefanobaghino:
Thanks for the review, @keith-hall! Here's a summary of what I've done to address your comments:

**Removed comments in `exec_pattern`** — restored in e371486. The `println!`s were intentionally removed but the explanatory comments got caught in the crossfire. They're back now.

**syntest should evaluate assertions only after the parser commits to a branch** — fully agreed, and this turned out to be a meaningful body of work:

- 81bd452 handles replayed ops in syntest so that cross-line backtracking produces correct scopes (when a `fail` rewinds to a previous line, the ops from the speculative branch are discarded and replaced by the winning branch's ops).
- f7d1fb6 defers assertion output until the parser commits, using the new `is_speculative()` API. While the parser is still exploring branches, syntest buffers any assertion failures instead of reporting them immediately. Once the branch is resolved (either committed or failed and retried), it evaluates assertions against the final scopes. This also serves as an example of how to get the committed scopes, as you suggested.

On the testing side, I also ran `cargo mutants` against the changed code in `parser.rs` and wrote targeted tests to kill surviving mutations across `find_best_match`, `search`, `exec_pattern`, `push_meta_ops`, and `handle_fail` (978b5b2, 252b4a4, a8953e3). The only remaining survivors are equivalent mutations (e.g. constant offsets in secondary sort keys, guards for unreachable states) or impractical ones (depth-100 stack limits, internal cache behavior).

stefanobaghino:
> LGTM (with the caveat that I haven't run the syntest example or anything to verify, but I imagine that will come next: as part of bumping the Sublime HQ Packages submodule and seeing what the state is).

I was planning to work on #222 and then bump the packages and identify the gaps, if that works for you.

keith-hall:
Even better, thank you so much for your hard work reviving syntect btw!

stefanobaghino:
> Even better, thank you so much for your hard work reviving syntect btw!

I'm taking advantage of a sabbatical I had to take for personal reasons to address this mildly uncomfortable feeling of not seeing Java highlighted correctly in `delta`. 😂

jrappen:
* Could you maybe rename the `ops` var where you fixed `ops.ops`? maybe a personal preference, though.
* Are you taking care of this limit? If possible, please add some output not as "error" log level obviously but maybe as "warning" or "info" log level. A test for this would also be appreciated.
  <https://www.sublimetext.com/docs/syntax.html#match-patterns>

  image

michaelblyons:
The 128 line limit is handled explicitly. Not sure about logging.

stefanobaghino:
On the 128-line limit: as @michaelblyons noted, it's already enforced — branch points are pruned at the start of every `parse_line` call and re-checked in `handle_fail`. There are tests for both the boundary (exactly 128 lines: still valid) and the expiry (129 lines: discarded). I've now also added structured warnings (following the same pattern as the syntax loading warnings) to `ParseLineOutput` so callers can see when a branch point expires. The existing tests now assert on the warning content too.

stefanobaghino:
I tweaked the description to mention this closes #271 so that when it's merged the issue is closed automatically. Let me know if you have anything against it.

stefanobaghino:
I have a patch ready for #222 ([diff](https://github.com/stefanobaghino/syntect/compare/branch-point...native-embed-escape?expand=1)) once this is merged.

I also went a bit further and bumped the packages on top of that to see what the next steps are and I think we might be in a position where we can cleanly move everything syntax tests from `st2` to `master` without problems. 🙂

If there's anything I can do to facilitate the review I'd be happy to help.

jrappen:
Looked good last time I checked.

jrappen:
@stefanobaghino Just as a heads up, if you look at the discussion in #271 you can see towards the very top that I had submitted #279 to temporarily set the submodules for Packages to st3 branch ... in case you needed a quick reference when you test your patches.

PR #615, opened 2026-03-28
Implement native embed/escape support

stefanobaghino:
## Summary

Fixes #222.

Replace the `embed`→`push`+`with_prototype` rewriting with a native `MatchOperation::Embed` variant where the escape regex takes strict precedence over all other patterns, fixing cases where embedded context patterns could consume escape characters first.

### Key changes

- Add `EscapeInfo` struct and `MatchOperation::Embed` variant to `SyntaxDefinition`
- Add `escape_stack` to `ParseState` with `EscapeEntry` tracking
- Modify `find_best_match` to check escape patterns first and truncate the search region for normal patterns
- Handle escape execution by popping stack to embed depth
- Snapshot `escape_stack` in `BranchPoint` for backtracking support
- Update YAML loader, linker, and public API snapshot
- Regenerate binary assets

### Tests

Each new test was mutation-verified against a specific production code path:
- `nested_embed_outer_escape_wins`: verifies escape precedence in nested embeds
- `embed_escape_with_backref_at_parse_time`: verifies backref resolution in escape patterns
- `embed_escape_cross_line`: verifies escape stack persistence across line boundaries
- `embed_inside_branch_then_fail_restores_escape_stack`: verifies escape stack restoration on branch failure

Issue #606, opened 2025-12-18
Replacement for bincode as a dep?


nazmulidris:
Given https://git.sr.ht/~stygianentity/bincode:
> "...bincode development has officially ceased and will not resume. Version 1.3.3 is considered a complete version of bincode that is not in need of any updates..."

Will syntect use another crate to replace bincode? 
https://github.com/trishume/syntect/blob/f329b32051d199741bcf30892e50cc16f36bf4cf/src/dumps.rs#L19-L24

CosmicHorrorDev:
removing dependencies from the public API is an overall goal #512. there are plenty of alternatives to consider in terms of internal use

cisco211:
For me its not just bincode:

```
cargo audit
    Fetching advisory database from `https://github.com/RustSec/advisory-db.git`
      Loaded 900 security advisories (from C:\Users\Cisco211\.cargo\advisory-db)
    Updating crates.io index
    Scanning Cargo.lock for vulnerabilities (495 crate dependencies)
Crate:     bincode
Version:   1.3.3
Warning:   unmaintained
Title:     Bincode is unmaintained
Date:      2025-12-16
ID:        RUSTSEC-2025-0141
URL:       https://rustsec.org/advisories/RUSTSEC-2025-0141
Dependency tree:
bincode 1.3.3
└── syntect 5.3.0
    └── egui_extras 0.33.3
        ├── keyring-gui 0.1.0
        │   └── keyring-desktop 0.1.0
        ├── egui_commonmark_backend 0.22.0
        │   └── egui_commonmark 0.22.0
        │       └── keyring-gui 0.1.0
        └── egui_commonmark 0.22.0

Crate:     yaml-rust
Version:   0.4.5
Warning:   unmaintained
Title:     yaml-rust is unmaintained.
Date:      2024-03-20
ID:        RUSTSEC-2024-0320
URL:       https://rustsec.org/advisories/RUSTSEC-2024-0320
Dependency tree:
yaml-rust 0.4.5
└── syntect 5.3.0
    └── egui_extras 0.33.3
        ├── keyring-gui 0.1.0
        │   └── keyring-desktop 0.1.0
        ├── egui_commonmark_backend 0.22.0
        │   └── egui_commonmark 0.22.0
        │       └── keyring-gui 0.1.0
        └── egui_commonmark 0.22.0

warning: 2 allowed warnings found
```

CosmicHorrorDev:
YAML is mentioned in

- #537

[...]

Issue #588, opened 2025-07-24
Use `dep:` prefix for feature dependencies


CosmicHorrorDev:
**NOTE: breaking change**

Without the `dep:` prefix optional dependencies create implicit feature flags which leaks internal details along with cluttering up the crate's feature list (particularly when you have fun combinations like `regex-fancy` vs `fancy-regex`). Considering none of the implicit feature flags are supposed to be used by end users it would be nice to switch to using the `dep:` prefix in the next semver breaking release

Issue #574, opened 2025-02-23
Merging two syntax sets panics


jyn514:
I wrote this code:
```rust
let mut merged_syntaxes = SYNTAX_SET.clone().into_builder();
for extra in extra_syntaxes.clone().into_builder().syntaxes() {
    merged_syntaxes.add(extra.clone());
}
```
I expect it to run without panicking; the whole point of `into_builder()` is that it removes the SyntaxSet-specific relative indexes. instead, i get this panic:
```
thread '<unnamed>' panicked at /home/jyn/.local/lib/cargo/registry/src/index.crates.io-1949cf8c6b5b557f/syntect-5.2.0/src/parsing/syntax_set.rs:683:62:
index out of bounds: the len is 78 but the index is 129
stack backtrace:
   0: rust_begin_unwind
             at /rustc/f7cc13af822fe68c64fec0b05aa9dd1412451f7c/library/std/src/panicking.rs:695:5
...
   5: <alloc::vec::Vec<T,A> as core::ops::index::Index<I>>::index
             at /home/jyn/.local/lib/rustup/toolchains/nightly-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/alloc/src/vec/mod.rs:3359:9
   6: syntect::parsing::syntax_set::SyntaxSetBuilder::recursively_mark_no_prototype
             at /home/jyn/.local/lib/cargo/registry/src/index.crates.io-1949cf8c6b5b557f/syntect-5.2.0/src/parsing/syntax_set.rs:683:62
   7: syntect::parsing::syntax_set::SyntaxSetBuilder::recursively_mark_no_prototype
             at /home/jyn/.local/lib/cargo/registry/src/index.crates.io-1949cf8c6b5b557f/syntect-5.2.0/src/parsing/syntax_set.rs:717:29
   8: syntect::parsing::syntax_set::SyntaxSetBuilder::build
             at /home/jyn/.local/lib/cargo/registry/src/index.crates.io-1949cf8c6b5b557f/syntect-5.2.0/src/parsing/syntax_set.rs:598:17
   9: config::highlighting::resolve_syntax_and_theme::{{closure}}
             at /home/jyn/src/zola/components/config/src/highlighting.rs:53:21
...
```
that points at https://github.com/trishume/syntect/blob/v5.2.0/src/parsing/syntax_set.rs#L683. i suspect that `context_id` is not being renumbered correctly when calling `into_builder()`.

PR #616, opened 2026-03-31
Bump packages

stefanobaghino:
Reverts parts of the forced temporary changes in #279 which were necessary due to #271.

Supersedes #535.

The following PRs closed the gap required for the packages branch to point to `master` again:
* #610
* #612
* #614
* #615

The changes introduce breaking changes to the public API which require a new major in order to be released. In order to get it out, my recommendation would be to batch all breaking changes before getting 6.0.0 out. To this end, [the existing milestone can be used](https://github.com/trishume/syntect/milestone/1). I had a look around the repo and my understanding is the following.

### Currently in the 6.0.0 milestone

* #512: Public dump functions return `bincode::error::Result<()>`. Error types `LoadingError::WalkDir` wraps `walkdir::Error`, `SettingsError::Plist` wraps `plist::error::Error`. Removing or wrapping these dependency types in syntect-owned types changes public function signatures and error enum variants.
* #394: `SyntaxSet::load_defaults_nonewlines()` is a public function referenced across `easy.rs`, `html.rs`, and test code. Removing this mode is a breaking deletion.
* #297: Regex` and `Region` in `src/parsing/regex.rs` are public structs. `Region` wraps `regex_impl::Region` which uses raw pointers with the onig backend, making `ParseState` `!Send + !Sync`. Making `ParseState` Send+Sync requires restructuring these public types, changing auto-trait bounds.
* #540 (PR closing #482): `SyntaxReference` has `pub variables: HashMap<String, String>` and `SyntaxDefinition` has the same. Moving this field into compressed/lazy storage removes or changes a public field on both structs.

### Existing tickets I suggest to add to the 6.0.0 milestone

* #588: Non-optional deps (`walkdir`, `serde`, `once_cell`) create implicit feature flags. Downstream crates activating `syntect/walkdir` etc. as features will break when those implicit flags are removed. The fork has already partially applied this via commit `69c0544`.
* #468: `css_for_theme_with_class_style` returns `Result<String, Error>` but can never error. Changing to `String` is a breaking signature change.
* #36: `ScopeSelector` has all-public fields (`pub path: ScopeStack`, `pub excludes: Vec<ScopeStack>`) and `ScopeSelectors` has `pub selectors: Vec<ScopeSelector>`. The `&` (AND) operator requires that a scope stack simultaneously matches multiple independent selectors, which cannot be expressed with the current layout. Any fix requires adding fields to public structs or making them `#[non_exhaustive]`. This was the only issue that required investigation to determine it was breaking.
* #564: Changes `HighlightLines::new(syntax, theme)` to `HighlightLines::new(syntax, theme, options)` by adding a required `HighlightOptions` struct parameter.

### Can be deferred past 6.0.0

These are currently breaking but become non-breaking once #512 ships, because #512 wraps leaked dependency types behind syntect-owned types. They can be addressed in a 6.x minor/patch release instead of blocking the 6.0.0 release.

- #606: Currently breaking because dump functions return `bincode::error::Result` and the dump file binary format would change. Once #512 wraps `bincode::Result` in syntect-owned error types, swapping the serialization backend is an internal change that can ship as 6.1.0 or later.

- #609 -- Dependency updates (yaml-rust2 0.10 to 0.11)** Currently breaking because `ParseSyntaxError::InvalidYaml` wraps `yaml_rust2::scanner::ScanError`, a leaked dependency type. Once #512 wraps this in a syntect-owned error, bumping yaml-rust2 is an internal change. Also bumps fancy-regex 0.16 to 0.17 and dev deps (criterion, pretty_assertions). The `MatchPower::partial_cmp` signature change (`&MatchPower` to `&Self`) is cosmetic.

- #576 (PR closing #575): Public dependency on `once_cell` / Drop once_cell** `SCOPE_REPO` is declared as `pub static SCOPE_REPO: Lazy<Mutex<ScopeRepository>>` using `once_cell::sync::Lazy`. PR #576 replaces it with `OnceLock`. Deferrable **only if** #512 removes `SCOPE_REPO` from the public API (already deprecated since 5.3.0) or wraps its type. If `SCOPE_REPO` remains public with a `Lazy<...>` type after #512, this is still breaking and must be in 6.0.0 instead.

### 1. Structured warnings API for syntax loading

Create a ticket to replace `eprintln!` warnings during syntax loading and linking with a programmatic `warnings()` API on `SyntaxSetBuilder` and `SyntaxSet`.

Currently, `SyntaxSetBuilder::add_from_folder` and `SyntaxSetBuilder::build` print warnings to stderr when they encounter skipped files, version mismatches between `extends` parents, unresolved `extends` references, or regex re-resolution failures. This is problematic for library consumers: warnings are invisible in non-terminal contexts, cannot be filtered or forwarded to a logging framework, and pollute stderr in applications that use it for their own output.

The change should:

- Add a `warnings: Vec<String>` field to both `SyntaxSetBuilder` and `SyntaxSet` (skipped during serialization/deserialization).
- Replace all `eprintln!("Warning: ...")` calls in `add_from_folder` and `resolve_extends` with `self.warnings.push(...)`.
- Expose `pub fn warnings(&self) -> &[String]` on both types.
- Transfer collected warnings from `SyntaxSetBuilder` into the built `SyntaxSet`.
- Update the `syntest` example to demonstrate consuming warnings.

This is a **non-breaking additive change** (new public methods, new skipped field) and can ship in 6.0.0 or any later minor release. However, bundling it with 6.0.0 is preferable since the `resolve_extends` internals are already being touched for #540.

### 2. Mutation testing

Create a ticket to run mutation testing on the repository to improve confidence in the test suite, especially around the areas being changed for 6.0.0. The ticket should cover:

- Run `cargo-mutants` (or equivalent) on the modules most affected by 6.0.0 changes: `src/parsing/parser.rs`, `src/parsing/regex.rs`, `src/parsing/syntax_set.rs`, `src/highlighting/selector.rs`, `src/html.rs`, `src/dumps.rs`, `src/easy.rs`.
- Prioritize killing surviving mutants in code paths related to: scope selector matching (#36), `into_builder()` round-tripping (#574), Send+Sync safety (#297), and error type wrapping (#512).
- Establish a baseline mutation score before 6.0.0 work begins and track it as changes land.
- Note: the fork already has some mutation testing experience (commits `a8953e3`, `252b4a4` killing surviving mutations in branch/fail code). The same approach should be extended to the rest of the codebase.

### 3. Performance regression testing

Create a ticket to establish performance benchmarks before and after the 6.0.0 changes. The ticket should cover:

- **Parsing performance:** Benchmark `ParseState::parse_line` on representative files (large HTML, JavaScript, Markdown) to catch regressions from #297 (Send+Sync changes to `Region`), #540 (compressed variables requiring decompression), and PR #615 (embed/escape support).
- **Serialization performance:** Benchmark `dump_to_writer`/`from_reader` round-trips to catch regressions from #512 (new error wrapping) and any future bincode replacement (#606).
- **Highlighting performance:** Benchmark `HighlightLines::highlight_line` end-to-end to catch regressions from PR #564 (`HighlightOptions` overhead) and #36 (advanced scope selector matching).
- **Memory/size:** Measure `SyntaxSet` serialized size before and after #540/#482 (variable compression) and #481 (regex comment stripping, if included).

Existing benchmarks in `benches/` should be extended or a new criterion-based benchmark suite should be created. Benchmarks should be run as part of CI on the 6.0.0 branch before release.

### 4. Migration guide for 5.x to 6.0.0

Create a ticket to write a migration document covering all breaking changes in the 6.0.0 release. The document should:

- List every removed, renamed, or changed public API item with before/after code snippets.
- Cover the `nonewlines` removal (#394) with the recommended replacement pattern.
- Explain the new error types that replace leaked dependency types (#512).
- Document the `SyntaxReference.variables` change (#540/#482) and how to access variable data in the new API.
- Document the `dep:` feature flag changes (#588) and which `Cargo.toml` feature entries need updating.
- Document the `HighlightLines::new()` signature change (PR #564) and how to construct `HighlightOptions`.
- If PR #615 is included, document the new `MatchOperation::Embed` variant and `Pattern::IncludeWithPrototype`.
- If #36 is included, document the new `ScopeSelector`/`ScopeSelectors` struct layout.
- If #575 is included in 6.0.0, document the `SCOPE_REPO` removal or type change.

This guide should be published alongside the 6.0.0 release (e.g., in `CHANGELOG.md` or a dedicated `MIGRATION.md`).

### DIFF
--- a/src/parsing/syntax_set.rs
+++ b/src/parsing/syntax_set.rs
@@ -1343,8 +1343,8 @@ mod tests {
         );
         assert_eq!(
             &ps.find_syntax_for_file("testdata/test_first_line.test")
-                .unwrap()
-                .unwrap()
+                .expect("Error finding syntax for file")
+                .expect("No syntax found for file")
                 .name,
             "Ruby"
         );


keith-hall:
Thanks for the great ideas - I added the suggested tickets to the 6.0.0 milestone, though I do worry that the scope becomes very big. Which could delay getting all the changes into the hands of the library consumers ... Having a few major releases with clear upgrade documentation may be less daunting than one massive one... But I'm happy to hear counter arguments. 🙂

Feel free to create the new tickets, it will be better to discuss each one individually rather than try to coordinate all discussions here.

jrappen:
Maybe you could update the opening post / the commit message of the commit to reference this reverting parts of the forced temporary changes in #279 which were necessary due to #271, i.e. `gitmodules` lock on `st3` branch.

jrappen:
Also, maybe you could check if this supersedes #535 and if so add it to the references list as `- supersedes #535` so GitHub recognizes it.

stefanobaghino:
> Thanks for the great ideas - I added the suggested tickets to the 6.0.0 milestone, though I do worry that the scope becomes very big. Which could delay getting all the changes into the hands of the library consumers ... Having a few major releases with clear upgrade documentation may be less daunting than one massive one... But I'm happy to hear counter arguments. 🙂

That's a valid concern, thanks for raising it. My take: there hasn't been a major release in several years, and publishing the latest changes already requires one, so there's good reason to make a single transition.

The biggest long-planned change (removing external dependencies from the public API surface) is also the one with the largest impact on consumers. I think it's important for the project's long-term sustainability and counter-productive to delay further. Since it already forces a major version bump, the other breaking changes are relatively minor in comparison, which is why I'd take advantage of the situation and tackle them all in one go.

From what I can tell, each individual change is small and shouldn't add much latency to the release (definitely within what I'd expect for this project's cadence).

stefanobaghino:
> Maybe you could update the opening post / the commit message of the commit to reference this reverting parts of the forced temporary changes in #279 which were necessary due to #271, i.e. `gitmodules` lock on `st3` branch.

Done.

> Also, maybe you could check if this supersedes #535 and if so add it to the references list as `- supersedes #535` so GitHub recognizes it.

Done as well, although I don't believe GitHub recognizes "supersedes" automatically ([docs](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests)). I still think it's a good idea anyway. I'll post a message to #535 to raise this as well. Thanks for mentioning that.

stefanobaghino:
> Feel free to create the new tickets, it will be better to discuss each one individually rather than try to coordinate all discussions here.

Done:
* #619 
* #620 
* #621 
* #622

stefanobaghino:
Anything else that needs to be further reviewed or that I can help with? 🙂

PR #624, opened 2026-04-04
Add FIND_NOT_EMPTY support for None-operation match patterns

keith-hall:
applies the hint from https://github.com/trishume/syntect/issues/59#issuecomment-472718568 to fix the LaTeX syntax test failures. For plain match patterns which don't affect the context stack in any way, empty matches should be excluded. Oniguruma's `FIND_NOT_EMPTY` flag is apparently a compile time flag not a runtime one, so Copilot came up with the approach of a new `OnceCell` containing the compiled regex with this flag and adding a new `allow_empty` boolean to the `Regex` `search` method. And also applied some changes for fancy-regex, which doesn't have such a flag.

If we decide to rework this, I'm fine with that, just wanted to raise this PR as a possible solution. The Makefile syntax test failures will still need to be looked into, however.

Likely I will need to add a similar flag to fancy-regex.

### DIFF
--- a/src/parsing/parser.rs
+++ b/src/parsing/parser.rs
@@ -566,7 +566,7 @@ impl ParseState {
             let mut esc_regions = Region::new();
             if entry
                 .regex
-                .search(line, start, line.len(), Some(&mut esc_regions))
+                .search(line, start, line.len(), Some(&mut esc_regions), true)
             {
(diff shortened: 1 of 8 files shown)


keith-hall:
I have a PR open to add a similar flag to fancy-regex: https://github.com/fancy-regex/fancy-regex/pull/240

stefanobaghino:
Should be superseded by #636.

Issue #617, opened 2026-03-31
Easy branch point aware parsing/highlighting api


keith-hall:
Now that we have branch point support, I think it makes sense to expose a new API in the easy module which would abstract away the complexities of dealing with the speculative states and allow just being able to send lines of text to the parser/highlighter (in my mind it would handle the parse state itself, but I guess for caching purposes the caller may still want to know the state at each line), and perform a callback or something when a line has completed parsing (either because a branch succeeded or the backtracking state was pruned due to the line length limits of branching) and thus the scopes/operations are final.

Issue #618, opened 2026-04-01
Modern syntax definitions support


kobrir:
After #323 was closed I decided to try syntect with the modern syntax definitions from the sublimehq/Packages but it seems that there are still some issues with syntect.
For the test I downloaded sublimehq/Packages repository and ran syncat with this command line:
```
syncat --no-default-syntaxes --extra-syntaxes ./Packages-master test.js
```
And got this error message:
```
thread 'main' panicked at examples\syncat.rs:96:55:
called `Result::unwrap()` on an `Err` value: ParseSyntax(InvalidScope(TooManyAtoms), "./Packages-master\\ActionScript\\ActionScript.sublime-syntax")
```
I deleted the `ActionScript` folder and tried again, which produced this error:
```
thread 'main' panicked at examples\syncat.rs:96:55:
called `Result::unwrap()` on an `Err` value: ParseSyntax(MissingMandatoryKey("contexts"), "./Packages-master\\Batch File\\Batch File (Compound).sublime-syntax")
```
I continued deleting folders and eventually syncat managed to run without panicking but produced highlighting with some artifacts:

<img width="250" height="345" alt="Image" src="https://github.com/user-attachments/assets/206ef630-c171-4023-ba87-66173a34324c" />

Here is the test.js content I tried it on:
```javascript
function fn1()
{
  function fn2()
  {
    function fn3()
    {
    }
    function fn4()
    {
    }
  }
  return fn2;
}
```

jrappen:
You probably have to wait for at least 616. Stefano was gonna do a few follow ups to get everything done and working again one-by-one.

michaelblyons:
#616 is merged. Maybe try again?

keith-hall:
`syncat` might still not work properly because it doesn't respect branch points - as far as I know it would still need to rewind scope stack operations from failed branches and apply the new operations. #617 would help with this, but right now it's only implemented directly in the `syntest` example, which doesn't do any highlighting...

kobrir:
Now there is no panicking but as I understand it, syntect is just skipping the same files that I previously deleted manually. Here is the list of warnings that I have got:
```
Warning: skipping "./Packages-master\\ActionScript\\ActionScript.sublime-syntax": ./Packages-master\ActionScript\ActionScript.sublime-syntax: Invalid scope: Too many atoms. Max 2^16-2 atoms allowed.
[...]

PR #627, opened 2026-04-08
Streaming `HighlightedWriter` and layered renderer traits

stefanobaghino:
> **Stacked on #692.** Review just this PR's changes: https://github.com/stefanobaghino/syntect/compare/631-zero-width-escape-loop-guard...trait-based-classed-highlighter

## Summary

Introduces a streaming, branch-point-aware highlighter and a layered renderer trait design that together replace several legacy line-by-line APIs and address `SCOPE_REPO` removal in one coherent rework.

The change resolves four open concerns at once:

| Issue | What it asked for | How this PR addresses it |
|---|---|---|
| **#626** | A more ergonomic, less coupled replacement for `SCOPE_REPO` | The new `ScopeMarkup` / `StyledOutput` traits hand pre-resolved atom strings (`&[&str]`) to implementors. `Scope::with_atom_strs()` is the closure-based escape hatch for the rare case that needs raw access. `SCOPE_REPO` is removed from the public API. |
| **#617** | An "easy" branch-point-aware parsing/highlighting API that hides speculative state from callers | `HighlightedWriter` implements `std::io::Write` and buffers rendered output during speculative parsing, replaying corrected ops once speculation resolves. Callers just write bytes and get correctly-highlighted output back — no awareness of branch points required. |
| **#614** | Branch-point support; flagged that line-by-line APIs become incorrect under cross-line failure | `line_tokens_to_classed_spans` is deprecated for exactly this reason: when the parser retroactively replays a span of previously-emitted ops, the rendered output for those lines has already been written and cannot be retracted. `HighlightedWriter` is the streaming-shaped replacement. |
| **#611** | Line highlighting support in `ClassedHTMLGenerator` | Can now be implemented as a ~15-line `ScopeMarkup` impl that delegates to `ClassedHTMLScopeRenderer` and emits a `<span class="hl">` wrapper from `begin_line` / `end_line`. See the sketch below. |

### `syntect::io::HighlightedWriter`

A streaming syntax highlighter that implements `std::io::Write`. Bytes go in, rendered output flows to an inner sink. Branch-point speculation is handled internally — callers never see speculative state. End-of-input cleanup runs implicitly via `Drop` on a best-effort basis; call `into_inner` explicitly to recover the inner sink and propagate errors.

Constructed via a fluent type-state builder:

- `HighlightedWriter::from_themed(syntax, ss, theme, output)` — pair a `Theme` with any `StyledOutput` (`AnsiStyledOutput` for terminals, `HtmlStyledOutput` for inline-styled HTML, your own format for LaTeX/etc.)
- `HighlightedWriter::from_markup(syntax, ss, markup)` — pair a `ScopeMarkup` (CSS-classed HTML or any other stateless markup renderer)
- `HighlightedWriter::from_renderer(syntax, ss, renderer)` — low-level escape hatch for raw `ScopeRenderer`

Each entry point returns a `HighlightedWriterBuilder` pre-populated with a `Vec<u8>` sink and no resume state. Configure further with `.with_output(...)` / `.with_state(...)` and finish with `.build()`.

### `syntect::rendering` — three traits, three audiences

```text
syntect::rendering::ScopeMarkup     ←  stateless renderers (CSS classes, XML, …)
syntect::rendering::StyledOutput    ←  theme-aware emitters (ANSI, inline HTML, LaTeX, …)
                                       wrapped by ThemedRenderer<O> which manages
                                       Highlighter / style stack / merge state
syntect::rendering::ScopeRenderer   ←  low-level engine trait, for advanced cases
                                       that need raw Scope / &[Scope] access
```

- **`ScopeMarkup`** — slim trait for renderers that map scope structure 1:1 to output structure. Receives only pre-resolved atom strings (no raw `Scope`, no `&[Scope]`, no `bool` return). The empty-scope optimisation is unconditional. This is the trait the wastebin use case wants.

- **`StyledOutput`** + **`ThemedRenderer<O>`** — small trait for theme-aware renderers. Implementors only describe `begin_style` / `end_style` / `write_text`. The `ThemedRenderer<O>` adapter owns the `Highlighter`, mirrors the scope stack, and merges adjacent same-styled tokens on the implementor's behalf. Built-in `AnsiStyledOutput` (terminals) and `HtmlStyledOutput` (inline-styled HTML) live in `syntect::rendering` and `syntect::html` respectively. The `latex-demo` example shows a ~25-line custom format.

- **`ScopeRenderer`** — the low-level engine trait the writer actually consumes. Stays public as the escape hatch for advanced cases that need raw `Scope` / `&[Scope]` access or want fine-grained control over the empty-scope truncation via the `bool` return from `begin_scope`. Most users should reach for `ScopeMarkup` or `StyledOutput` instead.

### Backward-compatible wrappers

- **`ClassedHTMLGenerator`** — public surface unchanged. Internally now wraps `HighlightedWriter` driven by a `MarkupAdapter<ClassedHTMLScopeRenderer>`. The previously-existing line-oriented `parse_html_for_line_which_includes_newline` API still compiles; it just funnels bytes through the streaming writer.
- **`highlighted_html_for_string`** / **`highlighted_html_for_file`** — public signatures unchanged. Internally use `HighlightedWriter::from_themed` with `HtmlStyledOutput`.

### Removed / deprecated

- **Removed**: `SCOPE_REPO` (already deprecated since 5.3.0).
- **Deprecated since 6.0.0** (`#[deprecated]`, still compiles):
  - `HighlightLines` / `HighlightFile` — line-oriented; cannot interleave branch-point speculation with output. Migrate to `HighlightedWriter`.
  - `line_tokens_to_classed_spans` — same hazard. Migrate to `ClassedHTMLGenerator` or `HighlightedWriter::from_markup`.

## #611 — line highlighting, with the new traits

The Yew-docs use case from #611 wants to wrap selected lines (e.g. `[2, 5, 6]`) in a `<span class="hl">…</span>` so they can be styled differently. The proposed API was a new `ClassedHTMLGenerator::new_with_class_style_and_highlighted_lines(syntax, ss, style, highlighted)` constructor.

With the layered traits, this becomes a ~15-line `ScopeMarkup` impl that wraps `ClassedHTMLScopeRenderer` and uses the `begin_line` / `end_line` hooks:

```rust
use std::collections::BTreeSet;
use syntect::html::{ClassStyle, ClassedHTMLScopeRenderer};
use syntect::rendering::ScopeMarkup;

pub struct LineHighlightedClassedHTMLScopeRenderer {
    inner: ClassedHTMLScopeRenderer,
    highlighted_lines: BTreeSet<usize>,
    inside_highlight: bool,
}

impl LineHighlightedClassedHTMLScopeRenderer {
    pub fn new(style: ClassStyle, highlighted_lines: impl IntoIterator<Item = usize>) -> Self {
        Self {
            inner: ClassedHTMLScopeRenderer::new(style),
            highlighted_lines: highlighted_lines.into_iter().collect(),
            inside_highlight: false,
        }
    }
}

impl ScopeMarkup for LineHighlightedClassedHTMLScopeRenderer {
    fn begin_line(&mut self, line_index: usize, output: &mut String) {
        // Issue #611 uses 1-based line numbers; the engine is 0-based.
        if self.highlighted_lines.contains(&(line_index + 1)) {
            output.push_str(r#"<span class="hl">"#);
            self.inside_highlight = true;
        }
    }

    fn end_line(&mut self, _line_index: usize, output: &mut String) {
        if self.inside_highlight {
            output.push_str("</span>");
            self.inside_highlight = false;
        }
    }

    fn begin_scope(&mut self, atom_strs: &[&str], output: &mut String) {
        self.inner.begin_scope(atom_strs, output);
    }
    fn end_scope(&mut self, output: &mut String) {
        self.inner.end_scope(output);
    }
    fn write_text(&mut self, text: &str, output: &mut String) {
        self.inner.write_text(text, output);
    }
}
```

Usage:

```rust
let renderer = LineHighlightedClassedHTMLScopeRenderer::new(ClassStyle::Spaced, [2, 5, 6]);
let mut w = HighlightedWriter::from_markup(syntax, &ss, renderer).build();
w.write_all(code.as_bytes())?;
let html = String::from_utf8(w.into_inner()?)?;
```

Or, kept inside `ClassedHTMLGenerator` as a convenience constructor matching the issue's proposed shape:

```rust
impl<'a> ClassedHTMLGenerator<'a> {
    pub fn new_with_class_style_and_highlighted_lines(
        syntax: &'a SyntaxReference,
        syntax_set: &'a SyntaxSet,
        style: ClassStyle,
        highlighted_lines: impl IntoIterator<Item = usize>,
    ) -> ClassedHTMLGenerator<'a> {
        ClassedHTMLGenerator {
            inner: HighlightedWriter::from_markup(
                syntax,
                syntax_set,
                LineHighlightedClassedHTMLScopeRenderer::new(style, highlighted_lines),
            )
            .build(),
        }
    }
}
```

This is what the layered design unlocks: line highlighting is no longer a special case the `ClassedHTMLGenerator` constructor has to know about — it's a `ScopeMarkup` decorator that anyone can write. The same shape works for adding line numbers, link decoration, syntax error markers, or any other line-level markup.

I'd prefer to land #611's actual implementation as a follow-up so this PR stays focused on the structural rework, but the architecture is in place for it now.

### CSS-classed HTML (`ClassedHTMLGenerator`)

No change. Your existing code keeps working:

```rust
let mut gen = ClassedHTMLGenerator::new_with_class_style(syntax, &ss, ClassStyle::Spaced);
for line in LinesWithEndings::from(code) {
    gen.parse_html_for_line_which_includes_newline(line)?;
}
let html = gen.finalize();
```

### Inline-styled HTML / file rendering

`highlighted_html_for_string` / `highlighted_html_for_file` keep their public signatures.

### Custom renderer (the wastebin use case from #626)

Was: copy-paste `line_tokens_to_classed_spans` and use `SCOPE_REPO`. Now:

```rust
use syntect::rendering::ScopeMarkup;

struct MyRenderer { /* … */ }

impl ScopeMarkup for MyRenderer {
    fn begin_scope(&mut self, atom_strs: &[&str], output: &mut String) {
        // atom_strs is e.g. ["markup", "underline", "link", "markdown"]
        // — no SCOPE_REPO access needed
    }
    fn end_scope(&mut self, output: &mut String) { /* … */ }
    fn write_text(&mut self, text: &str, output: &mut String) { /* … */ }
}

let mut w = HighlightedWriter::from_markup(syntax, &ss, MyRenderer { /* … */ }).build();
w.write_all(code.as_bytes())?;
let html = String::from_utf8(w.into_inner()?)?;
```

### Theme-aware terminal output (was: `HighlightLines`)

Was:
```rust
let mut h = HighlightLines::new(syntax, &theme);
for line in LinesWithEndings::from(code) {
    let regions = h.highlight_line(line, &ss)?;
    print!("{}", as_24_bit_terminal_escaped(&regions[..], true));
}
```

Now:
```rust
let mut w = HighlightedWriter::from_themed(syntax, &ss, &theme, AnsiStyledOutput::new(false))
    .with_output(io::stdout().lock())
    .build();
io::copy(&mut source, &mut w)?;
```

Or with the convenience defaults (`Vec<u8>` sink, accumulate-then-print):
```rust
let mut w = HighlightedWriter::from_themed(syntax, &ss, &theme, AnsiStyledOutput::new(false)).build();
w.write_all(code.as_bytes())?;
let bytes = w.into_inner()?;
```

### Custom themed format (LaTeX, etc.)

Was: build a 80+ line `ScopeRenderer` impl from scratch managing `Highlighter`, `style_stack`, `last_written_style`. Now: implement `StyledOutput` (~25 lines) and let `ThemedRenderer<O>` manage the state.

See `examples/latex-demo.rs` for a complete LaTeX `\textcolor` emitter.

## References

- #626 — Replace `SCOPE_REPO` with a more ergonomic and less coupled API
- #617 — Easy branch-point-aware parsing/highlighting API
- #614 — Branch point parser support (and the cross-line correctness hazard it surfaces)
- #611 — Add line highlighting support to `ClassedHTMLGenerator`
- #575 — Original `SCOPE_REPO` deprecation discussion
- #512 — Remove dependencies from public API (parent effort)


matze:
Looks great!

stefanobaghino:
~Since opening the draft, the `ScopeRenderer` trait has been generalized: it's now generic over `W: io::Write` (defaulting to `Vec<u8>`), so implementors can target any writable sink (`File`, `BufWriter`, `TcpStream`, etc.) without needing an adapter. The built-in rendering loop still buffers into `Vec<u8>` and converts to `String` at the boundary, so existing behavior is unchanged.~ (EDIT: this was undone in a follow-up commit as ultimately not really a productive change)

@matze — Thanks for the feedback. I'll move this from draft to ready for review.

src/html/mod.rs
```diff
@@ -3,35 +3,36 @@ use crate::easy::{HighlightFile, HighlightLines};
...
+/// the original `ClassedHTMLGenerator`.
 ///
 /// There is a [`finalize()`] method that must be called in the end in order
-/// to close all open `<span>` tags.
-///
-/// Note that because CSS classes have slightly different matching semantics
-/// than Textmate themes, this may produce somewhat less accurate
-/// highlighting than the other highlighting functions which directly use
-/// inline colors as opposed to classes and a stylesheet.
+/// to close all open tags.
```
keith-hall: I'll give my usual question - is this comment deemed no longer relevant? I think it still applies 🙂

stefanobaghino: Good catch — restored the note about CSS class matching semantics. It's inherent to the class-based approach, not to the old implementation, so it still applies.

stefanobaghino: Addressed in https://github.com/trishume/syntect/pull/627/commits/98a02a39c99f2a2c6abcecc8ed830f10549a5a52.

stefanobaghino:
I'm having a better look and I believe there are a few design quirks that need ironing out before opening up for review. Specifically, the whole mechanism is quite general now, but the naming is a bit off from this point of view. Preserving the existing names to facilitate the migration makes sense, but not to the point of being misleading. I'll continue working on this tomorrow.

stefanobaghino:
Follow-up on the naming concerns raised [above](https://github.com/trishume/syntect/pull/627#issuecomment-4209304013). Here's a summary of the naming changes since then:

### Renamed

- **`ClassedHTMLGenerator<'a, R>`** → **`DocumentGenerator<'a, R: ScopeRenderer>`** (`syntect::generator`) — the generic parsing + rendering driver is now its own type in a dedicated module, with a format-agnostic name. `ClassedHTMLGenerator<'a>` remains as a newtype wrapper around `DocumentGenerator<'a, HTMLScopeRenderer>` with its public API unchanged.
- **`HtmlScopeRenderer`** → **`HTMLScopeRenderer`** — consistent with `ClassedHTMLGenerator` naming convention.
- **`render_line_to_classed_spans()`** → **`render_line()`** — the core rendering loop is format-agnostic; the old name implied HTML classes and spans. Internal variable names (`span_delta` → `scope_delta`, etc.) and comments were updated to match.

### Motivation

The generic machinery (`ScopeRenderer`, `render_line`, `DocumentGenerator`) is format-agnostic — a renderer could produce LaTeX, ANSI, plain text, or any other format. The previous names (`ClassedHTMLGenerator`, `render_line_to_classed_spans`) leaked HTML semantics into format-neutral code. The rename draws a clear boundary: HTML-specific types live in `syntect::html`, generic types live in `syntect::renderer` and `syntect::generator`.

`ClassedHTMLGenerator` keeps its name and API for backward compatibility — it's just a thin wrapper now.

I've also reverted the `ScopeRenderer` output generalization from `W: io::Write` back to `&mut String`, since `DocumentGenerator::finalize()` returns a `String` anyway and rendering to `String` directly is simpler for implementors.

stefanobaghino:
> I'm having a better look and I believe there are a few design quirks that need ironing out before opening up for review. Specifically, the whole mechanism is quite general now, but the naming is a bit off from this point of view. Preserving the existing names to facilitate the migration makes sense, but not to the point of being misleading. I'll continue working on this tomorrow.

It ended up being faster than I imagined. 🙂

stefanobaghino:
I'm relatively happy with the current branch state and believe this is now definitely ready for review. Thanks for your patience.

matze:
> I've also reverted the ScopeRenderer output generalization from W: io::Write back to &mut String, since DocumentGenerator::finalize() returns a String anyway and rendering to String directly is simpler for implementors.

For performance conscious applications avoiding intermediate allocations could be a boon but at least for wastebin this makes no difference because at the end an owned response has to be produced anyway.

stefanobaghino:
> > I've also reverted the ScopeRenderer output generalization from W: io::Write back to &mut String, since DocumentGenerator::finalize() returns a String anyway and rendering to String directly is simpler for implementors.
> 
> 
> 
> For performance conscious applications avoiding intermediate allocations could be a boon but at least for wastebin this makes no difference because at the end an owned response has to be produced anyway.

It's a valid point, the one that initially drove me to generalize over the writable output. I reverted with `ClassedHTMLGenerator::finalize` in mind, but let me see if I can generalize that in the new `DocumentGenerator` instead. 🤔

stefanobaghino:
I'm having a better look and I'm under the impression that there might be contact points between this PR and #617. I'm trying to evaluate whether a single abstraction can take care of these use cases without adding complexity. This means that we would have buffering bounded by the maximum amount of lines in a cancellable scope (128 lines if I remember correctly), which is in conflict with the idea of minimizing allocations, but unavoidable if you want to deal correctly with branch points (and the limit imposed by the Sublime Text specification makes it reasonable, IMO). Let me know if this seems sensible or if you believe I'm missing something.

stefanobaghino:
**What changed since the original `DocumentGenerator` design**

The PR has evolved significantly since the first round of commits (anchor: `d1d2cc4`). Posting this so reviewers don't have to reconstruct the journey from the commit graph.

### Where it started

The first version of this PR introduced a `ScopeRenderer` trait + a `DocumentGenerator<R: ScopeRenderer>` driver, both designed primarily to address #626 (`SCOPE_REPO` replacement). `ClassedHTMLGenerator` became a thin newtype wrapper around `DocumentGenerator<HTMLScopeRenderer>` and the public API stayed line-oriented.

That shape worked for #626 — atom strings became part of the trait, downstream consumers no longer needed direct repository access — but it left two structural problems on the table:

1. **It didn't address #617.** `DocumentGenerator` was still line-oriented in the same way `HighlightLines` was: callers fed in lines, drove parsing themselves, and had to be aware of any speculative state. The whole point of #617 is to *hide* that machinery from the user.

2. **It silently inherited #614's cross-line correctness hazard.** Once branch-point parsing landed (#614), any line-by-line API has the same problem: when the parser retroactively replays a span of previously-emitted ops, the rendered output for those lines is already in the caller's hands and cannot be retracted. `DocumentGenerator` was structurally identical to `HighlightLines` in this respect — both produce wrong output for syntaxes that exercise cross-line branch-point failures, and there's no way for the caller to know it's happening unless they understand the syntax internals.

### Where it landed

Rather than ship "another `HighlightLines`-shaped API that has the same hazard," the design pivoted toward a streaming model:

- `DocumentGenerator` was replaced by **`syntect::io::HighlightedWriter`**, which implements `std::io::Write` and handles branch-point speculation internally. Bytes go in, rendered output flows to a sink, the buffering/replay machinery is invisible to the caller. This is a direct implementation of #617.

- The line-by-line APIs that *can't* be made cross-line-correct without restructuring (`HighlightLines`, `HighlightFile`, `line_tokens_to_classed_spans`) are now **deprecated** with notes pointing at `HighlightedWriter` / `ClassedHTMLGenerator`. They still compile and still work for grammars without branch-point hazards — the deprecation just makes the previously-implicit foot-gun visible.

- The `ScopeRenderer` trait got refined into a **layered design**. The original single trait was trying to serve two very different audiences: stateless renderers (the wastebin case from #626 — CSS class wrappers, XML, etc.) and stateful theme-aware renderers (ANSI escape codes, inline-styled HTML, LaTeX `\textcolor`). The single trait carried dead clutter for the first group and forced the second group to manually re-implement the same `Highlighter` + style-stack + merge-state boilerplate.

  After the refinement:

  - **`ScopeMarkup`** is the slim trait for stateless renderers. Drops `_scope` / `_scope_stack` / the always-`true` bool return; receives only pre-resolved atom strings. This is the trait wastebin would implement.
  - **`StyledOutput`** + **`ThemedRenderer<O>`** is the small format-specific surface for theme-aware renderers. Implementors only describe `begin_style` / `end_style` / `write_text`; the adapter manages the highlighter, style stack, and adjacent-token merging. Writing a new themed format (the latex-demo) is now ~25 lines instead of ~80.
  - **`ScopeRenderer`** stays public as the low-level escape hatch for advanced cases that need raw `Scope` / `&[Scope]` access.

### Side benefits that fell out

- **Better ecosystem integration.** `HighlightedWriter` implements `std::io::Write`, so it composes naturally with `io::copy(&mut source, &mut writer)`, `write!`/`writeln!` macros, and any `io::Write` sink (`stdout().lock()`, `Vec<u8>`, network streams, decompressors, …). The `syncat` example dropped its bespoke `BufReader::read_line` loop in favour of a single `io::copy` call.

- **Examples shrank.** `latex-demo.rs` went from a custom 80-line `ScopeRenderer` to a 25-line `StyledOutput` impl. `syncat.rs` lost its line-management loop. `parsyncat.rs` similarly slimmed.

- **#611 becomes a small follow-up.** Line highlighting is no longer a feature `ClassedHTMLGenerator` has to bake in — it's a ~15-line `ScopeMarkup` decorator that uses `begin_line` / `end_line` to wrap selected lines in `<span class="hl">…</span>`. See the dedicated section in the (revised) PR description for a concrete sketch.

- **`Drop` cleanup.** `HighlightedWriter` runs end-of-input cleanup (drain trailing partial line, flush pending speculative lines, close any scopes still open at EOF) implicitly via `Drop` on a best-effort basis. Callers that want errors propagated or want the inner sink back call `into_inner` explicitly — matching `BufWriter` / `BufReader` / `Cursor` convention. The `latex-demo` example showcases the Drop-only pattern (writes to `stdout().lock()`, no need to recover the sink).

### Issues this addresses, in order of "how directly"

| Issue | Status |
|---|---|
| #626 — `SCOPE_REPO` replacement | ✅ Removed; the new traits hand atom strings directly to renderers |
| #617 — easy branch-point-aware API | ✅ `HighlightedWriter` is exactly this |
| #614 — branch-point cross-line hazard | ✅ Streaming writer handles correctly; line-by-line APIs deprecated |
| #611 — line highlighting in `ClassedHTMLGenerator` | 🟡 Architecturally enabled (~15-line `ScopeMarkup` wrapper); concrete impl best handled as a follow-up so this PR stays focused |

### What the diff looks like

Roughly +1900 / -700 since `d1d2cc4` across 20 files. The main moving parts:

- `src/io.rs` (new) — `HighlightedWriter` + `HighlightedWriterBuilder` + the `Write` impl + `Drop`
- `src/rendering.rs` (new) — the three traits, the two adapters, `AnsiStyledOutput`, `render_line` (now `pub(crate)`)
- `src/html.rs` — `ClassedHTMLScopeRenderer` migrated to `ScopeMarkup`; `HtmlStyledOutput` added; legacy `line_tokens_to_classed_spans` deprecated
- `src/easy.rs` — restored to `master` content modulo `#[deprecated(since = "6.0.0")]` annotations on `HighlightLines` / `HighlightFile`
- examples + bench helper updated to the new APIs
- `tests/snapshots/public-api.txt` and `testdata/test{3,4,5}.html` regenerated

src/rendering.rs
```diff
@@ -0,0 +1,611 @@
...
+pub struct AnsiStyledOutput {
+    include_bg: bool,
+}
+
+impl AnsiStyledOutput {
+    /// Create a new ANSI emitter.
+    ///
+    /// If `include_bg` is `true`, the background colour escape code is also
+    /// emitted alongside the foreground.
+    pub fn new(include_bg: bool) -> Self {
```
keith-hall: I wonder if is worth having an enum or something for the `include_bg` because otherwise the call site gives no indication what the true/false means

src/util.rs
```diff
@@ -108,8 +110,8 @@ pub fn as_latex_escaped(v: &[(Style, &str)]) -> String {
             "{}\\textcolor[RGB]{{{},{},{}}}{{",
             if first { "" } else { "}" },
             style.foreground.r,
-            style.foreground.b,
-            style.foreground.g
+            style.foreground.g,
```
keith-hall, line 113: Is there a regression test for this bug fix?

stefanobaghino:
I'm temporarily putting this on hold to work on #618 and ensure things are correct before moving further.

src/util.rs
stefanobaghino: I added plenty of tests, [including one for this](https://github.com/trishume/syntect/pull/627/changes#diff-4007187965c1fea4fd252a76f5f1007b2ea59f6d86e2f454edee547c5036be24R419-R436). 🙂

src/rendering.rs
stefanobaghino: How about just exposing two pure stateless instances, one with the default current behavior (with background) and one without background?

keith-hall: Sounds like a good idea 🙂

stefanobaghino:
Now that #692 (finally!) removes all errors after bumping to the built-in packages to ST build 4202 (as tracked by #631), I'll iterate on this. I don't currently expect the PR to have this exact shape. Since the overall experience with stacked PRs seemed like a net positive (do challenge me if that's not the case), I'll probably structure as such when it approaches a ready-state. For now, I'd like to get this PR to a reasonable shape and see how a migration would happen for a number of relevant consumers of this library. The extent of the changes necessary to modernize syntect is such that I anyway don't expect that to land any time soon. Furthermore, I think that now that end-to-end correctness is pinned down, there's room to refactor `parser.rs` for increased maintainability, so this work is not the end of the story either, but rather a stop along the way to validate that the direction serves downstream consumers.

src/rendering.rs
stefanobaghino: Done in https://github.com/trishume/syntect/pull/627/changes/81cd64bdf77aa99c357a2db2896052a046110dd7..3644b80a7ce605b1948db136646c4695b9e78328

PR #629, opened 2026-04-12
Add zsh-patina to list of projects using syntect

michel-kraemer:
This pull request adds [zsh-patina](https://github.com/michel-kraemer/zsh-patina) to the list of projects using syntect.

zsh-patina is a Zsh plugin using `syntect` for syntax highlighting of commands while they are being entered.

Thank you so much for your work! Without `syntect`, zsh-patina would not have been possible! 🤗


keith-hall (approved):
Nice, thanks

PR #630, opened 2026-04-12
Modern syntax definition support

### LINKED ISSUE #618: Modern syntax definitions support

stefanobaghino:
Closes #618.

## Summary

Bumps `testdata/Packages` from the 2017 ST3-era snapshot to current upstream master (ST4-era) and fixes all issues preventing modern `.sublime-syntax` files from loading and highlighting correctly.

### Loading fixes (commit 1)

The original report identified three loading failures. All are fixed:

- **Scopes with >8 atoms** (`TooManyAtoms` for ActionScript): truncate to 8 atoms instead of erroring — the first 8 still provide meaningful matching.
- **Missing `contexts:` key** (`MissingMandatoryKey` for Batch File (Compound)): allow extends-only syntaxes with no `contexts:` section.
- **Broken regexes from unresolved parent variables** (HTML extending HTML (Plain)): defer regex validation for extends-based syntaxes; `re_resolve_all_regexes` fixes them up after the parent's variables are merged.

Three additional issues were found and fixed during testing:

- **`clear_scopes` double-Clear for `set:` operations** causing the white-background artifact in nested JS functions (the rendering bug from the original report).
- **`lines_include_newline` hardcoded to `false`** in `resolve_extends`, breaking regex re-compilation for patterns containing `\n` in repetition groups.
- **`__start` context not regenerated** after extends, causing the child syntax's scope to be lost.

### Embedded syntax highlighting fixes (commits 2–3)

After the loading fixes, all 189 syntax files load but embedded highlighting (JS/CSS in HTML, SQL in Ruby heredocs, etc.) was still broken. Root causes:

- **`pop` + `embed` combination silently ignored** (93 occurrences across 74 syntax files): the YAML loader's if-else chain checked `pop` before `embed`, so `embed:` was never reached. Fixed by adding `pop_count` to `MatchOperation::Embed`.
- **`pop` + `branch` combination silently ignored** (20 occurrences across 20 syntax files): same class of bug. JavaScript's arrow function disambiguation uses this pattern — without it, every identifier caused a cascade of `else-pop`s that emptied the class body scope stack. Fixed by adding `pop_count` to `MatchOperation::Branch`.
- **Escape regex not re-resolved after extends**: `re_resolve_all_regexes` re-resolved match regexes but not escape regexes in `Embed` operations. For extends-based syntaxes, escape patterns referencing parent variables (e.g. `{{script_content_end}}`) resolved to empty strings, causing embeds to terminate immediately.
- **Nested escape dispatch used wrong entry**: `exec_escape` always used `escape_stack[0]` (outermost). When an inner embed's escape fires (e.g. ERB `%>` inside a JS `<script>` embed), the wrong captures and stack depth were used. Fixed by threading the matching escape index from `find_best_match` through to `exec_escape`.

### JavaScript in `syncat`

618-js

### YAML with InspiredGitHub theme (`testdata/test3.html`)

618-test4

### Ruby/Rails with embedded content (`testdata/test4.html`)

618-test3

## Test plan

- [x] `cargo test` — all 186 tests pass
- [x] `cargo clippy --all-targets` — no new warnings
- [x] `cargo fmt --check` — clean
- [x] `make assets` — zero syntax-skipping warnings
- [x] Visual inspection


keith-hall:
I see some failures in CI when fancy-regex is used instead of Oniguruma. As a quick check, does pointing syntect to fancy-regex's main branch make any difference? There are a load of compatibility fixes not released yet...

stefanobaghino:
Follow-up to the bumped `testdata/Packages` (#618): eight parser state-machine fixes that close ~25,900 `syntest` assertion failures across both regex backends.

Each commit is a discrete bug with a captured regression test:
- syntest harness recognising all four annotation markers
- `pop: N` unwinding every popped context's meta scopes
- `bp_index` guard across cross-line replay
- branch-point unwinding when all alternatives fail
- branch match's own `pat.scope` surviving fail-retry
- no double-insert of `top_level_scope` on `extends`
- `Restore` ordering in `set` from a context with `clear_scopes`
- cross-line replay resumes after the branch match instead of re-parsing from col 0

Files previously panicking under `catch_unwind` (js.js, typescript.ts, and Haskell / Ruby / Rails html.erb on fancy-regex) now run to completion and surface their real assertion failures honestly — the summary counts look worse on those files but are more accurate.

Remaining failures cluster around two areas that need larger design work: `clear_scopes` semantics on non-last contexts of a multi-push (Python t-strings, C#11, Rails ERB) and the 128-line branch-point rewind horizon (TSQL, Haskell).

Runs green on `cargo test`, `make syntest`, and `make syntest-fancy`.

Is this a good state to land, or would you rather I keep going on the remaining clusters as part of this PR?

stefanobaghino:
One note: `examples/syntest.rs` skips `Perl/syntax_test_perl.pl` because the bumped Perl syntax hangs the parser on its POD-embedded language sections (zero-width lookahead escape `(?=^{{pod}})` combined with `embed` + `branch_point` inside the embedded JSON/SQL/HTML syntaxes).

The hang is reproducible from the first commit that bumps `testdata/Packages` — pre-existing, not introduced by any of the fixes in this PR. A local investigation confirmed neither existing loop guard catches it (`would_loop` doesn't treat a zero-width escape match as non-consuming, and the `stack.len() > 100` check doesn't trip since the stack stays shallow).

The skip is narrow (one file, logged when running without `--summary`, invisible in the known-failures diff). A proper fix likely means extending `would_loop` to cover zero-width escape matches, or short-circuiting embed re-entry at an unchanged cursor position — both want a small synthetic repro rather than the 5899-line Perl file as the primary test case. Happy to do that as a follow-up if you'd like it in this PR.

michaelblyons:
This is the `comment-pod-verbatim-body` context in *Perl.sublime-syntax* because of zero-length `escape` matches being rewound in a branch or something?

Is there a mechanism for patching the syntax to skip the embedded languages and then just list known test failures?

src/parsing/metadata.rs
```diff
@@ -500,13 +500,13 @@ mod tests {
         let r = RawMetadataEntry::load(comments_file);
         assert!(r.is_ok());
 
-        let indent_file: &str = "testdata/Packages/Go/Indentation Rules.tmPreferences";
+        let indent_file: &str = "testdata/Packages/Go/Indention Rules.tmPreferences";
```
michaelblyons: We should just open a PR on https://github.com/sublimehq/Packages/tree/master/Go to fix the filename typo.

src/parsing/parser.rs
```diff
@@ -1686,8 +2003,8 @@ mod tests {
         let mut state1 = ParseState::new(syntax);
         let mut state2 = ParseState::new(syntax);
 
-        assert_eq!(ops(&mut state1, "class Foo {", ss).len(), 11);
-        assert_eq!(ops(&mut state2, "class Fooo {", ss).len(), 11);
+        assert_eq!(ops(&mut state1, "class Foo {", ss).len(), 13);
+        assert_eq!(ops(&mut state2, "class Foo {", ss).len(), 13);
```
michaelblyons, line 2043: This is just the same test as the line above, no?

src/parsing/syntax_set.rs
```diff
@@ -2094,6 +2121,110 @@ mod tests {
...
+            name: Child
+            scope: source.child
+            file_extensions: [child]
+            extends: Base.sublime-syntax
+            contexts: {}
+            "#,
+            true,
+            None,
+        )
+        .unwrap();
```
michaelblyons, line 2164:
```suggestion
        let base = SyntaxDefinition::load_from_str(
            r#"
            name: Base
            scope: source.base
            file_extensions: [base]
            variables:
              ident: '[a-z]+'
            contexts:
              main:
                - match: '{{ident}}'
                  scope: variable.base
            "#,
            true,
            None,
        )
        .unwrap();
        // No `variables:` section — ident comes solely from Base.
        let child = SyntaxDefinition::load_from_str(
            r#"
            name: Child
            scope: source.child
            file_extensions: [child]
            extends: Base.sublime-syntax
            contexts:
              main:
                - match: '\${{ident}}'
                  scope: variable.child
            "#,
            true,
            None,
        )
        .unwrap();
```

src/parsing/syntax_set.rs
```diff
@@ -2094,6 +2121,110 @@ mod tests {
...
+        builder.add(child);
+        let ss = builder.build();
+
+        let syntax = ss.find_syntax_by_name("Child").unwrap();
+        let mut parse_state = ParseState::new(syntax);
+        let ops = parse_state
+            .parse_line("$abc\n", &ss)
+            .expect("parse failed")
+            .ops;
+        let expected = (0, ScopeStackOp::Push(Scope::new("variable.base").unwrap()));
```
michaelblyons:
```suggestion
        let expected = (0, ScopeStackOp::Push(Scope::new("variable.child").unwrap()));
```

src/parsing/metadata.rs
michaelblyons: https://github.com/sublimehq/Packages/pull/4519

stefanobaghino: Thanks for fixing this upstream in sublimehq/Packages#4519, Michael! I've bumped the `testdata/Packages` submodule to pick up the rename and updated this test path in the last two commits (fcdc076, 869b9bf).

src/parsing/parser.rs
stefanobaghino: Good catch — the duplicate is a symptom, not the intent. The original test ran `class Foo {` vs `class Fooo {` to assert that `ParseState` equality was structural (same contexts/scopes regardless of matched-text bytes). That invariant no longer holds after the cross-line branch-replay work: `ParseState::pending_lines` retains the literal source text while branch points are unresolved (and Java's `declarations` branch is still open after `class X {`), so different class names now compare unequal.

When I bumped the counts from 11 → 13 for the Packages upgrade I copy-pasted the same literal on both sides and lost the `Fooo` marker without leaving a note. I've just pushed a commit that keeps both inputs identical but adds a comment explaining why, and what the test asserts now (identical input → equal states; advancing one state → divergence).

stefanobaghino: https://github.com/trishume/syntect/pull/630/commits/d5662f691ae3eb0aaf9d7470574c541805343057

keith-hall:
> Is this a good state to land, or would you rather I keep going on the remaining clusters as part of this PR?

Maybe let's leave it here for this PR, review it and then queue up other fixes in another PR to keep the diffs smaller and easier to review.

Side note: Seeing the branch points stay active for a long time is helpful for deciding whether or not the sublime-syntax file needs to be refactored to work with shorter branches. I think the problem is that Sublime doesn't surface this information nor does it offer a way to accept/commit a branch short of popping it off the stack. So I think it's great that syntect gives us more insight here 🥇

stefanobaghino:
> > Is this a good state to land, or would you rather I keep going on the remaining clusters as part of this PR?
> 
> Maybe let's leave it here for this PR, review it and then queue up other fixes in another PR to keep the diffs smaller and easier to review.

Sounds good, thanks for confirming. I marked the PR as ready for review and I'll post a series of comments giving some guidance into what each range of commits does to help with the review. I already have a few more changes ready to go, but I'll withhold them for now and open each as a separate PR. 🙇🏻 

> Side note: Seeing the branch points stay active for a long time is helpful for deciding whether or not the sublime-syntax file needs to be refactored to work with shorter branches. I think the problem is that Sublime doesn't surface this information nor does it offer a way to accept/commit a branch short of popping it off the stack. So I think it's great that syntect gives us more insight here 🥇

It's just too bad these are not changes that can go upstream, I guess. 🙂

stefanobaghino:
A reader's guide to the commit history. Each range below is shaped like a
standalone PR; the links walk the diff for just that range.

### 1 · [BASE…`84b73ee`](https://github.com/trishume/syntect/pull/630/changes/BASE..84b73ee97e667c1a77b3482f91e007f954264994) — Packages bump + load-phase fixes

Bumps `testdata/Packages` from the 2017 ST3-era snapshot to current upstream
master (ST4-era) and fixes every loading-phase regression that surfaces. All
189 `.sublime-syntax` files load after this range; remaining issues are
parse-time correctness, not load-time. Production fixes:

- Truncate scopes >8 atoms instead of rejecting the syntax (ActionScript).
- Allow `extends:`-only syntaxes with no `contexts:` (Batch File Compound).
- Defer regex validation when `extends:` is present; re-resolve after the
  parent's variables are merged (fixes child syntaxes like HTML referencing
  `{{letter}}` from HTML (Plain)).
- Fix `clear_scopes` double-Clear for `set:` — only one `Restore` was
  emitted, stranding a scope on the clear-stack and surfacing as
  white-background artifacts in nested JS functions.
- Thread `lines_include_newline` through `resolve_extends` so
  `re_resolve_all_regexes` uses the right newline mode (previously hardcoded
  `false`, which broke `\n`-in-repetition patterns in newlines mode).
- Regenerate `__start` after extends resolution so child syntaxes push their
  own file scope (`text.html.rails`) instead of the parent's.
- Support `pop + embed` on the same match (93 patterns across 74 files;
  primary cross-language embedding mechanism — JS/CSS in HTML, SQL in Ruby
  heredocs, TOML in Rust frontmatter).
- Support `pop + branch`, re-resolve escape regexes after extends, and
  dispatch nested escapes through the correct stack entry (ERB `%>` inside
  a JS embed).
- Update metadata test paths and relax `can_parse_includes` for backend
  differences in the `lines_include_newline=false` path.
- Skip `Perl/syntax_test_perl.pl` in syntest. The bumped Perl syntax's
  POD-embedded languages combine a zero-width lookahead escape
  (`(?=^={{pod}})`) with embed + `branch_point` in a way that wedges the
  parser indefinitely — reproducible from the first commit that bumps the
  submodule, pre-existing, not introduced by any fix in this PR. The skip
  is narrow (one file, invisible in `--summary` diffs) and unblocks CI,
  which was otherwise hitting the 1-hour GitHub Actions cutoff. A proper
  fix wants a minimal synthetic repro rather than the 5,899-line Perl file
  — left for a follow-up.

### 2 · [`1316705`](https://github.com/trishume/syntect/pull/630/changes/1316705caa559b168b1cddcf25fb42ac8b2ab907) — syntest harness: `@+` and `>` annotation markers

Stand-alone harness fix, worth its own range because it's the only change
that doesn't touch the parser. Sublime's syntax-test format uses four
annotation markers (`<-`, `^+`, `@+`, `>`); syntect's `examples/syntest.rs`
only recognised the first two. Lines containing only `@+` / `>` were
treated as source text, which drifted `test_against_line_number` onto the
annotation lines themselves, mis-mapping every downstream `^` assertion.

The match anchors at the start of the post-testtoken substring (so the `>`
in HTML `-->` doesn't trigger) and the `regex` crate's lack of lookaround is
worked around in Rust. Scope checking for `@+`/`>` is not implemented — just
recognising them is enough to keep line tracking correct. Closes ~720
failures each on Oniguruma and fancy-regex (Pascal, Haskell, LaTeX bibtex,
Lisp, Cabal, Zsh). Ten unit tests capture the rules and false-positive
guards.

### 3 · [`1316705`…`9fbe03f`](https://github.com/trishume/syntect/pull/630/changes/1316705caa559b168b1cddcf25fb42ac8b2ab907..9fbe03fe167b64520a092c1e5d7715778cd9bdc2) — bulk failure elimination (~24,400 asserts)

Three parser state-machine fixes. The headliner is branch-point exhaustion;
the other two are shaped similarly (scope-stack bookkeeping around
multi-context operations).

- **`pop: N` meta-scope unwinding** — `pop: N > 1` previously only unwound
  the top context's scope atoms, orphaning the `meta_scope` /
  `meta_content_scope` of the N-1 deeper contexts. Concrete trigger is
  Makefile's `\)` popper in `variable-substitution-name-inside-paren`.
  Cleans up Makefile/CSS/Git rebase/LaTeX bib entirely; −540 net
  (+72 on C#11, +320 on Zsh — both downstream `clear_scopes` issues
  surfaced rather than caused).
- **`bp_index` guard across cross-line replay** — `parse_line_inner`
  recursion inside the replay can add/remove `branch_points`, invalidating
  the `bp_index` captured at the top of `handle_fail`. Adds a bounds check
  before the final bookkeeping write. Eliminates an `index out of bounds`
  panic on js.js / typescript.ts (those files were already caught by
  `catch_unwind`, so summary counts don't move, but it's the right kind of
  correctness fix).
- **Branch-point exhaustion unwinding** — when every alternative of a
  `branch_point` failed, the last alternative's pushed contexts and scope
  atoms were left on the stack, stuck forever in an unreachable context.
  Zsh's `brace-interpolations` deliberately omits the
  `brace-interpolation-fallback` alt, so `echo {no}` triggers this.
  Fix snapshots/restores parser state and advances past the branch-point
  position by one char. Same-line path only (cross-line exhaustion left to
  the successful-retry path, where it already behaves). **Zsh
  scope.zsh: 23,771 → 0**, Haskell −85.

Together these close ~24,400 assertion failures.

### 4 · [`9fbe03f`…`0070c91`](https://github.com/trishume/syntect/pull/630/changes/9fbe03fe167b64520a092c1e5d7715778cd9bdc2..0070c91fe7d317428365e3a7c0bee53e700e2d07) — smaller clusters (~720 asserts + 5 panics)

Four more state-machine fixes, each smaller but hitting distinct categories.

- **Branch match's `pat.scope` across fail-retry** — on retry, the branch
  match's own scope (e.g. `keyword.operator.comparison.sql` on `LIKE`) was
  truncated off `ops` with alt[0]'s work and never re-emitted; the winning
  alternative ended up with no scope on the trigger keyword. ~260 asserts.
- **Stop double-inserting `top_level_scope` in extending syntaxes** —
  `add_initial_contexts` ran twice when `extends:` was present (initial
  load + `resolve_extends`). Second pass pushed the file scope onto
  `__main.meta_content_scope` again, so Git Diff and Rails ERB variants
  appeared with `[source.diff.git, source.diff.git]` on every line.
  ~227 asserts.
- **`Restore` ordering in `set:` from a `clear_scopes` context** — the
  `Restore` emitted in `push_meta_ops`'s initial phase landed *above* the
  target's pushed `meta_scope`; the non-initial `Pop` then ate the restored
  atom instead of the intended meta-scope. Moves `Restore` to the
  non-initial phase, after the `Pop`. **JSON 3 → 0**, JS typescript_template
  22 → 0, plus smaller wins in JS templates and Rails.
- **Resume cross-line replay after the branch match, not column 0** — the
  first buffered line's pre-branch prefix was re-parsed under the new
  alternative (which only applies after the trigger). New
  `parse_line_inner_from` helper takes a starting column; emits the
  pre-branch prefix verbatim plus the trigger's pat.scope, then resumes at
  match_end. **C#8: 40 → 0**, big JS wins, and — the real correctness story
  — **eliminates parser panics** on js.js, typescript.ts, and Rails html.erb
  / Ruby / Haskell on fancy-regex. Those files were previously swallowed by
  `catch_unwind`; they now surface their real (pre-existing) assertion
  counts honestly.

### 5 · [`0070c91`…`285144e`](https://github.com/trishume/syntect/pull/630/changes/0070c91fe7d317428365e3a7c0bee53e700e2d07..285144e0633b37fb075128316477cca845a4b767) — follow-up fix + re-bump + review polish

Mixed range driven by review feedback and one more parser fix.

- `cargo fmt` and a small doc-comment quoting fix.
- **Single-context `set:` fires target's `clear_scopes` pre-match** — for a
  `set:` whose target declares `clear_scopes`, the `Clear` was emitted in
  the non-initial phase, after the trigger token had been scoped. The outer
  `meta_content_scope` survived through the trigger's match span (visible
  on Lisp's `(defun fn (...)` parameter list). Moves `Clear` to the initial
  phase **only for single-context `set:`** — Makefile's multi-context
  `set: [value-to-be-defined, eat-whitespace-then-pop]` relies on the
  late Clear stripping the preceding-context mcs, not the file scope.
  Lisp 2 → 0.
- **Submodule re-bump** `f61fa0a1 → 1ba99a47`, plus tracking one
  upstream filename rename (Go indentation rules).
- **Documentation & review response**: document the weakened
  `can_compare_parse_states` invariant; apply two suggested edits to
  `src/parsing/syntax_set.rs`; rewrite the
  `extends_variable_inherited_without_override` comment for clarity.

Net across the PR: ~25,950 syntest assertions eliminated, 23 regression
tests captured, five parser panics turned into honest (surfaced) assertion
counts. Remaining clusters (TSQL, Haskell cross-line, Rails ERB /
Python t-strings / C#11 `clear_scopes`-on-non-last-context) are 
follow-up candidates; per the thread above they'll go in
subsequent PRs.

stefanobaghino:
Opened #631 as a tracking issue for the `syntest` failures left over after this PR lands — subsequent per-cluster PRs can reference it so progress stays visible in one place. Worth considering for the [syntect 6.0.0](https://github.com/trishume/syntect/milestone/1) milestone, since some of the remaining fixes may involve parser-behaviour or public-API changes.

stefanobaghino:
@keith-hall Is there some way in which I can make this easier to review?

keith-hall:
Sorry for the delay, I have lots of things to do, plus a day job and a family, so I don't get much time left for other tasks. The commit breakdown is very helpful, thanks for that. I will try to finalize my review of this PR tonight.

stefanobaghino:
> Sorry for the delay, I have lots of things to do, plus a day job and a family, so I don't get much time left for other tasks. The commit breakdown is very helpful, thanks for that. I will try to finalize my review of this PR tonight.

No, I'm sorry if that came across as trying to nudge. I'm genuinely trying to understand if I can make this easier and appreciate the time you are putting into this. You have absolutely nothing to be sorry about, and if I can help I'm happy to. Please take the time you need for other commitments before syntect. 🙂

keith-hall (approved):

Issue #631, opened 2026-04-14
Track remaining syntest failures after #630


stefanobaghino:
Tracking issue for the `syntest` assertion failures that remain after
#630 lands.

#630 bumps `testdata/Packages` to upstream master and fixes the
loading-phase and a first wave of parse-phase issues that the bumped
Packages expose (~25,950 assertion failures closed, 23 regression tests
captured). Several clusters intentionally remain out of scope — they
want larger design work or targeted per-cluster repros — and will be
addressed in follow-up PRs.

This issue exists as a single place to link those follow-ups from, so
that the overall progress against the current `known_syntest_failures*`
baselines stays visible.

### How to contribute

- Open a PR for one cluster at a time (keeps review tractable).
- Reference this issue from the PR description.
- Update `testdata/known_syntest_failures.txt` /
  `testdata/known_syntest_failures_fancy.txt` and add a regression test
  shaped around a minimal synthetic repro where possible.

### Suggested milestone

Would fit under milestone [syntect 6.0.0](https://github.com/trishume/syntect/milestone/1)
— the fixes may involve parser-behaviour or public-API changes that are
easier to ship as part of a semver-major bump.

### Linked PRs

Stacked on #630, ordered bottom-up:

- [x] #636 — Use engine `FIND_NOT_EMPTY` flag on scope-only match patterns (supersedes #624, #632, #633, #634)
- [x] #635 — Apply branch alternative's `meta_scope` to trigger on fail-retry
- [x] #638 — Inherit `meta_include_prototype` from parent on `meta_append` / `meta_prepend` (closes TSQL `LIKE` failures)
- [x] #639 — Apply `branch_point` exhaustion state-unwind to cross-line case (closes TypeScript / C#9 failures)
- [x] #640 — Ungate `exec_escape` mcs-pop skip from guest syntax version (closes Rails HTML ERB failures)
- [x] #641 — Syntest visibility and v2 `clear_scopes` parity (closes Python f/t-string and Bash `def-function` failures; surfaces previously-masked scope-stack-error bailouts)
- [x] #642 — Support `pop: N + set: X` as a combined match operation (closes JavaScript `js.js` failures)
- [x] #643 — Skip stale same-name `branch_point`s in fail-retry lookup (closes Haskell raw-string quasi-quote failures, 271 → 141)
- [x] #644 — Restrict non-consuming `pop` loop guard to `pop:1` (closes Haskell Cluster B + JS class, 141 → 72)
[...]

PR #632, opened 2026-04-14
Strip top-level empty alternatives from compiled regexes

stefanobaghino:
First of a stack of follow-ups to #630 chipping away at the remaining `syntest` failures tracked in #631.

**Review just this PR's changes:** [`issue-618-modern-syntax-support...631-strip-top-level-empty-alts`](https://github.com/stefanobaghino/syntect/compare/issue-618-modern-syntax-support...631-strip-top-level-empty-alts) — diff excluding #630's own contents.

Cabal's operator regex `\|\||&&||!` has a stray empty alternative between `&&` and `!` — under leftmost-first matching, the empty alt wins zero-width at the `!` position and `!` is never scoped. Same shape shows up elsewhere in Packages, so fix the class: pre-process compiled regex patterns to collapse runs of consecutive unescaped `|` at the **top level** (outside `(...)` groups and `[...]` classes). In-group empties are left alone — D's assignment regex `(...|>>>||\*|...)=` relies on a middle empty alt making the group optional.

Impact: Cabal cabal.cabal 1 → 0 (both backends). Tests capture the stripping rules and a Cabal end-to-end reproduction.

Stacked on #630. GitHub can only diff against `master`, so this PR's file changes currently show #630's contents as well; the commit list shows only the one commit that belongs to this PR. Merge only after #630, or (better) after #630 lands rebase this branch directly onto `master`.

Refs: #631


stefanobaghino:
Superseded by #636 per keith-hall's suggestion — #636 uses the regex engine's native `FIND_NOT_EMPTY` flag on scope-only match patterns instead of preprocessing regexes to strip empty alternatives. On Oniguruma the same Cabal cabal.cabal 1 → 0 win, plus wider coverage across the other empty-alt shapes without the stripper.

PR #633, opened 2026-04-14
Strip leading empty alternatives inside groups

stefanobaghino:
Follow-up to #630 in the stack started by #632. Extends #632's top-level empty-alt stripper to cover the case *inside* a group.

**Review just this PR's changes:** [`631-strip-top-level-empty-alts...631-strip-leading-empty-alts-in-groups`](https://github.com/stefanobaghino/syntect/compare/631-strip-top-level-empty-alts...631-strip-leading-empty-alts-in-groups) — diff excluding the previous stack entry.

Rust's `prelude_types` variable is written

    (?x:
      # std::boxed
      |Box
      # std::option
      |Option|Result
      …
    )

The author prefixed every alternative with `|`, including the first. Under leftmost-first matching the leading empty alt wins zero-width at every position, so `\b{{prelude_types}}\b` never matches `Box` / `Option` / `Vec` / `String` — they fall through to `storage.type.rust` instead of `support.type.rust`. The same shape appears in `std_traits`.

#632 deliberately left in-group empty alts alone (D's assignment regex relies on a middle empty alt to make the group optional). This PR extends the stripper to drop **leading** empty alts inside a group while still preserving middle/trailing empties. Extended-mode (`(?x:…)`) whitespace and `#…\n` comments between `(?x:` and the leading `|` pass through so the Rust shape is still recognised as a leading empty alt.

Impact (both backends): all seven Rust syntest files pass cleanly — 46 assertion failures closed per backend.

Refs: #631. Stacked on #632 — review/merge after #632 lands.


keith-hall:
I'm pretty confident that the correct solution here is https://github.com/trishume/syntect/pull/624

stefanobaghino:
Superseded by #636 — the `FIND_NOT_EMPTY` engine flag makes the Rust `prelude_types` shape `(?x:|Box|Vec)` match its real alternatives on both backends. Same Rust wins (46 asserts on Oniguruma, same on fancy-regex) without the in-group leading-empty-alt stripping.

PR #634, opened 2026-04-14
Retry search past zero-width scope-only matches

stefanobaghino:
Follow-up to #630 in the stack started by #632.

**Review just this PR's changes:** [`631-strip-leading-empty-alts-in-groups...631-retry-zero-width-scope-only-matches`](https://github.com/stefanobaghino/syntect/compare/631-strip-leading-empty-alts-in-groups...631-retry-zero-width-scope-only-matches) — diff excluding the previous stack entry.

A scope-only (`MatchOperation::None`) pattern that matches zero-width at the current position is skipped to avoid infinite loops, but discarding it entirely also hid any non-empty match of the same pattern later on the line. In Ruby's `?\u{012ACF 0gxs}`, `\h{0,6}` matched zero-width at the space and was dropped, so the lower-priority `\S` fallback scoped the `0` as `invalid.illegal.escape` instead of `constant.numeric.value`.

After the zero-width skip, retry the regex from one character past `match_end` until a non-empty match is found or the line is exhausted. Do not cache `None` after a zero-width skip — that result is only authoritative from the caller's `start`, and the cache is keyed by pattern only.

Impact: Ruby `syntax_test_ruby.rb` 1 → **0** (both backends).

Refs: #631. Stacked on #633 — review/merge after #630, #632 and #633 land.


keith-hall:
Again I'm pretty sure the correct solution is to set the regex engine option to ignore empty matches https://github.com/trishume/syntect/pull/624

stefanobaghino:
> Again I'm pretty sure the correct solution is to set the regex engine option to ignore empty matches #624

Yup, thanks for pointing me to that. How can #624 make progress so that I can rebase everything on top of it and drop this and #633?

keith-hall:
Now that there is a pull request for fancy-regex and we are unsure when it will get reviewed, it may make sense to unblock ourselves by pointing our dependency of fancy-regex to that branch temporarily, and redo the changes properly with that in mind, so both Oniguruma and fancy-regex features would use the respective engine flags on `MatchOperation::None` match patterns. 

If you agree, and want to take it on, feel free to make a new PR, and I can close #624 - it was originally there to serve as an example anyway, with the option to improve or rework as we see fit.

stefanobaghino:
Superseded by #636 per https://github.com/trishume/syntect/pull/634#issuecomment-4246441608. The engine's `FIND_NOT_EMPTY` flag handles the Ruby `\h{0,6}` zero-width-hiding-later-non-empty case at the regex level, avoiding the retry loop and the cache-semantics complication.

Issue #240, opened 2019-03-09
bug: `tokens_to_classed_spans` is private


PoiScript:
The `tokens_to_classed_html` function is deprecated since 3.1.0 in favor of `tokens_to_classed_spans`. But `tokens_to_classed_spans` is actually [private](https://github.com/trishume/syntect/blob/6d6cf175a221544a1a2136bbe559d7fd117fe1ad/src/html.rs#L161).

trishume:
Whoooops, will try and fix that and push a new release later today

trishume:
I published `v3.2.0` with this fixed.

PR #636, opened 2026-04-15
Use engine FIND_NOT_EMPTY flag on scope-only match patterns

stefanobaghino:
Follow-up to #630 chipping away at the remaining `syntest` failures tracked in #631. Supersedes #624, #632, #633, and #634 per https://github.com/trishume/syntect/pull/634#issuecomment-4246441608.

## Approach

Scope-only (`MatchOperation::None`) match patterns can silently stall the parser when the regex matches zero-width at the current position. Instead of working around this with regex preprocessing (#632, #633) or a retry loop (#634), push the constraint down to the regex engine itself: compile a second variant of each regex with `FIND_NOT_EMPTY` and use it only for `None`-operation matches. Every other operation (`Push`, `Set`, `Pop`, `Embed`, `Branch`, `Fail`) legitimately needs zero-width matches and continues to use the regular variant.

- **Oniguruma**: native `REGEX_OPTION_FIND_NOT_EMPTY`.
- **fancy-regex**: `RegexBuilder::find_not_empty(true)`, from fancy-regex/fancy-regex#240.

Cargo.toml points the `fancy-regex` dep at that PR's `find_not_empty` branch temporarily (`b8cd8fc1`). Once #240 lands and is released, a follow-up flips it back to a published version.

## Public API

`Regex::search` gains an `allow_empty: bool` parameter. Acceptable for the 6.0.0 milestone.

## Impact

Both backends now produce identical results on the affected files:

| File | Before | After |
|---|---:|---:|
| `Haskell/tests/syntax_test_cabal.cabal` | 1 | **0** |
| `Ruby/syntax_test_ruby.rb` | 1 | **0** |
| `Rust/tests/syntax_test_closures.rs` | 6 | **0** |
| `Rust/tests/syntax_test_enum.rs` | 6 | **0** |
| `Rust/tests/syntax_test_generics.rs` | 12 | **0** |
| `Rust/tests/syntax_test_identifiers.rs` | 1 | **0** |
| `Rust/tests/syntax_test_macros.rs` | 3 | **0** |
| `Rust/tests/syntax_test_struct.rs` | 3 | **0** |
| `Rust/tests/syntax_test_types.rs` | 15 | **0** |

48 assertion failures closed on both Oniguruma and fancy-regex.

## Tests

Three regression tests:

- `scope_only_pattern_that_matches_zero_width_finds_later_non_empty` — Ruby `\h{0,6}` at space hiding later `0`.
- `scope_only_pattern_with_leading_empty_alt_in_group_matches_name` — Rust `(?x:|Box|Vec)`.
- `scope_only_pattern_with_middle_empty_alt_matches_bang` — Cabal `\|\||&&||!`.

All three pass on both backends.

## Merge order

Draft until fancy-regex/fancy-regex#240 merges and is released. Then the Cargo.toml dep gets flipped back to a published version in a follow-up.

Refs: #631

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -322,9 +322,8 @@ checksum = "877a4ace8713b0bcf2a4e7eec82529c029f1d0619886d18145fea96c3ffe5c0f"
 
 [[package]]
 name = "fancy-regex"
-version = "0.16.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "998b056554fbe42e03ae0e152895cd1a7e1002aec800fdc6635d20270260c46f"
+version = "0.17.0"
+source = "git+https://github.com/fancy-regex/fancy-regex?branch=find_not_empty#b8cd8fc1f296be10b4cca3c82efd183b747639ed"
 dependencies = [
  "bit-set",
  "regex-automata",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ features = ["metadata"]
 [dependencies]
 yaml-rust2 = { version = "0.10.4", optional = true, default-features = false }
 onig = { version = "6.5.1", optional = true, default-features = false }
-fancy-regex = { version = "0.16.2", optional = true }
+fancy-regex = { git = "https://github.com/fancy-regex/fancy-regex", branch = "find_not_empty", optional = true }
 walkdir = "2.5"
 regex-syntax = { version = "0.8", optional = true }
 plist = { version = "1.8", optional = true }
--- a/src/parsing/parser.rs
+++ b/src/parsing/parser.rs
@@ -612,7 +612,7 @@ impl ParseState {
             let mut esc_regions = Region::new();
             if entry
                 .regex
-                .search(line, start, line.len(), Some(&mut esc_regions))
+                .search(line, start, line.len(), Some(&mut esc_regions), true)
             {
                 let (esc_start, _esc_end) = esc_regions.pos(0).unwrap();
                 if esc_start < search_end {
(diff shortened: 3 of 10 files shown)


keith-hall:
> Oniguruma: 48 assertion failures closed. fancy-regex: 23 closed. The remaining fancy-regex failures (Cabal, Rust closures, Rust types) come from a behavioural difference: fancy-regex's `FIND_NOT_EMPTY` rejects the whole match when the first alternative returns zero-width, while Oniguruma backtracks to the next alternative. Follow-up to be filed against fancy-regex.

Thanks for this, I have now pushed some changes to fancy-regex (same branch/PR) to fix this.

stefanobaghino:
Rebased onto current master (post-#630 merge) and bumped fancy-regex to `b8cd8fc1` — the latest tip of the `find_not_empty` branch, which includes @keith-hall's fixes for the backtracking-past-zero-width-alternatives behaviour.

Results: all three previously-failing fancy-regex cases now pass:

- Cabal `cabal.cabal`: 1 → **0**
- Rust `closures.rs`: 6 → **0**
- Rust `types.rs`: 15 → **0**

Both backends are now identical on the affected files — 48 assertion failures closed each. The `#[cfg(feature = "regex-onig")]` gate on the Cabal test is removed since it passes on both backends. Updated known-failures baselines accordingly.

Marked as ready for review.

I believe it's fine to point to the `fancy-regex` branch for now, but ultimately we'd want to point to a released version. My take would be that if there are more fixes that need to go into `fancy-regex`, we might want to continue pointing to specific branches until we are sure everything's fixed and `fancy-regex` can be released, to avoid inducing release churn. @keith-hall You are in a better position to make the call, of course. 🙂

keith-hall (approved):
Thanks. The plan is to make a fancy-regex release in exactly a week's time btw: https://github.com/fancy-regex/fancy-regex/issues/162#issuecomment-4231594861

PR #638, opened 2026-04-18
Inherit meta_include_prototype from parent on meta_append / meta_prepend

stefanobaghino:
Follow-up to #630, #635, #636. Chipping away at the remaining `syntest` failures tracked in #631.

## Root cause

The `meta_append: true` / `meta_prepend: true` extends-merge modes in `syntax_set.rs` copy `meta_scope`, `meta_content_scope`, and `clear_scopes` from parent when the child doesn't explicitly set them, but not `meta_include_prototype`. Since `Context`'s default for that field was `true`, a child that uses `meta_append` without restating the parent's `meta_include_prototype: false` ended up with the syntax's prototype attached to what should be a prototype-free context.

Concrete trigger: TSQL extends SQL (basic) and uses `meta_append: true` to add a `[…]` character-set rule to `inside-like-single-quoted-string`. The SQL base declares that context with `meta_include_prototype: false` so its `--` comment rule (from the SQL prototype) does NOT fire inside LIKE strings. TSQL didn't restate that, so after the merge the SQL prototype attached, and `--` characters inside LIKE string bodies were scoped as comments — cascading into 4,918 assertion failures across `syntax_test_tsql.sql`.

## Fix

Change `Context::meta_include_prototype` from `bool` to `Option<bool>`. `None` means the YAML never set the field; `Some(b)` means it was explicitly stated. The `Append` / `Prepend` merge branches inherit the parent's value when the child's is `None`, same shape as the existing `meta_scope` / `meta_content_scope` / `clear_scopes` inheritance. Consumers that check whether a context should include the syntax's prototype read the field with `.unwrap_or(true)` (Sublime's default).

Per [keith-hall's review suggestion](https://github.com/trishume/syntect/pull/638#issuecomment-4274601080), the `Option<bool>` shape replaces an earlier iteration that tracked explicit-vs-default via a parallel `meta_include_prototype_explicit: bool` field — one field is cleaner and removes the risk of the two drifting out of sync.

`parse_context` no longer threads `is_prototype` — the prototype context's self-attachment is suppressed by the `no_prototype` set in `SyntaxSetBuilder::link_syntaxes`, so no special initial value is needed per context type.

## Public API

Breaking:
- `Context::meta_include_prototype: bool` → `Context::meta_include_prototype: Option<bool>`.
- `Context::new(meta_include_prototype: bool)` → `Context::new(meta_include_prototype: Option<bool>)`.

Consistent with the 6.0.0 semver-major milestone that #631 targets. Packdumps (`assets/default*.packdump`) regenerated — serialized layout of `Context` changed with the field type. Public-API snapshot updated accordingly.

## Impact (both backends)

| File | Before | After |
|---|---:|---:|
| `SQL/tests/syntax/syntax_test_tsql.sql` | 4918 | **0** |
| `Rails/tests/syntax_test_rails.html.erb` | 273 | 264 |
| `Rails/tests/syntax_test_rails.haml` (Oniguruma only) | 94 | 86 |

Net: ~4935 assertion failures closed on Oniguruma, ~4927 on fancy-regex.

## Tests

- `meta_append_inherits_meta_include_prototype_from_parent` — synthetic regression test using a pair of YAMLs (parent with `--` comment prototype + `meta_include_prototype: false` base context + child that `meta_append`s a rule without restating the flag). After merge, `--` inside the base span must not take a `comment.*` scope.

## Notes

The 8 "branch point 'like-expressions' expired (exceeded 128-line rewind limit)" warnings still show up on the TSQL file. They're pre-existing — LIKE patterns without ESCAPE keep their branch_point alive until the stack drops below `bp.stack_depth` or 128 lines pass. Not in scope for this PR; worth filing separately if the warnings become noisy for downstream consumers.

Refs: #631


keith-hall:
> Track whether `meta_include_prototype` came from an explicit YAML setting on each context via a new `meta_include_prototype_explicit` flag

I wonder whether it would feel cleaner to have `meta_include_prototype` become an `Option<bool>`, then when it is None, it would be clear that it wasn't explicitly set, and save needing another flag

stefanobaghino:
Good call — switched to `Option<bool>` in the force-push. One field instead of the parallel `_explicit` bool, same merge behaviour (inherit on `None`, keep child on `Some`), and the default-true semantic is a `.unwrap_or(true)` at consumption sites.

Side effects:

- `Context::meta_include_prototype: bool → Option<bool>` and `Context::new` signature change — both breaking, tagged for the 6.0.0 track per #631.
- `parse_context` dropped its `is_prototype` parameter; the prototype context's self-attachment is already suppressed by the `no_prototype` set in `SyntaxSetBuilder::link_syntaxes`, so no special per-context initial value is needed.
- Packdumps (`assets/default*.packdump`) regenerated — `Context`'s serialized layout changed with the field type.
- `tests/snapshots/public-api.txt` updated.

All 177 library tests pass on both backends; `make syntest` / `make syntest-fancy` diff clean. Commit in place, stacked descendants (#639, #640) rebased onto the new tip.

keith-hall:
Looks good, I was considering approving and merging but I see it is still a draft. Anything still to be done for this PR?

stefanobaghino:
> Looks good, I was considering approving and merging but I see it is still a draft. Anything still to be done for this PR?

Nope, just marked as ready-for-review. 🙂

stefanobaghino:
Btw let me know if the current approach with stacked PRs is helpful, or if there's any change you would like to make. Happy to adapt the workflow to facilitate reviews.

keith-hall (approved):

keith-hall:
I like stacked PRs, I think everything is okay

PR #639, opened 2026-04-18
Apply branch_point exhaustion state-unwind to cross-line case

stefanobaghino:
Follow-up to #630, #635, #636, #638 (all merged). Chipping away at the remaining `syntest` failures tracked in #631.

## Root cause

The same-line `branch_point` exhaustion path at `handle_fail` already restores the pre-branch snapshot when all alternatives fail (commit `9fbe03f`), but the cross-line case just removed the branch record and returned `Ok(false)`. The last alternative's pushed contexts stayed on the state stack forever, contaminating every subsequent line.

Concrete trigger: `sublimehq/Packages#3598`'s intentionally-incomplete TypeScript type expression:

```typescript
type x = {
    bar: (cb: (
};
```

Two nested `ts-function-type` branches both exhaust across lines (alt[0] `ts-type-function-parameter-list` and alt[1] `ts-type-group` each fail — `(cb: (` is neither a valid parameter list nor a valid grouped type). Before the fix, `meta.type.js` / `meta.group.js` from the nested `(` groups stayed anchored on the scope stack and cascaded into 274 assertion failures in `syntax_test_typescript.ts` (plus another 10 in `C#9.cs` via a similar shape).

## Fix

Unify the two exhaustion paths in `handle_fail`:

1. Always restore `stack`, `proto_starts`, `escape_stack`, `first_line`, `non_consuming_push_at`, and truncate `ops` to `ops_snapshot_len`.
2. For the cross-line case, additionally re-parse the buffered lines under the restored (pre-branch) state and surface the corrected ops via `flushed_ops` / `ParseLineOutput::replayed`.

### Related fix: don't drain `pending_lines`

The cross-line retry and exhaustion paths no longer drain `pending_lines` — they slice-and-clone instead. Nested branches on the same line both need access to the buffered lines for their own replays; draining in the inner retry left the outer branch with nothing to replay on its subsequent retry/exhaustion.

## Impact (both backends)

| File | Before | After |
|---|---:|---:|
| `JavaScript/tests/syntax_test_typescript.ts` | 274 | **0** |
| `C#/tests/syntax_test_C#9.cs` | 10 | **0** |

Net: ~284 assertion failures closed on both Oniguruma and fancy-regex.

## Tests

- `cross_line_branch_exhaustion_unwinds_state` — synthetic shape with two alternatives that both push contexts on line 1 and both fail on line 2. The primary regression guard is the TypeScript test file staying out of `known_syntest_failures.txt`.

Refs: #631

### DIFF
--- a/src/parsing/parser.rs
+++ b/src/parsing/parser.rs
@@ -990,41 +990,95 @@ impl ParseState {
             // both `brace-interpolation-sequence` and
             // `brace-interpolation-series` failed and there was no
             // fallback alternative (Zsh explicitly excludes
-            // `brace-interpolation-fallback`). Same-line only — a
-            // cross-line branch_point that never succeeded is left
-            // alone here because replaying through it is handled
-            // via the cross-line path in the successful case.
+            // `brace-interpolation-fallback`).
+            //
+            // Cross-line exhaustion takes the same shape, plus a replay
+            // of the buffered lines under the pre-branch state so
+            // callers see corrected ops for lines they've already been
+            // handed. Without this, the unterminated TypeScript type
+            // expression at `sublimehq/Packages#3598`
+            // (`type x = { bar: (cb: (\n};`) left the inner
+            // `ts-type-function-parameter-list-body` on the stack
+            // forever, contaminating every subsequent line's scope
+            // stack with `meta.type.js, meta.group.js` — 274 cascading
+            // assertion failures in `syntax_test_typescript.ts`.
             let is_cross_line = bp.line_number < cur_line;
-            if !is_cross_line {
-                let stack_snapshot = bp.stack_snapshot.clone();
-                let proto_starts_snapshot = bp.proto_starts_snapshot.clone();
-                let escape_stack_snapshot = bp.escape_stack_snapshot.clone();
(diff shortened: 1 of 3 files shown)


src/parsing/parser.rs
```diff
@@ -3421,6 +3477,113 @@ contexts:
...
+        for (_, op) in &o {
+            stack.apply(op).unwrap_or(());
+        }
+        // This assertion is weak because we don't run the full
+        // replay/stack-reset that the syntest harness does. The
+        // concrete regression is guarded end-to-end by
+        // `syntax_test_typescript.ts` staying absent from
+        // `known_syntest_failures.txt`. Keep this as a fast unit
+        // probe that at least exercises the code path.
+        let _ = stack;
```
keith-hall: I would prefer to have regression tests which properly assert scopes, otherwise if the syntax test failures come back due to reworking some code, there is no easy way to identify what/where the problem is because this test might still pass.

stefanobaghino: https://github.com/trishume/syntect/pull/639/changes/81d287a2ad18b52c82cb89b859f2ceb1cf11c5e2

keith-hall (approved):

Issue #245, opened 2019-03-17
Example for returning HTML code of syntax-highlighted file using sublime-syntax file.


ChrisMGeo:
Sorry if this is bothersome, but I'm pretty new to Rust so I don't know how to do what I am asking with the current knowledge I have.

What I am asking for is this, say I have a string that contains the file's text, a string corresponding to the location of a  sublime-syntax file and a string corresponding to the location of a tmtheme file, how would I return a syntax-highlighted HTML string.

Memorion:
Here are the docs: https://docs.rs/syntect/3.2.0/syntect/html/index.html 
and here is a sample: https://github.com/trishume/syntect/blob/master/examples/synhtml.rs

ChrisMGeo:
@Memorion Thanks for the answer and sorry for the late reply, but I believe that the function I am looking for is syntect::html::highlighted_html_for_string(https://docs.rs/syntect/3.2.0/syntect/html/fn.highlighted_html_for_string.html), however how do i declare a SyntaxSet and SyntaxReference (sorry if that's wrong) with a sublime-syntax file.

trishume:
Look at the docs for `SyntaxSet` and `SyntaxSetBuilder`, there are a variety of ways to load files, folders and built in syntaxes. Then a bunch of methods to find a specific syntax.

ChrisMGeo:
@trishume Thanks for the feedback. However I have one small doubt/feature-request  -  From what I understand the `SyntaxSet` is a group of syntaxes (I think it is `SyntaxReference`) from which a single syntax is used for highlighting. Wouldn't it be better if you could do the `syntect::highlighted_html_for_string` with a `SyntaxReference` only and without a `SyntaxSet`.
Please do correct me if I'm wrong anywhere.

keith-hall:
my understanding is that, because many syntax definitions refer to other syntax definitions, the `SyntaxSet` is necessary. Otherwise PHP wouldn't have HTML and JS highlighted properly etc.

trishume:
Yah Keith is correct, if you want to highlight with only one syntax you can make a set of only one syntax. Additionally the `SyntaxReference` is actually just a lightweight reference into the `SyntaxSet`.

Enselic:
This seems sorted out. Let's close this issue now.

PR #644, opened 2026-04-20
Restrict non-consuming pop loop guard to pop:1

stefanobaghino:
Stacked on #643. **Isolated diff:** https://github.com/stefanobaghino/syntect/compare/631-haskell-quasi-quote...631-haskell-decl-data-pop

## Summary

`find_best_match`'s `pop_would_loop` guard (`src/parsing/parser.rs`) flagged *any* non-consuming `Pop` after a non-consuming push as looping, but `pop: N` with N ≥ 2 drops the stack strictly below the pre-push depth — the outer context no longer has the same trigger in scope, so no loop is possible. Restrict the guard to `Pop(1)`.

## Impact

- `Packages/Haskell/tests/syntax_test_haskell.hs`: **141 → 72**. Closes Cluster B: `meta.declaration.data` from the enclosing `constructors` `meta_content_scope` no longer leaks past Haskell's `declaration-type-end` branch (alt[1] = `pop: 2` on `::`). Covers `func :: Type` after a data decl, `var = Con` assignments, and the multi-function-list shape.
- `Packages/JavaScript/tests/syntax_test_js_class.js`: leaves the baseline entirely (4 assertions, same `pop-N-after-non-consuming-push` shape).

No other baseline entries change.

## What's left for Haskell 72

- Cluster C — function-identifier / operator scoping around `(<:>)` and comma-separated function lists (~43 lines). PR #643's body mis-described this as "class-context type-parens"; actual failures are function-identifier scopes.
- Cluster A — `keyword.declaration.{data,type}` missing on context-qualified decls with `=>` (~22 lines).
- Cluster D — string-continuation + operator edge cases (~7 lines).

## Tests

New regression `non_consuming_pop_n_below_pre_push_depth_is_not_a_loop` in `src/parsing/parser.rs`, next to the existing `can_parse_non_consuming_pop_that_would_loop*` cases.

Refs: #631

### DIFF
--- a/src/parsing/parser.rs
+++ b/src/parsing/parser.rs
@@ -669,9 +669,17 @@ impl ParseState {
                         min_start = match_start;
 
                         let consuming = match_end > start;
+                        // A non-consuming `pop: N` after a non-consuming push
+                        // only loops when N == 1 — that restores the exact
+                        // pre-push stack, so the push rule fires again. With
+                        // N >= 2 the stack drops strictly below the pre-push
+                        // depth, so the outer context no longer has the same
+                        // trigger in scope (e.g. Haskell's `immediately-pop2`
+                        // as the fallback branch alternative for
+                        // `declaration-type-end`).
                         pop_would_loop = check_pop_loop
                             && !consuming
-                            && matches!(match_pat.operation, MatchOperation::Pop(_));
+                            && matches!(match_pat.operation, MatchOperation::Pop(1));
 
                         let push_too_deep = matches!(
                             match_pat.operation,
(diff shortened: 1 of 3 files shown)


keith-hall:
> `find_best_match`'s `pop_would_loop` guard (`src/parsing/parser.rs`) flagged _any_ non-consuming `Pop` after a non-consuming push as looping, but `pop: N` with N ≥ 2 drops the stack strictly below the pre-push depth — the outer context no longer has the same trigger in scope, so no loop is possible. Restrict the guard to `Pop(1)`.

Question: what if a match pattern matching zero length pushes 3 contexts onto the stack, the top 2 of which are non consuming, the 2nd one popping 2 levels so the 3rd one is never evaluated for matches. Wouldn't it be back to looping again?

stefanobaghino:
Good catch — the scenario does loop, but the gap is pre-existing rather than something this patch opens.

`non_consuming_push_at` stores `(match_end, self.stack.len() + 1)` (`parser.rs:537`) — pre-push depth + 1, fine for single-context pushes. For a K-push the post-push depth is D+K; `check_pop_loop` only arms at D+1, so intermediate unwind depths silently bypass the guard. With `push: [a,b,c]`, c's `pop:1` moves D+3 → D+2 and b's `pop:2` moves D+2 → D — never hits D+1. Old `Pop(_)` guard behaves identically on that trace.

None of the bundled multi-push fixtures exercise this shape (shebang-style wrappers always end with a `pop: 1` in the topmost context, which lands at the armed depth). Happy to file a follow-up issue with the minimal repro and a fix sketch — want me to take that on?

stefanobaghino:
Filed as #656.

keith-hall (approved):

stefanobaghino:
Heads up — after this merged, `master` no longer builds from a clean checkout: the `find_not_empty` branch that `Cargo.toml`'s `fancy-regex` git dep pointed at was removed upstream once fancy-regex/fancy-regex#242 and #245 landed on `main`. Existing checkouts kept building because the SHA is cached in `Cargo.lock`, but `cargo update` (and any fresh CI checkout that rebuilds the dep cache) fails to resolve the dep.

#651 was already in flight to close the last `known_syntest_failures_fancy.txt` divergences via the same upgrade. I've moved it to the base of the follow-up stack and marked it ready for review so it can land first and unblock master. The rest of the stack (#645, #646, #648, #649, #652, #653, #654, #655, #657) was rebased onto #651 in the same logical order — only PR descriptions where the parent changed (#645's stack ref shifted from #644 to #651, #652's from #651 to #649) needed updating; everything else is a no-op rewrite.

Tracking issue #631 has been updated with the new ordering.

PR #645, opened 2026-04-20
Re-emit branch_point captures on fail-retry

stefanobaghino:
Stacked on #651. **Isolated diff:** https://github.com/stefanobaghino/syntect/compare/close-fancy-oniguruma-baseline-gap...631-haskell-decl-captures-retry

## Summary

`handle_fail` already re-emits a branch_point match's `pat.scope` (e.g. `keyword.operator.comparison.sql` on `LIKE`) after a fail rewind. It did not do the same for `captures:`-derived scopes, so Haskell `data CtxCls ctx => ModId.QTyCls` (where `data-signature` fails into `data-context`) dropped the `keyword.declaration.data.haskell` capture from the `data` token.

## Fix

Snapshot the computed capture ops on the `BranchPoint` record at creation time and replay them inside the pat_scope brackets in both fail paths (same-line and cross-line). New helper `build_capture_ops` factors the existing inline logic out so the new call site and the existing one share it.

## Impact

`Packages/Haskell/tests/syntax_test_haskell.hs`: **72 → 50**. Closes Cluster A. Remaining clusters for follow-up PRs:

- Cluster C (~43): `meta.function.identifier.haskell` missing on `functions` branch_point replay into `function-name` (`(<:>)` prefix operator decls, comma-separated function lists). The `meta_content_scope` re-emit on fail-retry is already in place — different bug.
- Cluster D (~7): string-continuation and `-->` operator edge cases.

## Tests

New regression `branch_point_capture_scopes_survive_fail_retry` in `src/parsing/parser.rs` — two-alt branch whose trigger match carries both `scope:` and `captures:`.

Refs: #631

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -322,8 +322,9 @@ checksum = "877a4ace8713b0bcf2a4e7eec82529c029f1d0619886d18145fea96c3ffe5c0f"
 
 [[package]]
 name = "fancy-regex"
-version = "0.17.0"
-source = "git+https://github.com/fancy-regex/fancy-regex?branch=find_not_empty#b8cd8fc1f296be10b4cca3c82efd183b747639ed"
+version = "0.18.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e1e1dacd0d2082dfcf1351c4bdd566bbe89a2b263235a2b50058f1e130a47277"
 dependencies = [
  "bit-set",
  "regex-automata",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ features = ["metadata"]
 [dependencies]
 yaml-rust2 = { version = "0.10.4", optional = true, default-features = false }
 onig = { version = "6.5.1", optional = true, default-features = false }
-fancy-regex = { git = "https://github.com/fancy-regex/fancy-regex", branch = "find_not_empty", optional = true }
+fancy-regex = { version = "0.18.0", optional = true }
 walkdir = "2.5"
 regex-syntax = { version = "0.8", optional = true }
 plist = { version = "1.8", optional = true }
--- a/src/parsing/parser.rs
+++ b/src/parsing/parser.rs
@@ -157,6 +157,14 @@ struct BranchPoint {
     /// ESCAPE …`, every non-whitespace before the `LIKE` fires
     /// `else-pop` in the escape-alternative, derailing the stack).
     prefix_ops: Vec<(usize, ScopeStackOp)>,
+    /// Capture Push/Pop ops emitted alongside the branch_point match's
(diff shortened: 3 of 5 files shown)


stefanobaghino:
Cluster C root cause isn't in `handle_fail` — the `meta_content_scope` re-emit is correct. `variable-name-end` lacks its prototype link, so the prototype's line-comment rule doesn't absorb the assertion-comment lines between `(<:>)` and `::`; `(?=\S) → pop: 2` fires instead, and `fail: functions` never gets a chance on the subsequent `::`.

The link is stripped by `recursively_mark_no_prototype` transitively via `prototype → preprocessor-pragmas → preprocessor-pragma-body (mip: false) → preprocessor-pragma-signature-value → include functions → branch variable-name → push variable-name-end`.

I tried three static refinements (stop-at-`mip: false`, `#anon_` check, main-reachable subtract). Each either regressed `can_parse_context_included_in_prototype_via_named_reference` / `no_prototype_for_contexts_inline_in_prototype`, or caused ~20k new syntest failures across SQL (`inside-double-dash-comment` leaks prototype → nested-comment push), LaTeX, Rails haml, the HTML `tricky_test_syntax` case, and a Zsh panic. The asymmetry is fundamental: Haskell's `functions` *wants* prototype linked (reached via main and via prototype), SQL's `inside-double-dash-comment` doesn't (same reach shape). Static marking can't tell them apart — distinguishing them needs runtime context-origin tracking. Deferred.

keith-hall (approved):

stefanobaghino:
Hey @keith-hall, not a nudge, just making sure the expectations are aligned: would you like me to rebase this PR before you merge? IIUC you can even just review the PRs on the stack one by one and merge the top-most one that passes all your checks.

FYI I'm making good progress getting rid of outstanding issues detected by syntest. I think the stack will soon fulfill its duty, at the end of which I would like to ensure some baseline performance check and start a cleanup aimed at minimizing the code necessary to pass the test suite. We can talk about that once I get there.

keith-hall:
I had the same idea about just reviewing multiple PRs and merging the highest one to save the hassle of constantly rebasing, waiting for CI on each PR etc :+1: 

Sounds good, thanks for the progress report. All this hard work manipulating stack operations in the right order does really make me wonder whether it would be simpler to just have a concept of a "token storage": simply assigning scopes to positions/text on lines as opposed to having to worry about what order to pop a meta scope in before pushing something else or clearing scopes, undoing cleared scopes etc.

stefanobaghino:
@keith-hall since this thread is where the token-storage idea came up: the current arc is up as #696–#701, which retires the in-place op-correction machinery in favor of checkpointed re-execution from a decision trail — cutting away most of the stack-op ordering choreography discussed here. The overall stack is #631-tracked items → #627 (new writer API) → #696–#701 (refactoring/simplification).

PR #646, opened 2026-04-20
Restore cleared scopes before Pop in `pop: N + set: X`

stefanobaghino:
Stacked on #645. **Isolated diff:** https://github.com/stefanobaghino/syntect/compare/631-haskell-decl-captures-retry...631-batch-quoted-value-end

## Summary

In `push_meta_ops`'s non-initial Set branch, `pop: N + set: X` with `N > 1` and `clear_scopes:` on the popping context emitted the compound Pop before Restoring the cleared atoms. The intermediate popped frame's `meta_content_scope` is split across the live scope stack and `clear_stack`; Pop — counting the full pre-Clear total — then ate atoms from frames below the popped range.

Trigger: Batch File `cmd-set-quoted-value-inner-end` (`clear_scopes: 1`) firing `pop: 2, set: ignored-tail-outer` — every quoted `set "var"=...` line lost `meta.command.set.dosbatch` on its trailing content.

## Fix

Emit Restore before Pop when `set_pop_count > 1`. Plain `set:` (`set_pop_count == 1`) keeps the existing Pop-then-Restore order, gated by `v2_set_to_target_with_clear_scopes_clears_parent_meta_content_scope` (Lisp `defun` shape).

## Impact

`Packages/Batch File/tests/syntax_test_batch_file.bat`: **74 → 0** on both Oniguruma and fancy-regex. Baseline deltas in `testdata/known_syntest_failures*.txt`. No other entry changes on either backend.

## Tests

New regression `pop_n_set_with_cur_clear_scopes_restores_before_popping_deeper_frames` in `src/parsing/parser.rs` — 3-deep stack, middle frame with 2-atom `meta_content_scope`, top frame with `clear_scopes: 1` firing `pop: 2, set: target`.

Refs: #631

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -322,8 +322,9 @@ checksum = "877a4ace8713b0bcf2a4e7eec82529c029f1d0619886d18145fea96c3ffe5c0f"
 
 [[package]]
 name = "fancy-regex"
-version = "0.17.0"
-source = "git+https://github.com/fancy-regex/fancy-regex?branch=find_not_empty#b8cd8fc1f296be10b4cca3c82efd183b747639ed"
+version = "0.18.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e1e1dacd0d2082dfcf1351c4bdd566bbe89a2b263235a2b50058f1e130a47277"
 dependencies = [
  "bit-set",
  "regex-automata",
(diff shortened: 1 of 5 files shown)


jrappen:
@keith-hall Could you take a look to advance things further in the PR chain?

stefanobaghino:
> @keith-hall Could you take a look to advance things further in the PR chain?

@keith-hall If there's anything at all you believe I can do to make the review easier, I'd be happy to do anything you believe makes sense. If you want to skim through the remaining PRs and get a sense of whether the general approach makes sense, I'm also happy to take a step back and see whether there's something that can be done to change the approach itself.

keith-hall (approved):

PR #651, opened 2026-04-22
Close fancy-regex / Oniguruma baseline gap

stefanobaghino:
Bumps the `fancy-regex` git dep to `fancy-regex/fancy-regex` `main` now that fancy-regex/fancy-regex#242 and #245 have landed, closing the three remaining divergences between `testdata/known_syntest_failures.txt` and `testdata/known_syntest_failures_fancy.txt` — the two baselines are now byte-for-byte identical.

Restores a green `master` after the `find_not_empty` fancy-regex branch was removed.

Refs: #631

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ features = ["metadata"]
 [dependencies]
 yaml-rust2 = { version = "0.10.4", optional = true, default-features = false }
 onig = { version = "6.5.1", optional = true, default-features = false }
-fancy-regex = { git = "https://github.com/fancy-regex/fancy-regex", branch = "find_not_empty", optional = true }
+fancy-regex = { version = "0.18.0", optional = true }
 walkdir = "2.5"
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -24,7 +24,7 @@ features = ["metadata"]
 [dependencies]
 yaml-rust2 = { version = "0.10.4", optional = true, default-features = false }
 onig = { version = "6.5.1", optional = true, default-features = false }
-fancy-regex = { git = "https://github.com/fancy-regex/fancy-regex", branch = "find_not_empty", optional = true }
+fancy-regex = { git = "https://github.com/fancy-regex/fancy-regex", branch = "main", optional = true }
```
keith-hall: maybe let's point to the latest release instead of main branch, they are currently identical anyway
```suggestion
fancy-regex = { version = "0.18.0", optional = true }
```

keith-hall (approved):

PR #662, opened 2026-04-26
Route handle_fail cross-line re-emit through push_meta_ops

stefanobaghino:
Stacked on #660.

Compare against PR #660's head:
https://github.com/stefanobaghino/syntect/compare/631-java-class-extends-annotation-leak...631-java-multiline-enum-meta-scope-leak

Cross-line counterpart to #660: `handle_fail`'s cross-line branch
re-emit kept the pre-#660 bespoke meta_scope/meta_content_scope
push, missing the popped contexts' Pop. Routes through
`push_meta_ops` with a synthetic Set/Push, mirroring the same-line
fix.

Closes the simple cross-line case (qualified annotation followed
by class/enum/interface on the next line). The dominant
`syntax_test_java.java` cascade past the multi-line enum at lines
2260-2297 has additional layers — separate follow-up. Java
baseline: 9956 → 9935.

Unit test in `src/parsing/parser.rs::tests`:
`cross_line_pop_n_branch_point_alt_fail_unwinds_meta_scope`.

### DIFF
--- a/examples/syntest.rs
+++ b/examples/syntest.rs
@@ -278,6 +278,7 @@ fn get_line_assertion_details<'a>(
 fn process_assertions(
     assertion: &AssertionRange<'_>,
     test_against_line_scopes: &[ScopedText],
+    next_line_scopes: Option<&[ScopedText]>,
 ) -> Vec<RangeTestResult> {
     // format the scope selector to include a space at the beginning, because, currently, ScopeSelector expects excludes to begin with " -"
     // and they are sometimes in the syntax test as ^^^-comment, for example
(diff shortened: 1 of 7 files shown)


keith-hall (approved):

stefanobaghino:
All PRs in the approved segment of the stack (#646, #648, #649, #652, #653, #654, #655, #657, #658, #659, #660, #662) are now marked ready for review. Since this branch contains every commit of the other eleven, there are two ways to land the segment:

**Option A — merge one at a time, bottom-up** (#646 → #648 → #649 → #652 → #653 → #654 → #655 → #657 → #658 → #659 → #660 → #662), each with a merge commit, same as the stack so far. After each merge GitHub automatically shrinks the next PR's diff to its own delta, so no rebasing is needed in between. This keeps one merge commit per PR in the first-parent history of `master`.

**Option B — merge just this PR with a merge commit.** That lands the whole approved segment in one operation; GitHub then detects that each intermediate PR's head is reachable from `master` and marks #646–#660 as merged automatically. CI on this branch has already tested the exact tree `master` would become, whereas sequential merging creates intermediate trees CI never saw. The trade-off is that the first-parent history shows a single merge for all twelve PRs.

Either way the merge needs to be a **merge commit, not a squash** — a squash would rewrite the SHAs, leave the intermediate PRs open, and force a full rebase of the rest of the stack (#663 onwards, which is pending review and shrinks automatically under both options above).

examples/syntest.rs
```diff
@@ -516,6 +539,10 @@ fn test_file(
                             col += len;
                         }
                     }
+                    let next_idx = start_idx + i + 1;
+                    if next_idx < buf_len {
+                        corrected_baselines.push((next_idx, stack.clone()));
```
keith-hall, line 544: I do wonder whether we can avoid some of the cloning using CoW or other tactics, but I think it makes sense to get the implementation correct in terms of output first, then we can optimize later - especially if the new approach at https://github.com/trishume/syntect/pull/645#issuecomment-4887364474 will make a lot of this redundant, I'm looking forward to getting to those PRs :+1:

PR #668, opened 2026-04-27
Skip cur mcs Pop in Set when wrapper has embed_scope_replaces

stefanobaghino:
**Review just this PR's changes:** https://github.com/stefanobaghino/syntect/compare/631-java-lambda3-inner-annotation...631-markdown-fence-embed-scope-token-drop

The Set initial-phase Pop at `src/parsing/parser.rs:1992` unconditionally popped `cur_context.meta_content_scope.len()` even when cur_context's mcs was never pushed because the context immediately below has `embed_scope_replaces=true`. That dropped the topmost wrapper-pushed `embed_scope` token (e.g. `source.shell.bash` from a Markdown fenced bash block). Mirrors the skip already in the Pop branch at `src/parsing/parser.rs:1912`.

Markdown syntest baseline: 158 → 31. Python: 45 → 32 (free benefit — Python's syntax also relies on `embed_scope_replaces`).

Stacked on #667.


src/parsing/parser.rs
```diff
@@ -5966,4 +6851,876 @@ contexts:
...
+    }
+
+    /// Regression guard for the `embed_scope`-replaces / inner Set
+    /// interaction. With a wrapper context that pushes 3 mcs scopes
+    /// and `embed_scope_replaces=true`, the embedded syntax's first
+    /// `set:` rule must not drop the topmost wrapper scope — its mcs
+    /// pop must be skipped because the embedded main's mcs was never
+    /// pushed.
+    #[test]
+    #[ignore = "requires testdata/Packages submodule"]
```
keith-hall: I'm a bit confused about these ignores - we have the submodule so why ignore the test? Ignoring means they don't run in CI, meaning they could be failing, and have no (guaranteed) value...

stefanobaghino: Fair point — CI does check out the submodule, so the ignore only cost coverage (the attribute followed the pattern #630 introduced). I've removed it from this test, from the three the stack adds later (#673, #677, #686), and — since this is where the topic came up — from the two #630 put on master, so all six now run in the regular suite. Doing so immediately validated your concern: the Packages bump in #680 had silently invalidated this test's expected scope atom (`source.shell.bash` → `source.shell.bash.embedded.markdown` in the fenced-bash `embed_scope`) — fixed in #680.

keith-hall (approved):

PR #677, opened 2026-04-28
Defer LRD blank-line pop to next-line baseline

stefanobaghino:
A non-terminated Markdown link reference definition title spanning a
content line and a following empty line had its `meta_scope` collapsed
on the wrong line boundary. Without the fix, the chained
`link-title-continuation` + `link-def-attr-continuation` exhaustion
emitted the LRD pop on line 2's `\n` (= line 3's stack baseline),
dropping `meta.link.reference.def.markdown` on the empty line. ST
emits that pop one boundary later (between line 3 and line 4).

Two-part guard in `src/parsing/parser.rs`:

- `parse_next_token`: skip a non-consuming `Branch` whose match lands
  at or past the replay line's end when inside `replay_ctx`. Prevents
  a chained `branch_point` from being created at end-of-replay-line.
- `handle_fail` same-line path: when the rewind position is 0 of an
  empty line (length ≤ 1, just `\n`), advance the cursor to
  `line.len()` so a follow-up `match: '' pop: N` emits past-EOL.
  `ScopeRegionIterator` wraps that to the next line's baseline.

Baseline: Markdown 1 → 0 (`syntax_test_markdown.md` row removed from
`testdata/known_syntest_failures{,_fancy}.txt`). No other rows change.

Stacked on #676. Isolated diff:
https://github.com/stefanobaghino/syntect/compare/631-zsh-range-clear-parent-string-scope...631-markdown-lrd-blank-line


stefanobaghino:
CI failures looks due to a transient failure when grabbing Zig from Codeberg (same check on #678 succeeded), but I don't seem to have the permissions to restart it.

PR #680, opened 2026-04-29
Bump testdata/Packages to v4202

stefanobaghino:
**Review just this PR's changes:** [`631-syntest-haskell-meta-function-identifier...631-syntest-bump-packages-submodule`](https://github.com/stefanobaghino/syntect/compare/631-syntest-haskell-meta-function-identifier...631-syntest-bump-packages-submodule)

Stacked on #679.

Pins `testdata/Packages` to the shipped `v4202` tag (`91ad8085`,
"[D, Makefile, Rust] Standardize build output scopes"). v4202 is
the most recent stable release before the C# v2 migration
`8621831d` and the regex embed grammar `c735169b`; pinning here
keeps `regex_string` on its legacy `embed_scope` form, sidestepping
the wrapper-mcs divergence between syntect and ST DEV that
produced the `syntax_test_C#11.cs: 35` baseline entry.

Companion fix in `parser.rs::push_meta_ops`: keep the auto-injected
top-level scope across v2 set's cur.mcs Pop. The previous Pop sized
to `cur_context.meta_content_scope.len()` overcounted on `main`,
where `add_initial_contexts` injects the syntax's top-level scope
at position 0. ST 4200 stable keeps that scope visible at the
trigger of TOML's `[section]` rule (probed via the `sublime-mcp`
skill); without this fix, the v4202-era
`Rust/tests/syntax_test_frontmatter.{rs,md}` would fail. The
upstream workaround `20212766` is post-v4202 and out of scope here.

`testdata/known_syntest_failures{,_fancy}.txt` lose the C#11 row
(net –35); Java row at 212 unchanged. `make syntest` and
`make syntest-fancy` end clean. `parsing::syntax_set::tests::can_load`
ticks 185 → 184 (one context added upstream post-v4202).


stefanobaghino:
## Why the v4202 pin (and not later) — context on the C#11 baseline entry

The `syntax_test_C#11.cs: 35` row that lived on `master`'s
`testdata/known_syntest_failures.txt` is upstream-test-vs-syntect
divergence introduced by `c735169b` ("[C#] Add regex to argument
strings in known functions", #4522, 2026-04-15) — the first commit
*past* v4202 that adds the new hidden grammar
`Packages/C#/Embeddings/Regex (for C#).sublime-syntax`
(`scope: source.regexp.cs`,
`extends: Packages/Regular Expressions/RegExp (Basic).sublime-syntax`,
`hidden: true`).

### What c735169b changed

C#'s `regex_string` context has two arms — non-verbatim `"` and
verbatim `` `@"` ``. Pre-c735169b they shared
`embed: scope:source.regexp; embed_scope: meta.string.cs meta.regexp.cs`.
Post-c735169b they switch to `embed: scope:source.regexp.cs` with
arm-specific embed scopes:

```yaml
- match: '"'
  embed: scope:source.regexp.cs
  embed_scope: meta.string.cs string.quoted.double.cs source.regexp
- match: '@"'
  embed: scope:source.regexp.cs
  embed_scope: meta.string.cs string.quoted.double.verbatim.cs source.regexp
```

The test (`syntax_test_C#11.cs:686, 719`) was updated alongside to
assert content scopes that **drop `.verbatim.cs`** even on the
verbatim arm:

```
[GeneratedRegex(@"^some(?<named_capture>[^/]{1,100})$")]
//                 ^^^^^... meta.string.cs string.quoted.double.cs source.regexp meta.mode.basic.regexp
```

But the verbatim arm's
`embed_scope: meta.string.cs string.quoted.double.verbatim.cs source.regexp`
was **not** updated to drop `.verbatim.cs` — the YAML still names
the verbatim form.

### Empirical: what each implementation produces

Probed via the `sublime-mcp` skill (ST 4200 stable on macOS arm64,
with `~/Library/Application Support/Sublime Text/Packages/C#`
symlinked at `testdata/Packages/C#`) and via
`cargo run --release --example syntest`:

| Position | Test asserts | ST 4200 stable | syntect | ST DEV (inferred) |
| --- | --- | --- | --- | --- |
| `:686:22` (non-verbatim regex `"abc\|def"`) | `string.quoted.double.cs source.regexp meta.mode.basic.regexp` | `string.quoted.double.verbatim.cs source.regexp …` (fail) | same as ST stable (fail) | matches test (pass) |
| `:719:23` (verbatim regex `` `@"^some…"` ``) | same | `string.quoted.double.verbatim.cs source.regexp …` (fail) | same (fail) | matches test (pass) |

`run_syntax_tests("Packages/C#/tests/syntax_test_C#11.cs")` against
ST 4200 stable returns `FAILED: 2 of 562 assertions failed` at
exactly those two positions — the same divergence syntect reports
and counts at 35 in path-1 sum.

The non-verbatim arm pulling `.verbatim.cs` at line 686 (where the
trigger is just `"`, not `` `@"` ``) suggests ST is coalescing /
sharing wrappers across the two `regex_string` arms — interesting
on its own but secondary; the test passes in ST DEV because DEV's
renderer drops `.verbatim.cs` from the emitted scope chain
entirely. That mechanism is internal to ST's renderer and out of
scope for syntect to replicate without ST DEV available locally.

### Why upstream CI doesn't catch it

`sublimehq/Packages` `.github/workflows/ci.yml` runs a 2-build
matrix:

- `stable build 4200`: `optional: true` →
  `continue-on-error: true` (non-gating).
- `dev (latest)`: `optional: false` (gating).

The 2 assertions are red on the stable job, but its
`continue-on-error` makes the matrix succeed. Only `dev (latest)`
is the actual gate — and ST DEV passes the test.

### Why v4202 specifically

Tested by `git checkout` in the submodule +
`make syntest` + `cargo test --release --all-features --lib`:

| Tag | Submodule SHA | C#11 row | Other syntest deltas | Unit-test fixture deltas |
| --- | --- | --- | --- | --- |
| v4200 (= v4199) | `85b7cdc8` | gone | Rust frontmatter `.md`/`.rs` 1 each | `can_load` 185→184; `test4.html` regen for older `Cargo.sublime-syntax`; metadata `Indention` typo |
| **v4202 (= v4201)** | **`91ad8085`** | **gone** | Rust frontmatter `.md`/`.rs` 1 each (this PR fixes them via `parser.rs`) | `can_load` 185→184; metadata `Indention` typo |
| v4204 (= v4205) | `c735169b` | **back at 35** | Java unchanged | `can_parse_preprocessor_rules` (C directive refactor `44871676`); `can_load` 185→184 |

v4202 is the latest stable tag that predates `c735169b` (and its
prerequisite `8621831d`, "[C#] Migrate to sublime syntax version 2"
#4504, which introduces the verbatim/non-verbatim split in the
first place). Bumping past v4202 brings the C#11 row back unless
syntect grows a model of ST DEV's wrapper-coalescing renderer.

### Future direction

Either:

- Match ST DEV's wrapper-coalescing rule for shared `embed:`
  targets, exposed only post-c735169b. Requires ST DEV running
  locally to ground-truth the rule (Linux dev binary or Docker).
  Out of scope for this PR.
- Wait for upstream to update the verbatim arm's `embed_scope:` to
  drop `.verbatim.cs`. ST 4200 stable would then converge with ST
  DEV and syntect would pass too. Tracked as an upstream concern
  on `sublimehq/Packages`, not actionable from syntect.

For now, v4202 is the path of least resistance — it sidesteps the
divergence by pinning at the YAML state where the new hidden
`source.regexp.cs` grammar doesn't yet exist.

PR #682, opened 2026-04-29
Bound prefer_inner_replay_corrections by depth_diff to substitute siblings only

stefanobaghino:
**Review just this PR's changes:** [`631-java-cross-line-bp-fall-through...631-java-cross-line-allexhaust-pops`](https://github.com/stefanobaghino/syntect/compare/631-java-cross-line-bp-fall-through...631-java-cross-line-allexhaust-pops)

Stacked on #681.

The previous gate in
[`prefer_inner_replay_corrections`](https://github.com/stefanobaghino/syntect/blob/631-java-cross-line-allexhaust-pops/src/parsing/parser.rs)
(commit
[`0a2139a`](https://github.com/stefanobaghino/syntect/commit/0a2139a))
skipped substitution iff `inner.stack_depth > outer.stack_depth`.
That predicate collapsed two structurally distinct cases:

- **Sibling refinement** (substitute needed): inner is one level
  deeper than outer on the same line — e.g. cluster 1's
  `@Anno\n.\nAnno\n(par=1)\nenum E {}` with
  `outer=declarations(3), inner=annotation-identifier(4)`. Inner's
  qualified-identifier alt brings `meta.path.java` that outer's
  locally-computed parse drops.
- **Child of resolved alt** (substitute must skip): inner is much
  deeper, nested inside contexts outer's resolved alt pushed —
  e.g. multigen16's `outer=class-members(4),
  inner=object-type(11)`. Inner's reparse adds atoms outer's alt
  already provides (the
  `deeper_inner_bp_correction_does_not_double_outer_meta_scope`
  guard).

This PR tightens the gate to `depth_diff in {0, 1}`, separating
the two by the smallest viable structural signal. Java syntest
baseline drops 119 → 117. The cluster-1 ignored repro becomes a
passing regression test
(`cross_line_all_exhaust_with_pop_count_emits_popped_meta_scope_pops`).
The doubling guard stays green.

Cluster 2's multi-line qualified field type (depth_diff=5) and
cluster 3's `pop: 2` miscount remain unresolved; per investigation
captured in scratch on the stacked branch, neither has a
local-signal discriminator distinguishing it from the doubling
cascade. Those are deferred to a separate pass.


stefanobaghino:
## Cluster-2 / multigen16 investigation log

A separate investigation pass on a stacked branch tried several architectural directions to also fix cluster 2 (and unblock the `meta.field.type.java` regression at `syntax_test_java.java:3436`). All blocked by the cascade nature of the duplication. Recording the attempts so the next investigator doesn't repeat them.

### What the probe captured

Parsing `syntax_test_java.java` lines 2730–3460 with full tracing on, the cascade at parser line 702 (file 3432) fires **five** `prefer_inner_replay_corrections` calls. Three are skip cases under the depth gate; two are substitute cases:

| outer | inner | depth_diff | decision | outer.creator | inner.creator |
|---|---|---|---|---|---|
| `object-type(9)` | `qualified-object-type(10)` | 1 | substitute | `Some(4)` | `Some(4)` |
| `object-type(9)` | `qualified-object-type(10)` | 1 | substitute | `Some(4)` | `Some(4)` |
| `class-members(4)` | `object-type(9)` | 5 | skip — **cluster-2 seat** | `None` | `Some(4)` |
| `object-type(11)` | `class-members(4)` | -7 | skip | `Some(4)` | `None` |
| `class-members(4)` | `object-type(11)` | 7 | skip — **doubling seat** | `None` | `Some(4)` |

Cluster 2 (substitute wanted) and the multigen16 doubling (skip wanted) have **identical** lineage signatures — only `inner.stack_depth` differs (9 vs 11). No purely-local signal at the gate distinguishes them.

### Directions tried, all reverted

1. **Content discriminator** — skip iff `inner_count(s) > outer_count(s) > 0` for any scope `s`. Failed: the doubling at line 3436 is a cascade of earlier substitutions whose individual contents don't exhibit the duplication signature locally. Java syntest 117 → 424.

2. **BP lineage — `creator_bp_depth`** (stack_depth of `branch_points.first()` at creation). Both cluster 2 and the doubling case have `inner.creator_bp_depth = Some(4)` (the outer `class-members(4)` was active at inner's creation). 117 → 393 if used as override.

3. **BP lineage — `created_during_replay`** (`replay_ctx.is_some()` at creation). Both cases have it `true`. 117 → 423 if used as override.

4. **Sibling-by-creator-match** — substitute when `outer.creator_bp_depth == inner.creator_bp_depth`. No-op on the Java region: cluster 2 and doubling both have **asymmetric** creator depths (`outer.creator=None`, `inner.creator=Some(4)`), so the override never fires. 117 stays at 117.

5. **`merge_flushed` semantics — keep prior when new is deeper**. Idea: don't let late writes overwrite earlier-resolved refinements. Broke `cross_line_pop_n_branch_point_alt_fail_unwinds_meta_scope` (`` `@A.B\nclass E {}` `` annotation cleanup) — that test's correctness relies on a deeper alt's later write replacing the prior. The current `replace_all` / `keep_prefix_replace_suffix` rules are load-bearing for cross-line cases beyond cluster 2.

6. **Output coalescing — per-line relative-stack `Push(X)` skip**. Lifted the depth gate and scrubbed duplicates with a per-line tracker that skipped `Push(X)` when X was at top of the relative stack. Doesn't help: the bad Push fires *after* an intervening `Clear(TopN(1))` / `Restore` sequence resets the per-line tracker, while the running stack underneath retains the original push (re-asserted by `Restore`). Pattern:
   ```
   Push(type), Push(modifier), Pop(2)
   Restore                                    ; running stack: [..., type, modifier]
   Push(identifier), Clear(TopN(1))           ; running stack: [..., type, modifier]
   Push(type)                                 ; bad duplicate — under modifier
   ```
   Detecting the duplicate requires fully simulating `Clear`'s snapshot and `Restore` — i.e. reimplementing the parser's stack machinery in the dedup pass. 117 + coalescing → still 424.

### Findings summary

- The doubling at `syntax_test_java.java:3436` is a pre-existing **cascade** expressed via `Restore` re-pushing scopes that outer's resolved alt also pushes. Locally-correct gate decisions accumulate duplicate atoms across multiple merges.
- Local signals (lineage, content, simple coalesce) can't see the cascade.
- `merge_flushed`'s last-write-wins semantics are load-bearing — naive inversions break unrelated cross-line cases.

### Remaining viable directions (deferred)

- **Full stack simulation in coalescing** — track Clear/Restore snapshots, dedup `Push(X)` when X is on the running stack. Invasive (reimplements the parser's stack semantics in the dedup pass).
- **Stack-relative inner corrections** — change the parser to express inner's corrections as a delta-form so substitution doesn't reset/conflict with outer's stack ancestry. Architectural.
- **Syntax-definition-level rework** — change `Java.sublime-syntax` to avoid the `Clear/Restore` interaction with the qualified-object-type's `meta_scope`. Out of syntect's scope.

### Diagnostic infrastructure

Added but **not** included in this PR (kept on a stacked branch for future pickup, since it doesn't drive syntest down):

- A `tracing` instrumentation pass with hierarchical span dumps via `tracing-tree`, covering BP creation/removal/retry, `merge_flushed` decisions with prior/new BP info, gate decisions, replay-iteration spans, op-content traces. Reproduces the cascade tree at a glance.
- `_probe_cluster2_and_doubling` — an ignored test that parses lines 2730–3460 of `syntax_test_java.java` and dumps the trace; the table above came from this probe.
- `creator_bp_depth` and `created_during_replay` fields on `BranchPoint` / `BpInfo`.

If picking this back up, that infrastructure is the right starting point.

PR #690, opened 2026-05-06
Force `class-members` alt 5 via inline alt-5 re-run on composite discriminator

stefanobaghino:
**Review just this PR's changes:** https://github.com/stefanobaghino/syntect/compare/631-java-field-type-meta-path-drop...631-java-class-members-alt-cycle

Forces alt 5 (`member-maybe-field`) on the cross-line `class-members` `branch_point` when a composite discriminator fires at the post-`InnerCorrectionsCaptured` observation point in `handle_fail`'s alt-cycle path. The discriminator: `outer_bp_info.name == "class-members"`, `next_alt_index < 5`, the inner replay produced no corrections, the inner-replay-max BP is strictly deeper than outer, and the first replayed slot pushes a scope prefixed by `meta.function.return-type.java`.

When it fires, the rewind discards the failed alt's `replayed_ops`, restores BP-creation snapshots (`branch_points`, `stack`, `proto_starts`, `escape_stack`, `first_line`, `non_consuming_push_at`), re-executes the alt-cycle setup with alt 5 inline, re-runs the inner replay loop, and reassigns the outer locals so the existing merge code sees alt 5's results. The `next_alt_index < 5` short-circuit prevents a post-rewind re-fire.

A simpler "set `next_alternative = 5` and let the caller re-trigger" form regressed cluster-B sentinels — the caller's re-parse doesn't re-encounter the same BP at the same byte position, so the target alt's replay never runs. The inline re-run is what restores correctness on small fixtures while fixing the real-file cluster.

Java syntest baseline drops 75 → 0 cols on both default-onig and default-fancy backends, clearing the residual cross-line Java cluster tracked under #631.

Stacked on #686 → #687 → #688 → #689.

Refs: #631


testdata/known_syntest_failures.txt
```diff
@@ -1,24 +1,2 @@
...
-FAILED testdata/Packages/Markdown/tests/syntax_test_markdown.md: 1
-FAILED testdata/Packages/PHP/tests/syntax_test_php.php: 1
-FAILED testdata/Packages/Python/tests/syntax_test_python.py: 1
-FAILED testdata/Packages/Python/tests/syntax_test_python_strings.py: 1
-FAILED testdata/Packages/Rails/tests/syntax_test_rails.haml: 65
-FAILED testdata/Packages/Rails/tests/syntax_test_rails.html.erb: 23
-FAILED testdata/Packages/ShellScript/Bash/tests/syntax_test_scope.bash: 1
-FAILED testdata/Packages/ShellScript/Zsh/tests/syntax_test_scope.zsh: 1
-exiting with code 1
+exiting with code 0
```
stefanobaghino, line 2: @keith-hall Almost there. 🙂

PR #702, opened 2026-07-08
Fix as_latex_escaped: correct RGB channel order and backslash escaping

### LINKED ISSUE #472: `as_latex_escaped`: two small issues
First of all: I’m delighted that `as_latex_escaped()` even exists!

I have encountered two minor issues ([see Gist with unit tests](https://gist.github.com/rauschma/f36aabe3ac21e8a54a9a54e872d29198)):

* The last two components of RGB colors are flipped, i.e. the numbers in `\textcolor[RGB]{38,210,139}` should appear in this order: `38,139,210`
* A backslash should be escaped as `\textbackslash{}`, not as a double-backslash `\\` (which is a line break in LaTeX).

## Escaping

In general, I’m escaping as following in LaTeX:

```rs
HashMap::from([
    (r"#", r"\#"),
    (r"$", r"\$"),
    (r"%", r"\%"),
    (r"&", r"\&"),
    (r"~", r"\textasciitilde{}"),
    (r"_", r"\_"),
    (r"^", r"\textasciicircum{}"),
    (r"\", r"\textbackslash{}"),
    (r"{", r"\{"),
    (r"}", r"\}"),
])
```

Due due `commandchars=\\\{\}`, only these entries are needed inside `Verbatim`:

```rs
HashMap::from([
    (r"\", r"\textbackslash{}"),
    (r"{", r"\{"),
    (r"}", r"\}"),
])
```

teddytennant:
Fixes two bugs in `as_latex_escaped` (`src/util.rs`):

1. **Swapped color channels.** The `\textcolor[RGB]{...}` directive was emitted with the green and blue components transposed (`style.foreground.r, style.foreground.b, style.foreground.g`), so colors came out wrong.

2. **Broken backslash escaping.** A literal backslash in the text was escaped to `\\`, which is a line break inside the `commandchars=\\\{\}` Verbatim environment rather than a literal backslash, and the sequential `.replace()` loop could re-process characters it had just introduced. Escaping is now done in a single pass over the input, mapping `\` → `\textbackslash{}`, `{` → `\{`, `}` → `\}`, which produces LaTeX that compiles.

For input `a\b{c}` with foreground `RGB(10,20,30)` the output is now `\textcolor[RGB]{10,20,30}{a\textbackslash{}b\{c\}}`.

### Test

Added `test_as_latex_escaped`, which builds a `Style` with a known foreground color and text containing a backslash and braces, and asserts the channel order, `\textbackslash{}`, escaped braces, and the absence of a bare `\\`.

```
cargo test --lib as_latex
```

Closes #472.


src/util.rs
```diff
@@ -128,11 +125,14 @@ pub fn as_latex_escaped(v: &[(Style, &str)]) -> String {
         } else {
             write!(s, "{}", textcolor(&style, true)).unwrap();
         }
-        content = text.to_string();
-        for &(old, new) in LATEX_REPLACE.iter() {
-            content = content.replace(old, new);
+        for c in text.chars() {
```
keith-hall: I wonder if there is a faster way to go through the chars in a string, by searching for the slash or curly braces, and writing the rest in chunks instead of going char by char and pushing each one individually. Especially if the text contains none of those chars.

teddytennant:
Good idea. It now searches for the next `\`, `{` or `}` and copies everything between matches in one go, so text with none of those is a single append. Pulled it out into a `write_latex_escaped` helper and added tests for the no-special-chars and multi-byte cases.

keith-hall (approved):
Thanks, looks good. I'm a little unsure of when best to merge it, please bear with me. We are in the middle of supporting newer sublime-syntax features etc and I wonder about conflicts with https://github.com/trishume/syntect/pull/627 also

src/util.rs
teddytennant: yeah, 627 rewrites this function. this is just the rgb/escape fix on current master; if 627 lands first I can rebase.

teddytennant:
627 already has the RGB channel order, but it still maps `\` to `\\` via LATEX_REPLACE, which is a line break inside Verbatim. this PR is the escaping fix.

